PR #34, opened 2021-11-23
Update to a python 3.10 compatible black

JordonPhillips:
This updates black to a version that supports handling python 3.10 syntax like match statements.

mypy is still working on it sadly


nateprewitt (approved):
:shipit:

PR #35, opened 2022-01-05
Update deps

JordonPhillips:
This updates a the dependency versions for gradle, spotbugs, and smithy. One of the features introduced in Smithy 1.16 was a built in way to add synthetic input / output shapes to operations, so I added that in. The `GetCurrentTime` operation showcases this as it does not define an input shape.


nateprewitt (approved):
⛵

PR #39, opened 2022-02-07
Refactor PythonWriter to use CodegenWriter

JordonPhillips:
This refactors the `PythonWriter` to implement `CodegenWriter`, which handles a lot of the dependency management mess.

As part of this, I changed a few existing interfaces. Namely I made the `importStdLibDependency` use a more consistent argument layout, and updated existing calls of it to use the two argument form where possible (which was all cases).


nateprewitt (approved):
⛵️

PR #40, opened 2022-03-03
Make use of pants build system

JordonPhillips:
Eventually we're going to have a ton of packages to build and release, and doing so manually will be incredibly painful. This PR introduces the [pants](https://www.pantsbuild.org/) build system to take care of that. Some basic usage:

```
> ./pants fmt ::
> ./pants check ::
> ./pants lint ::
> ./pants test ::
> ./pants package ::
```

Pros:
* Centralizes configuration
* Build / publish all packages in one step
* Can write plugins in python

Cons:
* Complexity. If we want to integrate some tool that isn't already integrated, we'll have to write extra code.
* File bloat. It requires `BUILD` files in every directory for metadata.

To give some more context, here's what I think the overall folder structure could look like:

```
.
├── aws-clients
│   ├── aws-sdk-cloudformation
│   └── aws-sdk-s3
└── python-packages
    ├── smithy-aws-client
    ├── smithy-client
    ├── smithy-core
    └── smithy-server
```

Or, alternatively:

```
.
└── python-packages
    ├── aws-clients
    │   ├── aws-sdk-cloudformation
    │   └── aws-sdk-s3
    └── smithy-packages
        ├── smithy-aws-client
        ├── smithy-client
        ├── smithy-core
        └── smithy-server
```

With the `aws-clients` package representing one published package for each service. How the whole modularization thing will work in practice given the limits of pypa is still up for debate, but I do still think that's the goal.


nateprewitt (commented):
Overall, looks fine. Left a quick comment about linting setups.

joguSD (approved):

PR #41, opened 2022-03-04
Add request multiplexer

JordonPhillips:
This adds in a fairly simple request multiplexer to be used for SSDKs. Since it's using match, I pinned us to a a version of mypy that supports it. Hopefully they actually cut a release sometime soon.

This code will ideally go in a separate package that only the ssdks depend on, but for the sake of having a legible PR I put the code here until a plan for handling multiple packages is settled on.


crisidev (dismissed):
LGTM

nateprewitt (approved):

PR #47, opened 2022-05-04
Use directed codegen

JordonPhillips:
This updates the python generator to use the directed codegen. This reduces the amount of fiddly repetitive work we have to do.

This relies on https://github.com/awslabs/smithy/pull/1214 to some extent since we need the topological sorting.


nateprewitt (commented):
Looks good, just one minor comment about mypy (and I guess potentially black)

nateprewitt (approved):

PR #49, opened 2022-06-27
Add presence tracking to structure design

JordonPhillips:
This will be needed due to https://github.com/awslabs/smithy/pull/1286


nateprewitt (changes requested):
Two small comments, otherwise things look good!

nateprewitt (commented):
I think this looks good for the code changes. It looks like both Pants and the Python Codegen build are failing though. Do we know what's going on there?

PR #50, opened 2022-10-14
Update to pass with idl 2

JordonPhillips:
This updates some dependencies and, importantly, makes the code generator work with the IDL 2 updates.


nateprewitt (approved):
Looks good. One minor question, but otherwise should be set to merge.

PR #56, opened 2022-10-28
Add interface / base class for interceptors

JordonPhillips:
This introduces the base class for interceptors and the necessary helper classes.

Potentially the most questionable part of this is the context object. I've used `@property` to make the static properties "immutable" since we don't want people changing them outside of the methods designated for that. Of course nothing is stopping them from just mutating properties on the target objects. The question is, should I just give up even this small attempt at commanding the tide?

### DIFF
--- a/python-packages/smithy-python/smithy_python/interfaces/interceptor.py
+++ b/python-packages/smithy-python/smithy_python/interfaces/interceptor.py
@@ -1,0 +1,613 @@
+from copy import deepcopy
+from typing import Any, Generic, TypeVar
+
+Request = TypeVar("Request")
+Response = TypeVar("Response")
+TransportRequest = TypeVar("TransportRequest")
+TransportResponse = TypeVar("TransportResponse")
+
+
+class InterceptorContext(
+    Generic[Request, Response, TransportRequest, TransportResponse]
+):
+    def __init__(
+        self,
+        *,
+        request: Request,
+        response: Response | Exception,
+        transport_request: TransportRequest,
+        transport_response: TransportResponse,
+    ):
+        """A container for the current data available to an interceptor.
+
+        :param request: The modeled request for the operation being invoked.
+        :param response: The modeled response for the operation being invoked. This will
+            only be available once the transport_response has been deserialized or the
+            attempt/execution has failed.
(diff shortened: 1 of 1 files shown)


nateprewitt (commented):
A couple minor comments but otherwise looks good.

jonemo (commented):
A couple of comments on wording in the docstrings, plus one question about the `Exception` annotation for `response`. 

It's not quite clear to me what an actual implementation of the interceptor interface would look like that passes mypy muster and doesn't have to repeat much of the complexity of the type annotations from this file. For example, if I write an interceptor that implements two hooks and my messages are NamedTuples that serialize to str, would I have to write out all the following just for the type annotations?

```python
class MyInterceptorContext(InterceptorContext[NamedTuple, NamedTuple, str, str]):
    pass

class MyInterceptor(Interceptor[NamedTuple, NamedTuple, NamedTuple, NamedTuple]):
    def read_before_execution(
        self, context: InterceptorContext[NamedTuple, None, None, None]
    ) -> None:
        do_some_stuff()

    def modify_before_completion(
        self,
        context: InterceptorContext[NamedTuple, NamedTuple, str | None, str | None],
    ) -> NamedTuple | Exception:
        do_some_stuff()
```

(Apologies if I am missing one or more points in this review. I used this PR review as one way of familiarizing myself with this repo and our spec for this.)

nateprewitt (dismissed):
Can we get the typo fixed, other than that this looks good. I left a quick comment on potential exceptions for early access to some of the properties.

I think the @property approach is at good first step. I'll need to find the time to look at this but it would be nice if we could have some proxy object returned during the read steps that is discarded to avoid mutating.

That may not be feasible with copying/initialization, but it would be nice to at least get benchmarking on it.

jonemo (dismissed):
LGTM (pending the one typo): Implements the spec and this provides useful type check time guardrails for implementers of interceptors. I am curious how what the tradeoff between utility of these guardrails and the complexity they introduce will play out when actually writing interceptors, we'll see that in future PRs.

nateprewitt (dismissed):
🚀

nateprewitt (approved):

PR #60, opened 2022-11-10
Implement initial protocol test stub generation

haydenbaker:
### Description
Test stubs for request/response/response-error tests can be generated using running an instance of `HttpProtocolTestGenerator`. We can use this class within the implementation of a `ProtocolGenerator` to generate a suite of tests for the protocol

### Testing
When codegen is ran for the test model (`Weather`), the following is generated:

```python
# Code generated by smithy-python-codegen DO NOT EDIT.

from pytest import mark


async def test_write_get_city_assertions():
    """Does something
    """
    pass

@mark.skip()
async def test_write_get_city_response_assertions():
    """Does something
    """
    pass

async def test_write_no_such_resource_assertions_error_get_city():
    """Does something
    """
    pass

async def test_write_no_such_resource_assertions_error_get_city_announcements():
    """Does something
    """
    pass

async def test_write_no_such_resource_assertions_error_get_city_image():
    """Does something
    """
    pass

async def test_write_list_cities_assertions():
    """Does something
    """
    pass

```

The output includes request/response/error-response protocol tests for the client
> 
```
@mark.skip()
test_write_get_city_response_assertions
```

The test above is skipped due to a `true` value for `shouldSkip`, this is for example only.

### DIFF
--- a/codegen/smithy-python-codegen/build.gradle.kts
+++ b/codegen/smithy-python-codegen/build.gradle.kts
@@ -33,4 +33,5 @@ buildscript {
 dependencies {
     api("software.amazon.smithy:smithy-codegen-core:$smithyVersion")
     implementation("software.amazon.smithy:smithy-waiters:$smithyVersion")
+    implementation("software.amazon.smithy:smithy-protocol-test-traits:$smithyVersion")
 }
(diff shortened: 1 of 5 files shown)


JordonPhillips (changes requested):

JordonPhillips (changes requested):

JordonPhillips (approved):

JordonPhillips (changes requested):
Looks good, we just need a rebase.

JordonPhillips (approved):

JordonPhillips (changes requested):

JordonPhillips (approved):

PR #65, opened 2022-11-21
Add a basic endpoint provider

JordonPhillips:
This adds a basic endpoint provider that simply forwards along a static endpoint, parsing it if it's a string.

In the near future we should move all of these http implementations out into a more public space, but for now I elected to keep things together to keep this change as simple as possible.


nateprewitt (dismissed):
Couple minor nits but otherwise looks good!

jonemo (dismissed):

jonemo (approved):

PR #136, opened 2023-02-15
sigv4 implementation without using CRT

dlm6693:
*Description of changes:*
This PR implements the AWS Signature Version 4 algorithm for signing requests. It does so by implementing the `HTTPSigner.sign` interface, injecting an `Authorization` header into the resulting `HTTPRequest`. This signer is designed to be a standalone package that customers may use to sign the requests they send to AWS services.

It also implements the `string_to_sign` and `canonical_request` public interfaces. These are designed to be helper functions to customers that allow them to easily create important components of a signature version 4 signed request without doing so end to end.

This PR does NOT check for streaming requests which should be handled in the future. Additionally, checks on the body size are not performed. As there are significant performance implications for signing large request bodies, this may need to be done as a followup. In the mean time, a warning is added to logs indicating the potential for performance issues.

While calculating checksums, if the body is not rewindable/resettable, a buffer is created so the body can be read again in the future. After extensive testing, it was determined that the buffer should be a `BytesIO` object. While a list was found to have slightly lower peak memory usage, `BytesIO` periodically is able to perform optimizations on resident memory.


nateprewitt:
Resolving for inclusion at a later date.

PR #188, opened 2023-11-20
Upgrade pants, Makefile, CI, and scripts

haydenbaker:
#### pants
The latest version of pants is fairly different from the version we were using - much has changed.
- linters/fixers/etc options (in `pants.toml`) have changed a bit
- lockfile recommendations have changed, we'll use a single lockfile for the repo (python-default.lock)
  - `install_from_resolve` must be set to use plugins (like pytest-async, pytest-cov, etc.) for supported tools
    - Setting this means that pants uses the requirements given in the resolve
- pants no longer recommends checking in the pants runner script, and instead recommends using a `get-pants` installer script to pull in the latest binary
  - It's packaged as a PEX, and no longer requires a specific version of python (3.9) to be installed to run it
  - The binary will use the pants-version as specified by the `pants.toml` file
- Since we use make, `pants` will be installed to the repo root and can be ran using `./pants` (similar to before)
- The `init-pants` action is now used, it will set up pants for CI
  - The installation dir is not configurable in the action, so a wrapper script is copied in the repo-root as part of the `pants` make-recipe IF it is detected that we're in a CI environment (`CI` env-var is set to `true`)
- `pants fix` runs all the formatters/fixers, whereas `pants format` only runs the formatters 

#### make
As part of upgrading pants, several new recipes are added. Also, to make CI more ergonomic, I've transitioned our workflows to use `make` as much as possible.

#### examples
An example model (examples/weather) is added with configuration to generate a rest-json python client
  - With the smithy-cli installed, you can run `smithy build` to generate the client

#### isort issue
When upgrading pants, isort begins to behave strangely
 - An issue was opened for this: (https://github.com/pantsbuild/pants/issues/19516)
 - However, It is more than likely related to this: (https://github.com/pantsbuild/pants/issues/15069)
   - When executing separate commands for each package (instead of `::`), isort will correctly distinguish 1p and 3p import statements (i.e. pytest)
   - Since this is a bug, we'll need to use separate commands for each python package (until #15069 is fixed)


.github/workflows/ci.yml
```diff
@@ -17,18 +17,26 @@ jobs:
 
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v1
-      - name: Set up Python for Pants - 3.9
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.9
+        uses: actions/checkout@v3
```
JordonPhillips, line 20:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/lint.yml
```diff
@@ -8,17 +8,25 @@ on:
...
-      with:
-        python-version: 3.9
-    - name: Set up Python for linting - 3.11
-      uses: actions/setup-python@v4
-      with:
-        python-version: "3.11"
-    - name: Run pre-commit
-      uses: pre-commit/action@v3.0.0
+      - name: Checkout Repository
+        uses: actions/checkout@v1
```
JordonPhillips, line 15:
```suggestion
        uses: actions/checkout@v4
```

JordonPhillips (dismissed):

.github/workflows/lint.yml
haydenbaker: Whoops, I meant to update all the usage of checkout. Good catch!

JordonPhillips (approved):

PR #190, opened 2023-11-29
Add fault attribute to generated errors

haydenbaker:
*Description of changes:*
- Generated service errors (modeled with `@error`) should have the [fault (client or server)](https://smithy.io/2.0/spec/type-refinement-traits.html#error-trait) as an attribute


JordonPhillips (changes requested):
This needs to be in one of the base classes/ interfaces so that it can be seen generically by things like retry handlers.

JordonPhillips (dismissed):
Actually, we should be able to use protocols instead and it'll be fine. I've updated the PR to add fault to the unknown api error and make use of more code blocks. Also rebased it onto the split packages.

python-default.lock
haydenbaker, line 1: Whenever dependencies change, we need to make sure we upgrade and commit the lock-file. 
```
./pants generate-lockfiles
```

JordonPhillips (approved):

PR #195, opened 2023-12-14
Split up smithy-python

JordonPhillips:
This refactors smithy-python into two packages: smithy-core and smithy-http, separating concerns to bring things more in line with the split other sdks are moving towards. This allows non-http clients to exclude http features altogether. It also makes awscrt and aiohttp into optional dependencies.

Additionally, this attempts to break out async components into their own sections so that it will be more feasible to provide a sync client later on if necessary.

Another goal of this was to remove the weird `_private` directories we had going on.

Aside from adding in checks for the optional dependencies, the actual logic has not changed, only the locations of bits of code.

There is one bit that needs resolving in a follow-up, which is making the default http client configurable.


python-packages/smithy-http/BUILD
```diff
@@ -0,0 +1,31 @@
...
+# for now we need to duplicate things.
+python_requirements(
+    name="requirements",
+    source="requirements.txt",
+)
+
+# python_requirements(
+#     name="pyproject-requirements",
+#     source="pyproject.toml",
+# )
```
nateprewitt, line 31: Did we mean to delete this or was there a reason to keep it commented out?

python-packages/smithy-http/requirements.txt
```diff
@@ -0,0 +1,2 @@
+aiohttp>=3.8.6,<3.10.0
+awscrt>=0.15,<1.0
```
nateprewitt, line 2: Is this for pants/testing? Ideally we'd be able to just install `.[aiohttp]` or `.[awscrt]` during testing to avoid specifying these ranges twice.

nateprewitt (commented):
A few small comments/questions, but otherwise looks good. Let me know what you think and we can ship it.

python-packages/smithy-http/BUILD
JordonPhillips: Ah yeah, it needs to be removed. Essentially there was a problem where including the local dependencies in the pyproject would cause pants to fail because it tries to find it as an external dependency. Pants doesn't currently have the capability to make additions to pyproject files unfortunately, only `setup.py`

nateprewitt (approved):

Issue #196, opened 2024-01-07
Connection error when using non-standard port


rangerthegood:
There is a problem when trying to use a non standard port number in the Config. For example:

```
    client = EchoService(Config(endpoint_uri="http://127.0.0.1:12345"))
    response = await client.echo_message(EchoMessageInput(message="spam"))
    print(response.message)
```

You will get the error: `ranger_echo.errors.ServiceError: Cannot connect to host 127.0.0.1:80 ssl:default [Connect call failed ('127.0.0.1', 80)]`. The code is attempting to connect to port 80 versus 12345.

rangerthegood:
Fixed in #197

PR #197, opened 2024-01-07
[bug] Fix codgen bug where port number was ignored

rangerthegood:
*Issue #196 *

*Description of changes:*

This change passes through the port from the parsed endpoint uri when creating a URI object. Prior to this change the port field of the URI was always None which in turn uses the default port for the protocol which is an issue as pointed out in #196.


JordonPhillips (approved):
Nice!

PR #198, opened 2024-03-12
Upgrade to Python 3.12

JordonPhillips:
Upgrading to python 3.12 provides a ton of benefits both in authoring and in runtime performance. Of particular note is the new type param syntax, which will massively improve the readability of the relatively large amount of generic typing that we make use of. In addition, there's huge performance gains in asyncio as a whole.

Note that this PR doesn't actually update any of the python code to use the new syntax - that can come as a separate PR. It's also likely that we'll need to update mypy (or move to ruff) to be able to use the new syntax.

To facilitate this upgrade, pants had to also be upgraded to 2.19. Remarkably, it wasn't a whole day affair to do. The big change that had to be made was to add *everything* to the lockfiles so that they wouldn't try to pull some pre-compiled pex that doesn't support 3.12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.11"]
+        python-version: ["3.12"]
 
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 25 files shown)


nateprewitt (approved):
:shipit:

PR #199, opened 2024-03-13
Dependabot

JordonPhillips:
Adds a dependabot configuration and manually updates some dependencies that dependabot definitely can't handle by itself. One of those changes it to the smithy gradle plugin, whose latest version applies a formatter to smithy files, which is why those formatting changes are there.

I expect there to be a few dependabot prs after this, but they *should* be grouped.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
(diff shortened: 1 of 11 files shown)


nateprewitt (dismissed):
⛵

hpmellema (approved):

nateprewitt (approved):

PR #210, opened 2024-03-14
Migrate settings to immutable record

JordonPhillips:
This migrates the settings object to be an immutable record (same thing as a python dataclass), reducing any risk from passing it around.

This also removes the protocol setting. It was never actually possible to set since the `fromNode` method didn't check for it. But also, it's just as easy to strip out the protocols you don't want via projection transforms.

Resolving the protocol was dramatically simplified and bundled in to the protocol generator resolver.


nateprewitt (approved):
🚀

PR #216, opened 2024-03-20
Add logging

JordonPhillips:
Adds in logging. Initially there was an idea to make this configurable, but that seems excessive. The built in logger is quite good, and external logging libraries tend to integrate with it more or less seamlessly.


nateprewitt (approved):
:shipit:

PR #219, opened 2024-04-10
Remove as_dict and from_dict

JordonPhillips:
The as_dict and from_dict methods exist as a compatiblity layer for boto3. This unfortunately results in a ton of generated code that essentially functions as an additional protocol. Upcoming changes to way serde and documents are handled will provide a better path forward that does not require so much excess code.


hpmellema (approved):

nateprewitt (approved):
:shipit:

PR #221, opened 2024-04-10
Use dataclasses for structures

JordonPhillips:
This updates structure / union / error generators to use dataclasses. All of the issues we previously had in using them have been resolved by updates to Python and the removal of the generated boto3 compat code.

This depends on #219 


codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/DirectedPythonCodegen.java
```diff
@@ -148,28 +148,27 @@ class $L(SmithyException):
...
-
-                    class $1L($2T, Generic[T]):
+                    @dataclass
+                    class $1L($2T):
                         ""\"Base error for all API errors in the service.""\"
-                        code: T
-                        fault: Literal["client", "server"]
+                        code: ClassVar[str]
+                        fault: ClassVar[Literal["client", "server"]]
+
```
nateprewitt, line 160: Was there a reason for the newline break between code/fault and message?

codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
```diff
@@ -121,120 +134,83 @@ private void renderError() {
...
+            var target = model.expectShape(member.getTarget());
+            writer.putContext("quote", recursiveShapes.contains(target) ? "'" : "");
+
+            var memberName = symbolProvider.toMemberName(member);
+            writer.write("""
+                    $L: ${quote:L}$T${?nullable} | None${/nullable}${quote:L} \
+                    = ${^useField}${defaultValue:L}${/useField}\
+                    ${?useField}\
+                    field(${?sensitive}repr=False, ${/sensitive}${defaultKey:L}=${defaultValue:L})\
+                    ${/useField}""", memberName, symbolProvider.toSymbol(member));
```
nateprewitt, line 210: This looks pretty intense, do you have an example of the generated code? Ideally we're no longer using `\` line continuations if possible.

nateprewitt (commented):
Two minor comments/questions, otherwise looks good to me!

codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
JordonPhillips: You can't really get away with not using the continuations without making it even harder to read. This will produce things like:

```python
@dataclass(kw_only=True)
class Foo:
    normal: int
    optional: int | None = None

    recursive: "Foo"
    optional_recursive: "Foo | None" = None

    default: int = 0
    default_list: list[str] = field(default_factory=list)
    default_map: dict[str, str] = field(default_factory=map)

    sensitive: int = field(repr=False)
    sensitive_default: int = field(default=0, repr=False)
    sensitive_default_collection: list[str] = field(default_factory=list, repr=False)
```

It's messy because all these things can combine.

nateprewitt (approved):

PR #225, opened 2024-04-18
Pyright

JordonPhillips:
We're using python 3.12, but mypy doesn't support the new typing features yet. This is pretty unfortunate because that's the whole reason we switched to 3.12. This PR changes the type checker we use to pyright instead, which does support those features.

It also addresses a rather large number of issues that pyright caught which mypy didn't.

Note that there's a few global suppressions being put into the generated configs. Those are all related to things going on in the serde code, and will be removed once we've switched to the codec model.

### DIFF
--- a/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -395,10 +395,6 @@ final class ClientGenerator implements Runnable {
             writer.addImport("smithy_core", "URI");
             writer.write("""
                         # Step 7f: Invoke endpoint_resolver.resolve_endpoint
-                        if config.endpoint_resolver is None:
-                            raise $1T(
-                                "No endpoint_resolver found on the operation config."
-                            )
                         if config.endpoint_uri is None:
                             raise $1T(
                                 "No endpoint_uri found on the operation config."
@@ -673,7 +669,7 @@ final class ClientGenerator implements Runnable {
                 }
             }
             writer.write("""
-                operation_plugins = [
+                operation_plugins: list[Plugin] = [
                     $C
                 ]
                 if plugins:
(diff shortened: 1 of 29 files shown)


codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -395,10 +395,6 @@ async def _handle_attempt(
             writer.addImport("smithy_core", "URI");
             writer.write("""
                         # Step 7f: Invoke endpoint_resolver.resolve_endpoint
-                        if config.endpoint_resolver is None:
-                            raise $1T(
-                                "No endpoint_resolver found on the operation config."
-                            )
```
nateprewitt, line 401: Why did this check get removed? Is this one of the suppressions you referred to in the overview?

codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -986,6 +986,11 @@ private void deserializeIndividualHeaders(
                     """, locationName, memberName, targetHandler);
                 }
             }
+            writer.write("""
+                    case _:
+                        pass
+
+                    """);
```
nateprewitt, line 993: It may be useful to note we're treating anything unknown as a no-op, this isn't the most obvious at first glance.

python-packages/smithy-core/tests/unit/test_types.py
```diff
@@ -1,5 +1,6 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  SPDX-License-Identifier: Apache-2.0
+# pyright: reportPrivateUsage=false
```
nateprewitt, line 3: Does this only apply to this file for checks? It also seems odd to have this connected to the Copyright block. Is the formatter pushing the two together?

JordonPhillips: Yeah, this is a file-wide check. I can separate by an empty line and it should be fine.

nateprewitt (approved):

PR #226, opened 2024-04-19
Mix sync and async bodies

JordonPhillips:
This updates the request/response interfaces to mix sync and ascync bodies, This is similar to what httpx does. It makes it easier to write generic serializers / deserializers, which are primarily going to be synchronous anyway. We'll eventually be explicitly extracting streaming inputs / outputs to handle them separately.


nateprewitt (dismissed):
One quick typo, otherwise looks good.

nateprewitt (approved):

PR #227, opened 2024-04-19
Add basic schema and shape classes

JordonPhillips:
This adds the schema class, which will be used to provide some information from the model at runtime. It additionally adds in some shape / trait classes that are needed to fill it out.

### DIFF
--- a/python-packages/smithy-core/smithy_core/schemas.py
+++ b/python-packages/smithy-core/smithy_core/schemas.py
@@ -1,0 +1,195 @@
+from dataclasses import dataclass, field
+from typing import (
+    TYPE_CHECKING,
+    NotRequired,
+    Protocol,
+    Required,
+    Self,
+    TypedDict,
+    runtime_checkable,
+)
+
+from .exceptions import ExpectationNotMetException, SmithyException
+from .shapes import ShapeID, ShapeType
+
+if TYPE_CHECKING:
+    from .traits import Trait
+
+
+@dataclass(kw_only=True, frozen=True, init=False)
+class Schema:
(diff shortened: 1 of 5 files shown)


mtdowling (dismissed):

python-packages/smithy-core/smithy_core/schemas.py
```diff
@@ -0,0 +1,195 @@
...
+                    member_target=members[k]["target"],
+                    member_index=i,
+                )
+        result = cls(
+            id=id,
+            type=type,
+            traits=traits,
+            members=struct_members,
+        )
+        print(repr(result))
```
nateprewitt, line 171:
```suggestion

```

python-packages/smithy-core/smithy_core/shapes.py
```diff
@@ -0,0 +1,128 @@
...
+    MAP = 17
+    STRUCTURE = 18
+    UNION = 19
+
+    MEMBER = 20
+
+    # We won't acutally be using these, probably
+    # SERVICE = 21
+    # RESOURCE = 22
+    # OPERATION = 23
```
nateprewitt, line 128: Just for my context, what are these for? Delineating between a service model, resource, and individual operations?

JordonPhillips: Those are different smithy shape types that aren't currently involved in serialization. It may be necessary to use them in the future. We might end up having to make an operation schema, for example, if we can't embed that information in another way.

nateprewitt (approved):

PR #230, opened 2024-04-24
Add length and repr to fields

nateprewitt:
This PR adds some minor usability dunders to the `Fields` class to allow us to verify the total number of `Field` entries in a `Fields` instance and represent those values in the `__repr__`.

### DIFF
--- a/python-packages/smithy-http/smithy_http/__init__.py
+++ b/python-packages/smithy-http/smithy_http/__init__.py
@@ -172,6 +172,12 @@ class Fields(interfaces.Fields):
     def __iter__(self) -> Iterator[interfaces.Field]:
         yield from self.entries.values()
 
+    def __len__(self) -> int:
+        return len(self.entries)
+
+    def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


python-packages/smithy-http/smithy_http/__init__.py
```diff
@@ -172,6 +172,12 @@ def __eq__(self, other: object) -> bool:
     def __iter__(self) -> Iterator[interfaces.Field]:
         yield from self.entries.values()
 
+    def __len__(self) -> int:
+        return len(self.entries)
+
+    def __repr__(self) -> str:
```
jonathan343, line 178: I think we should add a test for this similar to the one you added for `__len__`

jonathan343 (changes requested):
I think we should add a test for `__repr__` this similar to the one you added for `__len__`

jonathan343 (approved):
Ah I see, I just saw that the `__repr__` method for `Field` isn't tested. Thanks for the clarification. Approved

PR #231, opened 2024-04-24
Add tests for Field and Fields __repr__

nateprewitt:
This PR will add some tests around the `Field` and `Fields` classes to ensure we are returning the expected representation.


jonathan343 (approved):
:shipit:

PR #232, opened 2024-04-24
Add __contains__ support to Fields

nateprewitt:
Add `__contains__` to Fields to properly support the `in` Python keyword when searching the collection.

### DIFF
--- a/python-packages/smithy-http/smithy_http/__init__.py
+++ b/python-packages/smithy-http/smithy_http/__init__.py
@@ -178,6 +178,9 @@ class Fields(interfaces.Fields):
     def __repr__(self) -> str:
         return f"Fields({self.entries})"
 
+    def __contains__(self, key: str) -> bool:
+        return key in self.entries
+
 
 def quote_and_escape_field_value(value: str) -> str:
     """Escapes and quotes a single :class:`Field` value if necessary.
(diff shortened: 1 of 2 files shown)


python-packages/smithy-http/smithy_http/__init__.py
```diff
@@ -178,6 +178,9 @@ def __len__(self) -> int:
     def __repr__(self) -> str:
         return f"Fields({self.entries})"
 
+    def __contains__(self, key: str) -> bool:
+        return key in self.entries
```
jonathan343, line 182: Shouldn't we be normalizing the key (using `_normalize_field_name`) before checking if it exists `entries`? Similar to how we do it for the set, get, and remove methods?

python-packages/smithy-http/tests/unit/test_fields.py
```diff
@@ -222,3 +222,17 @@ def test_fields_length_value(fields: Fields, expected_length: int) -> None:
...
+
+
+@pytest.mark.parametrize(
+    "fields,key,contained",
+    [
+        (Fields(), "bad_key", False),
+        (Fields([Field(name="fname1")]), "fname1", True),
+        (Fields([Field(name="fname2")]), "fname1", False),
+        (Fields([Field(name="f1"), Field(name="f2")]), "f1", True),
+        (Fields([Field(name="f1"), Field(name="f2")]), "f3", False),
```
jonathan343, line 234: If we decided to normalize the key parameter, we should add a test with an uppercase input key like below:
```
(Fields([Field(name="f1"), Field(name="f2")]), "F1", False),
```
If not, ignore this comment.

nateprewitt: Yep good call out!

jonathan343 (approved):

PR #233, opened 2024-04-26
Update explicit set/get/remove functions to use native dunder methods

nateprewitt:
This PR proposes simplifying our `Fields` interface to use more idiomatic methods for accessing, setting, and removing `Field` entries. The original implementation was strictly following the Cross-SDK specification but creates an unintuitive user experience for Python developers. We could also keep the existing methods and alias them to the dunder implementations, but it seems like clutter and confusing imo.


jonathan343 (approved):
:shipit:

PR #234, opened 2024-04-29
Update Smithy and use typed properties.

JordonPhillips:
This switches all symbol property usage to use typed properties, which were introduced in Smithy 1.48.0. This allows us to more easily document these values and be confident in their typing, as the type of the property is stored in its definition now.

I'd been wanting to store these properties documented somewhere for a while, and this provided a good opportunity to do that.

To do this, I had to update to the latest version of Smithy. While I was at it, I also updated the gradle version.


nateprewitt (approved):
:shipit:

PR #245, opened 2024-05-17
Add serailizer interfaces and document implementations of them

JordonPhillips:
This adds interfaces for shape serializers, and implementations of those serializers for documents. It also fixes a pair of bugs in Document that were uncovered while working on tests.

Serializing in this way instead of generating serializers gives as a few major benefits:

* Protocols will be able to be changed at runtime
* Protocols will be implemented entirely in Python, making them easier to understand, modify, and write for external users.
* The amount of generated code is dramatically reduced. On average, package size will also be significantly reduced. The generated code for schemas is about half the size of the generated code for serializers. All the library code to implement protocols is of course added on top, but those will be shared between services, so the more services a customer uses, the more they'll benefit from this setup.

A follow-up PR will add deserialization support along similar principles.

### DIFF
--- a/python-packages/smithy-core/smithy_core/documents.py
+++ b/python-packages/smithy-core/smithy_core/documents.py
@@ -244,7 +244,7 @@ class Document:
         :param shape: The shape to convert to a document.
         :returns: A Document representation of the given shape.
         """
-        serializer = DocumentSerializer()
+        serializer = _DocumentSerializer()
         shape.serialize(serializer=serializer)
         serializer.flush()
         return serializer.expect_result()
@@ -312,7 +312,7 @@ class Document:
         return self.as_value() == other.as_value()
 
 
-class DocumentSerializer(ShapeSerializer):
+class _DocumentSerializer(ShapeSerializer):
     """Serializes shapes into document representations."""
 
     result: Document | None = None
@@ -396,7 +396,7 @@ class DocumentSerializer(ShapeSerializer):
 
 
 class _DocumentStructSerializer(InterceptingSerializer):
-    _delegate = DocumentSerializer()
+    _delegate = _DocumentSerializer()
     _result: Document
 
     def __init__(self, schema: "Schema") -> None:
@@ -415,7 +415,7 @@ class _DocumentStructSerializer(InterceptingSerializer):
 
 
 class _DocumentListSerializer(InterceptingSerializer):
-    _delegate = DocumentSerializer()
+    _delegate = _DocumentSerializer()
     _result: list[Document]
     _schema: "Schema"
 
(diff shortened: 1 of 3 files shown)


python-packages/smithy-core/smithy_core/serializers.py
```diff
@@ -0,0 +1,335 @@
...
+
+    @contextmanager
+    def begin_struct(self, schema: "Schema") -> Iterator[ShapeSerializer]:
+        delegate = self.before(schema)
+
+        try:
+            with delegate.begin_struct(schema) as s:
+                yield s
+        finally:
+            self.after(schema)
```
mtdowling, line 235: I'm dusting off my Python here... but does this work like finally in Java where it's run whether there's an error or not? If it is run just before throwing whatever's caught in the try statement, is that what you really want? I think a lot of times these after statements would do stuff like write a comma or something to an output buffer. Do you want to do that when there's an error?

python-packages/smithy-core/smithy_core/documents.py
```diff
@@ -284,3 +310,137 @@ def __eq__(self, other: object) -> bool:
...
+        return self.result
+
+    @override
+    @contextmanager
+    def begin_struct(self, schema: "Schema") -> Iterator[ShapeSerializer]:
+        delegate = _DocumentStructSerializer(schema)
+        try:
+            yield delegate
+        finally:
+            self.result = delegate.result
```
mtdowling, line 335: Why set the result when an error occurs?

python-packages/smithy-core/smithy_core/documents.py
```diff
@@ -284,3 +310,137 @@ def __eq__(self, other: object) -> bool:
         # JsonDocument, cares about the schema then it's responsible for transforming
         # the output of `as_value` to the canonical form.
         return self.as_value() == other.as_value()
+
+
+class DocumentSerializer(ShapeSerializer):
```
mtdowling, line 315: Is there any way to prevent access to this directly and hide it all behind a method that internally uses this? That way you can completely change how it works in the future and even remove this code if you really wanted to. Or I guess you just add a "_" to the name to say use at your own risk.

python-packages/smithy-core/smithy_core/serializers.py
JordonPhillips: Good call. Switched to try/except/else instead.

python-packages/smithy-core/smithy_core/documents.py
JordonPhillips: In this case I don't think it actually matters, but i switched it to try/except/else as well.

python-packages/smithy-core/smithy_core/documents.py
JordonPhillips: No way to prevent access I'm afraid, but I did prefix it

mtdowling (approved):

nateprewitt (approved):
:shipit:

PR #246, opened 2024-05-17
Add deserializer interfaces and document implementation of them

JordonPhillips:
Depends on #245 

This adds interfaces for shape deserializers, as well as document implementations for those interfaces. See #245 for additional context on why this approach is being taken.

Another advantage with both document serializers and deserializers available is a sort of compatibility layer with botocore:

```python
botocore_style_input: dict[str, Any] = {"foo": "bar"}
smithy_style_input: InputShape = Document(botocore_style).as_shape(InputShape)

smithy_style_output: OutputShape = OutputShape(foo="bar")
botocore_style_output: dict[str, Any] = Document.from_shape(smithy_style_output).as_value()
```

There will of course be some cases where this won't work, mostly regarding cases where botocore has customized the type or name of a parameter, but for the vast majority of inputs and outputs, this should work. We could write a compatibility package that addresses those extra cases if needed.

In this way people could migrate gradually without much difficulty.

### DIFF
--- a/python-packages/smithy-core/tests/unit/test_documents.py
+++ b/python-packages/smithy-core/tests/unit/test_documents.py
@@ -6,7 +6,13 @@ from typing import Any, cast
 
 import pytest
 
-from smithy_core.documents import Document, DocumentValue, _DocumentSerializer
+from smithy_core.deserializers import ShapeDeserializer
+from smithy_core.documents import (
+    Document,
+    DocumentDeserializer,
+    _DocumentSerializer,
+    DocumentValue,
+)
 from smithy_core.exceptions import ExpectationNotMetException
 from smithy_core.prelude import (
     BIG_DECIMAL,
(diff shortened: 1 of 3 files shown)


python-packages/smithy-core/tests/unit/test_documents.py
```diff
@@ -560,6 +566,66 @@ def serialize_members(self, serializer: ShapeSerializer) -> None:
...
 
+    @classmethod
+    def deserialize(cls, deserializer: ShapeDeserializer) -> "DocumentSerdeShape":
+        kwargs: dict[str, Any] = {}
+
+        def _consumer(schema: Schema, de: ShapeDeserializer) -> None:
+            match schema.expect_member_index():
+                case 0:
+                    kwargs["boolean_member"] = de.read_boolean(
+                        SCHEMA.members["booleanMember"]
```
mtdowling, line 577: This is a test case, but in actual generated types, generating member schemas so that you can reference them directly from deserializers would avoid the overhead of hashing and getting the schema from a hashmap. Looking up schemas in the hashmap is kind of canceling out the benefit of deserializing based on the generated member index. I think you'd want those to be `_` or even `__` prefixed.

JordonPhillips: In Python, property access is just a dict access, so there's no real difference. I'll be experimenting with slots to speed that up, but it'll add a good bit of generated code

mtdowling (dismissed):

nateprewitt (dismissed):

JordonPhillips:
Had to rebase

nateprewitt (approved):

PR #247, opened 2024-05-17
Add smithy-json package

JordonPhillips:
This adds an empty package to hold json serialization / deserialization code.


python-packages/smithy-json/pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "smithy_json"
+version = "0.0.1"
+description = "JSON serialization and deserialization for Smithy clients."
```
nateprewitt, line 8: Are we lumping json/rest-json into this package or is this specifically for RPC? I'm just curious how we're planning on partitioning protocols going forward.

JordonPhillips: This is just generic json support, not a protocol. Protocols will have their own packages (though aws json 1.0 and 1.1 should probably share a package since there's only minimal differences.

nateprewitt (approved):

PR #248, opened 2024-05-21
Add timestamp format enum

JordonPhillips:
This adds a `TimestampFormat` enum  that has helper methods on it for serialization and deserialization.

### DIFF
--- a/python-packages/smithy-core/smithy_core/types.py
+++ b/python-packages/smithy-core/smithy_core/types.py
@@ -2,7 +2,19 @@
 #  SPDX-License-Identifier: Apache-2.0
 import json
 from collections.abc import Mapping, Sequence
+from datetime import datetime
+from email.utils import format_datetime, parsedate_to_datetime
+from enum import Enum
 from typing import Any, TypeAlias
+
+from .exceptions import ExpectationNotMetException
+from .utils import (
+    ensure_utc,
+    epoch_seconds_to_datetime,
+    expect_type,
+    serialize_epoch_seconds,
+    serialize_rfc3339,
+)
 
 Document: TypeAlias = (
     Mapping[str, "Document"] | Sequence["Document"] | str | int | float | bool | None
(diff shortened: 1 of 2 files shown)


python-packages/smithy-core/smithy_core/types.py
```diff
@@ -41,3 +53,61 @@ def from_json(j: Any) -> "JsonBlob":
...
+    for more details.
+    """
+
+    DATE_TIME = "date-time"
+    """RFC3339 section 5.6 datetime with optional millisecond precision but no UTC
+    offset."""
+
+    HTTP_DATE = "http-date"
+    """An HTTP date as defined by the IMF-fixdate production in RFC 7231 section
+    7.1.1.1."""
```
nateprewitt, line 71: Should we be using the latest standard here (RFC 9110 5.6.7) or is there a notable difference in the older RFC we use?

python-packages/smithy-core/smithy_core/types.py
```diff
@@ -2,8 +2,20 @@
 #  SPDX-License-Identifier: Apache-2.0
 import json
 from collections.abc import Mapping, Sequence
+from datetime import datetime
+from email.utils import format_datetime, parsedate_to_datetime
```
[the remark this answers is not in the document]
JordonPhillips: If this were some dependency that we were pulling from I'd agree, but I don't think that sort of argument holds much weight when this is in the standard library. As it is, these are already being used in the generated serializer / deserializer implementations and are passing protocol tests.

python-packages/smithy-core/smithy_core/types.py
JordonPhillips: I don't think there's any difference, but I can update the reference here and in the smithy docs.

JordonPhillips: https://github.com/smithy-lang/smithy/pull/2298/

nateprewitt (approved):

PR #253, opened 2024-05-28
Add serailze methods to documents

JordonPhillips:
This adds a serialize method to documents and fixes a few small issues that cropped up when testing those changes. Notably an is_null method was added since it became apparent that otherwise there's no way to tell. More importantly though, document members no longer resolve to their target shapes, but rather the proper member shapes. This necessitated some changes to setitem. Lastly, some improvements were added to shape type inference that leave the result as document only when the value is null.

### DIFF
--- a/python-packages/smithy-core/smithy_core/documents.py
+++ b/python-packages/smithy-core/smithy_core/documents.py
@@ -5,7 +5,7 @@ from decimal import Decimal
 from typing import TypeGuard, override
 
 from .deserializers import DeserializeableShape, ShapeDeserializer
-from .exceptions import ExpectationNotMetException
+from .exceptions import ExpectationNotMetException, SmithyException
 from .schemas import Schema
 from .serializers import (
     InterceptingSerializer,
@@ -86,7 +86,7 @@ class Document:
                 self._raw_value = value
             else:
                 self._value = value  # type: ignore
-        elif isinstance(value, Sequence):
+        elif isinstance(value, Sequence) and not isinstance(value, str | bytes):
             if self._is_raw_list(value):
                 self._raw_value = value
             else:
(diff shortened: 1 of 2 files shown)


JordonPhillips:
Hold up on reviewing this, I need to make changes

python-packages/smithy-core/smithy_core/documents.py
```diff
@@ -86,20 +86,19 @@ def __init__(
                 self._raw_value = value
             else:
                 self._value = value  # type: ignore
-        elif isinstance(value, Sequence):
+        elif isinstance(value, Sequence) and not isinstance(value, str | bytes):
```
nateprewitt, line 89: This is the first time I've seen this. Was this a change with typing that enabled usage of this syntax vs a tuple of types?

python-packages/smithy-core/smithy_core/documents.py
```diff
@@ -212,17 +217,15 @@ def as_map(self) -> dict[str, "Document"]:
         )
 
     def _wrap_map(self, value: Mapping[str, DocumentValue]) -> dict[str, "Document"]:
-        if self._schema.shape_type is not ShapeType.STRUCTURE:
+        if self._schema.shape_type not in [ShapeType.STRUCTURE, ShapeType.UNION]:
```
nateprewitt, line 220: nit
```suggestion
        if self._schema.shape_type not in (ShapeType.STRUCTURE, ShapeType.UNION):
```

nateprewitt (commented):
Question and a minor nit, otherwise looks good to me.

python-packages/smithy-core/smithy_core/documents.py
JordonPhillips: Yeah, [PEP604](https://peps.python.org/pep-0604/), which added the union syntax, also made it useable in isinstance checks.

nateprewitt (approved):

PR #260, opened 2024-06-11
Generate serializer implementations

JordonPhillips:
This adds generated implementations for structures and unions to for the SerializeableStruct protocol.

Currently all http bound members are excluded so that the JSON codec can be integrated relatively quickly. They'll be added back in later when full protocol support exists.

I've also started putting the "Generator" classes into their own package to keep things tidy. Eventually the others will be moved into it.


JordonPhillips:
Closing this while i re-work it to remove duplication of list / map blocks

nateprewitt (approved):
⛵

PR #272, opened 2024-06-28
Add read_optional method to deserializers

JordonPhillips:
This adds a method to deserializers to read nulllable values from the source. This is primarily useful for sparse lists and maps.

### DIFF
--- a/python-packages/smithy-core/smithy_core/deserializers.py
+++ b/python-packages/smithy-core/smithy_core/deserializers.py
@@ -60,6 +60,19 @@ class ShapeDeserializer(Protocol):
         """
         ...
 
+    def read_optional[
+        T
+    ](self, schema: "Schema", optional: Callable[["Schema"], T]) -> T | None:
+        """Read an optional value from the underlying data.
+
+        This is intended to be used with sparse lists or maps.
+
+        :param schema: The shape's schema.
+        :param optional: A callable that takes a schema and reads a non-nullable value
+            from the underlying data.
+        """
+        ...
+
     def read_boolean(self, schema: "Schema") -> bool:
         """Read a boolean value from the underlying data.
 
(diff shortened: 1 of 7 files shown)


python-packages/smithy-core/smithy_core/deserializers.py
```diff
@@ -60,6 +60,19 @@ def read_null(self, schema: "Schema") -> None:
         """
         ...
 
+    def read_optional[
+        T
+    ](self, schema: "Schema", optional: Callable[["Schema"], T]) -> T | None:
```
JordonPhillips, line 65: I really hate the way black formats this

nateprewitt: Ruff does a better job with these cases. If we manually rewrite this as the below does `black` still mess it up, or is it left alone?

```suggestion
    def read_optional[T](
        self, schema: "Schema", optional: Callable[["Schema"], T]
    ) -> T | None:
```

mtdowling (dismissed):

python-packages/smithy-core/tests/unit/test_documents.py
```diff
@@ -540,12 +557,14 @@ def test_is_none():
...
         "mapMember": {"target": STRING_MAP_SCHEMA, "index": 9},
+        "sparseListMember": {"target": SPARSE_STRING_LIST_SCHEMA, "index": 11},
+        "sparseMapMember": {"target": SPARSE_STRING_MAP_SCHEMA, "index": 12},
     },
 )
 SCHEMA.members["structMember"] = Schema.member(
     id=SCHEMA.id.with_member("structMember"),
     target=SCHEMA,
-    index=len(SCHEMA.members),
+    index=10,
```
nateprewitt, line 567: Is it worth adding a quick comment on line 560 above noting index 10 is added below?

nateprewitt (dismissed):
One minor comment but otherwise looks good!

nateprewitt (approved):

PR #273, opened 2024-07-01
Generate schema-based deserializers

JordonPhillips:
This adds code-generated deserializers based on schemas. It also fixes some member filtering that was previously added but not working properly.


codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/SetupGenerator.java
```diff
@@ -111,6 +111,7 @@ private static void writePyproject(
                     reportUnusedFunction = false
                     reportUnusedVariable = false
                     reportUnnecessaryComparison = false
+                    reportUnusedClass = false
```
JordonPhillips, line 114: This is needed because union deserializers are classes, event stream deserializers are never called as they are filtered out currently.

nateprewitt (approved):

PR #277, opened 2024-07-15
Use JSONCodec for json pyaload serialization

JordonPhillips:
This updates payload serialization for RestJson1 to use codecs. Note that it's currently creating a new codec every time, but eventually that will be stored on the config.


nateprewitt (approved):
🚢

PR #294, opened 2024-08-15
Add empty smithy-event-stream package

JordonPhillips:
This adds an empty package to contain the event stream implementation.

### DIFF
--- a/python-packages/smithy-event-stream/pyproject.toml
+++ b/python-packages/smithy-event-stream/pyproject.toml
@@ -1,0 +1,49 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "smithy_event_stream"
+version = "0.0.1"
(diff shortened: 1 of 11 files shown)


python-packages/smithy-event-stream/pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm", "wheel"]
+build-backend = "setuptools.build_meta"
```
nateprewitt, line 3: Are we using setuptools in the other projects? Not part of this PR but as a note, we should definitely move to [`hatchling`](https://hatch.pypa.io/latest/) over setuptools.

nateprewitt (approved):

python-packages/smithy-event-stream/pyproject.toml
JordonPhillips: We are - but we can switch. I have no strong opinions on the topic. Lets do it as one big changeover though

PR #313, opened 2024-10-14
Event stream interfaces and HTTP streaming support

JordonPhillips:
This adds a `stream` operation to the HTTP client interface which opens a bidirectional-capable stream. This operation is implmented for the CRT, but not for aiohttp because it doesn't have bidirectional support.

The reason that this requires an entirely new operation is that a caller must be capable of immediately sending request frames without waiting for *any kind of response* and be capable of sending them at the same time as they are receiving server events. The `send` method kinda forces them to either wait for the initial responses or force the caller to guess when it's okay to start sending request frames. Buffering can be done of course, but that starts to pile on memory.

Having the implementation return an input stream also has the benefit of letting the underlying client create their own writer, which the CRT is tentatively planning to do in leu of wrapping a BytesIO.

Another potentially questionable decision here is the banning of setting the HTTP request body for streaming requests. An alternative would be to use that as initial data. But if that initial data is itself an async reader, you're potentially having to buffer a lot of data while you wait for the initial data to empty out.

This also adds in the interfaces for streaming that operations will use. Initially I was planning on having the HTTP interfaces also implement these event stream interfaces, but that would require passing codecs down through the HTTP clients where they don't necessarily belong.

### DIFF
--- a/python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
+++ b/python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
@@ -1,0 +1,86 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Protocol, Self, Any
+
+
+class InputEventStream[E](Protocol):
+    """Asynchronously sends events to a service."""
+
+    async def send(self, event: E) -> None:
+        """Sends an event to the service.
+        
+        :param event: The event to send.
+        """
+        ...
+
+    def close(self) -> None:
+        """Closes the event stream."""
+        ...
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any):
+        self.close()
+
+
+class OutputEventStream[E](Protocol):
+    """Asynchronously receives events from a service."""
+
+    async def receive(self) -> E | None:
+        """Receive a single event from the service.
+        
+        :returns: An event or None. None indicates that no more events will be sent by
+            the service.
+        """
+        ...
+
+    def close(self) -> None:
+        """Closes the event stream."""
+        ...
+
+    async def __anext__(self) -> E:
+        result = await self.receive()
+        if result is None:
+            self.close()
+            raise StopAsyncIteration
+        return result
+
+    def __aiter__(self) -> Self:
+        return self
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any):
+        self.close()
+
+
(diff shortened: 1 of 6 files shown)


python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
```diff
@@ -0,0 +1,90 @@
...
+
+    input_stream: I
+    _output_stream: O | None = None
+    _response: R | None = None
+
+    async def await_output(self) -> tuple[R, O]:
+        if self._response is not None:
+            self._response, self._output_stream = await self._await_output()
+
+        return self._response, self._output_stream  # type: ignore
```
nateprewitt, line 75: This chunk I don't know if I fully understand.

We have to establish our input stream outside of the event stream which I'm assuming happens in our `stream` implementation? We don't expect end users to be invoking this?

For the return types, what's in the initial response object vs what we read off the OutputEventStream? Is it just confirmation of the connection and then we're expected to read from the stream?

JordonPhillips: Say you have an operation that returns an event stream. It might have a definition like:

```python
@dataclass(kw_only=True)
class StreamChatsInput:
    room_id: str
    username: str


@dataclass(kw_only=True)
class StreamChatsOutput:
    inactivity_timeout: int
    active_users: list[str]
    

type Chats = ChatMessage
    | ChatReaction
    | ChatResponse
    | ChatLogin
    | ChatLogout
    | ChatKeepalive
    | ChatUnknown


async def stream_chats(self, input: StreamChatsInput) -> EventStream[
    InputEventStream[Chats], OutputEventStream[Chats], StreamChatsOutput]: ...
```

Before this operation returns, it MUST:

* Establish the input stream (`InputEventStream[Chats]`)
* Send the initial request (the contents of `StreamChatsInput`)

The input stream is therefore returned directly by the operation, nested inside the `EventStream` object. Customers may begin sending `Chats` right away.

It MUST NOT:

* Wait for the output stream to be established (`OutputEventStream[Chats]`)
* Wait for the initial response (the contents of `StreamChatsOutput`)

Since the output stream may not have been established and the initial response may not have been received when the operation returns, they are both deffered and must be awaited. The block you highlighted simply caches them so that subequent calls don't have to actually perform any operations on the underlying streams.

As far as what's in the initial response and initial request, it's everything on the top-level input/output that isn't the event stream. In the example above, the initial request dictates what user is connecting and what chat room they're connecting to. Initial messages MUST always be sent first, so the service can rely on that.

Similarly, the initial response MUST be the first message sent in response. This information is likely necessary to understand future messages, or to provide metadata about the stream as established by the service. In the example above it contains information about currently active users and the max amount of time between messages before the service will terminate a connection. This is important information, but an application need not wait on it before sending messages. For example, a chat bot that only establishes a connection to send a single message and then leaves would not necessarily care about that information and thus does not need to wait for the output stream to be established at all.

PR #318, opened 2024-10-17
Add types and code generation for event stream operation signatures

JordonPhillips:
This adds in the interfaces for event streams and updates code generation to use them.

Example generated operation:

```python
async def stream_messages(
    self,
    input: StreamMessagesInput,
    plugins: list[Plugin] | None = None,
) -> EventStream[
    InputEventStream[MessageStream],
    OutputEventStream[MessageStream],
    StreamMessagesOutput,
]:
    raise NotImplementedError()
```

These were part of separate prs that included http context, which has been removed. Doc strings have also been majorly beefed up. Example useage from the beefed up docs:

```python
async def main():
    client = ChatClient()
    input = StreamMessagesInput(chat_room="aws-python-sdk", username="hunter7")

    async with client.stream_messages(input=input) as stream:
        stream.input_stream.send(MessageStreamMessage("Chat logger starting up."))
        response_handler = handle_output(stream)
        stream.input_stream.send(MessageStreamMessage("Chat logger active."))
        await response_handler

    async def handle_output(stream: EventStream) -> None:
        _, output_stream = await stream.await_output()
        async for event in output_stream:
            match event:
                case MessageStreamMessage():
                    print(event.value)
                case MessageStreamShutdown():
                    return
                case _:
                    stream.input_stream.send(
                        MessageStreamMessage("Unknown message type received. Shutting down.")
                    )
                    return
```

### DIFF
--- a/python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
+++ b/python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
@@ -1,0 +1,177 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Any, Protocol, Self
+
+from smithy_core.deserializers import DeserializeableShape
+from smithy_core.serializers import SerializeableShape
+
+
+class InputEventStream[E: SerializeableShape](Protocol):
+    """Asynchronously sends events to a service.
+
+    This may be used as a context manager to ensure the stream is closed before exiting.
+    """
+
+    async def send(self, event: E) -> None:
+        """Sends an event to the service.
+
+        :param event: The event to send.
+        """
+        ...
+
+    def close(self) -> None:
+        """Closes the event stream."""
+        ...
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any):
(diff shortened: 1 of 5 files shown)


JordonPhillips:
I also had to add suppressions for new protocol tests that assert http-binding behaviors. We'll be re-writing all that binding code soon, so there's little value in fixing it now.

JordonPhillips:
I split the suppressions out into another pr since it was blocking everything

python-packages/smithy-event-stream/smithy_event_stream/aio/interfaces.py
```diff
@@ -0,0 +1,177 @@
...
+
+    This may include context necessary to interpret output events or prepare
+    input events. It will always be available before any events.
+    """
+
+    async def await_output(self) -> tuple[R, O]:
+        """Await the operation's output.
+
+        The EventStream will be returned as soon as the input stream is ready to
+        receive events, which may be before the intitial response has been received
```
nateprewitt, line 134:
```suggestion
        receive events, which may be before the initial response has been received
```

nateprewitt (changes requested):
Couple typos, otherwise looks good.

JordonPhillips:
No idea why codecov ate it

nateprewitt (approved):
Looks like a transient 503 for CodeCov, succeeded on retry.

PR #319, opened 2024-10-17
Write CRT data asynchronously

JordonPhillips:
This updates the CRT bindings to write data to its ByteIO body asynchronously rather than buffering it all into memory when making a call. This should enable us to support event streaming, and also prevent us from buffering a huge amount of data into memory, even for synchronous bodies of uncertain size.

This should also allow us to get rid of those consume_body utility functions that I detest. But one thing at a time.

Also, we really need to work on standing up a local http2 server for testing. Integ tests will be nice too and all but I'm nevertheless nervous about how hard this is to test.

### DIFF
--- a/python-packages/smithy-http/smithy_http/aio/crt.py
+++ b/python-packages/smithy-http/smithy_http/aio/crt.py
@@ -298,7 +298,7 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
 
         body = request.body
         if isinstance(body, bytes | bytearray):
-            # If the body is already directly in memory, wrap in in a BytesIO to hand
+            # If the body is already directly in memory, wrap in a BytesIO to hand
             # off to CRT.
             crt_body = BytesIO(body)
         else:
@@ -311,7 +311,7 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
                 # with read methods will be read in chunks so as not to exhaust memory.
                 body = AsyncBytesReader(body)
 
-            # Start the read task int the background.
+            # Start the read task in the background.
             read_task = asyncio.create_task(self._consume_body_async(body, crt_body))
 
             # Keep track of the read task so that it doesn't get garbage colllected,
--- a/python-packages/smithy-http/smithy_http/aio/crt.py
+++ b/python-packages/smithy-http/smithy_http/aio/crt.py
@@ -25,6 +25,7 @@ except ImportError:
     HAS_CRT = False  # type: ignore
 
 from smithy_core import interfaces as core_interfaces
+from smithy_core.aio.types import AsyncBytesReader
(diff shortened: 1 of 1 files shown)


python-packages/smithy-http/smithy_http/aio/crt.py
```diff
@@ -293,12 +295,42 @@ async def _marshal_request(
 
         path = self._render_path(request.destination)
         headers = crt_http.HttpHeaders(headers_list)
-        body = BytesIO(await request.consume_body_async())
+
+        body = request.body
+        if isinstance(body, bytes | bytearray):
+            # If the body is already directly in memory, wrap in in a BytesIO to hand
```
nateprewitt, line 301:
```suggestion
            # If the body is already directly in memory, wrap in a BytesIO to hand
```

python-packages/smithy-http/smithy_http/aio/crt.py
```diff
@@ -293,12 +295,42 @@ async def _marshal_request(
...
+            # If the body is async, or potentially very large, start up a task to read
+            # it into the BytesIO object that CRT needs. By using asyncio.create_task
+            # we'll start the coroutine without having to explicitly await it.
+            crt_body = BytesIO()
+            if not isinstance(body, AsyncIterable):
+                # If the body isn't already an async iterable, wrap it in one. Objects
+                # with read methods will be read in chunks so as not to exhaust memory.
+                body = AsyncBytesReader(body)
+
+            # Start the read task int the background.
```
nateprewitt, line 314:
```suggestion
            # Start the read task in the background.
```

python-packages/smithy-http/smithy_http/aio/crt.py
```diff
@@ -293,12 +295,42 @@ async def _marshal_request(
...
         )
         return crt_request
+
+    async def _consume_body_async(
+        self, source: AsyncIterable[bytes], dest: BytesIO
+    ) -> None:
+        async for chunk in source:
+            dest.write(chunk)
+        # Should we call close here? Or will that make the crt unable to read the last
+        # chunk?
```
nateprewitt, line 336: Have we tested this at all? My initial impression is we'd throw a ValueError in the CRT trying to read from a closed object. I would think this gets cleaned up for us automatically once it goes out of scope?

JordonPhillips: The problem is that it's hard to test the CRT without a live connection. If nothing else, the BytesIO will eventually get garbage collected since when this task complete there'll be no more references to it.

nateprewitt: Maybe add a mark this with a TODO so we know to come back eventually? Otherwise, that seems fine.

nateprewitt (approved):

PR #320, opened 2024-10-18
Add an asynchronous bytes provider

JordonPhillips:
This adds a class that allows bytes to be asynchronously exchanged. This primarily serves the purpose of enabling event streaming. Event streams will create a provider under the hood that will be written to. That provider will then be passed as the trasnport request body, which will read it as an async bytes iterable.

This will also fail until the protocol test suppression is merged.

### DIFF
--- a/python-packages/smithy-core/smithy_core/aio/types.py
+++ b/python-packages/smithy-core/smithy_core/aio/types.py
@@ -1,10 +1,12 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  SPDX-License-Identifier: Apache-2.0
+import asyncio
 from asyncio import iscoroutinefunction
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
 from io import BytesIO
-from typing import Self, cast
+from typing import Any, Self, cast
 
+from ..exceptions import SmithyException
 from ..interfaces import BytesReader
 from .interfaces import AsyncByteStream, StreamingBlob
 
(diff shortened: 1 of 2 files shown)


python-packages/smithy-core/tests/unit/aio/test_types.py
```diff
@@ -270,3 +276,166 @@ async def test_seekable_iter_chunks() -> None:
...
+
+async def test_provider_reads_initial_data() -> None:
+    provider = AsyncBytesProvider(intial_data=b"foo")
+    result: list[bytes] = []
+
+    # Start the read task in the background.
+    read_task = asyncio.create_task(drain_provider(provider, result))
+
+    # Wait for the buffer to drain. At that point all the data should
+    # be read, but the read dask won't actually be complete yet
```
nateprewitt, line 299:
```suggestion
    # be read, but the read task won't actually be complete yet
```

python-packages/smithy-core/tests/unit/aio/test_types.py
```diff
@@ -270,3 +276,166 @@ async def test_seekable_iter_chunks() -> None:
...
+async def test_provider_reads_written_data() -> None:
+    provider = AsyncBytesProvider()
+    result: list[bytes] = []
+
+    # Start the read task in the background.
+    read_task = asyncio.create_task(drain_provider(provider, result))
+    await provider.write(b"foo")
+
+    # Wait for the buffer to drain. At that point all the data should
+    # be read, but the read dask won't actually be complete yet
```
nateprewitt, line 323:
```suggestion
    # be read, but the read task won't actually be complete yet
```

nateprewitt (changes requested):

python-packages/smithy-core/smithy_core/aio/types.py
```diff
@@ -271,3 +273,153 @@ async def __anext__(self) -> bytes:
...
+
+class AsyncBytesProvider:
+    """A buffer that allows chunks of bytes to be exchanged asynchronously.
+
+    Bytes are written in chunks to an internal buffer, that is then drained via an async
+    iterator.
+    """
+
+    def __init__(
+        self, intial_data: bytes | None = None, max_buffered_chunks: int = 16
```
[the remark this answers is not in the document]
JordonPhillips: It's arbitrary. Ultimately there's only one coroutine reading, and N coroutines writing. If N is greater than 1, the buffer will grow. That said, the reader is going to work very quickly since it just dumps the data into a BytesIO immediately. Writers will likely be slower to push data. And I don't expect there to be that many writers to a single stream anyhow.

Depending on how real world usage ends up, we might be able to improve performance by having separate `Condition`s for readers and writers that share a lock. That may result in the side with fewer participants cycling through their queue faster. But I'm not certain about that, and it seems premature. I'd rather wait until we have actual benchmarks on a live service before adding that extra complication.

nateprewitt (approved):
:shipit: once tests pass.

PR #321, opened 2024-10-18
Skip new failing tests

JordonPhillips:
These new tests cover parts of serialization that are slated to be redone soon, so there's little value in fixing them right now.

Separate PR to break it out since so many other PRs are blocked on this


nateprewitt (commented):
~I may be missing something, but it looks like this already being done in #318 [here](https://github.com/smithy-lang/smithy-python/pull/318/files#diff-534ec6894c4aa9ca767f0d35161bb363a4d7d5f09c92e558996a9955b706218b)?~ Nevermind, mentally parsed out the second line as part of the PR agreement. Good to go 👍

nateprewitt (approved):

PR #322, opened 2024-10-22
Create high-level AWS event streams

JordonPhillips:
This updates the high-level event stream interfaces and creates AWS implementations of them.

The `EventStream` protocol was split into three protocols: `DuplexEventStream`, `InputEventStream`, and `OutputEventStream`. These three classes encompass the three different configurations that clients can expect, and each are typed with their particular use-case in mind. This lets the type declarations be more concise and accurate. Before, it could be extremely ambiguous from a typing perspective what you were getting.

The old `InputEventStream` and `OutputEventStream` classes were renamed to `AsyncEventPublisher` and `AsyncEventReceiver`, respectively. This is a more accurate description of what they do, particularly as they can be used for a service implementation as well.

In the AWS implementation, some changes needed to be made. Notably the `Event` class had to get a `decode_async` method to be able to read from an async stream. Then the calling of that method had to be pulled out of the deserializer so that both sync and async clients can use it. Test cases were updated to also test the async method.

Tests for the event stream classes will come in the form of protocol tests later on.

### DIFF
--- a/python-packages/aws-event-stream/aws_event_stream/aio/__init__.py
+++ b/python-packages/aws-event-stream/aws_event_stream/aio/__init__.py
@@ -1,0 +1,247 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+import asyncio
+from collections.abc import Callable
+from typing import Self
+
+from smithy_core.aio.interfaces import AsyncByteStream, AsyncWriter
+from smithy_core.codecs import Codec
+from smithy_core.deserializers import DeserializeableShape, ShapeDeserializer
+from smithy_core.serializers import SerializeableShape
+from smithy_event_stream.aio.interfaces import (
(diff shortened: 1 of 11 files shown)


python-packages/aws-event-stream/aws_event_stream/aio/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+            serfer implementation.
+        """
+        self.input_stream = _AWSEventPublisher(
+            payload_codec=payload_codec,
+            async_writer=async_writer,
+            signer=signer,
+            is_client_mode=is_client_mode,
+        )
+
+        self._deserialize = deserializer
```
nateprewitt, line 66: Did we mean to call this `deserializer` in some APIs but `deserialize` in others?

nateprewitt (approved):
Looks good, approving since the test failures are related to the latest Smithy release and we had a clean run yesterday with no functional changes.

PR #323, opened 2024-10-22
Generate RestJson event stream implementation

JordonPhillips:
This updates generic event stream generation with recently introduced changes and also introduces the concrete implementation for RestJson.

Testing for all of this will be done via protocol tests, and in the early days manual testing.

Since a lot of this is effectively throwaway code, I was more liberal with type ignoring and using Any types than I otherwise would be. The request pipeline is going to be moving to pure python soon^tm, and the typing issues will be resolved at that time.

This depends on #322


nateprewitt (dismissed):

nateprewitt (commented):
Looks good, approving since the test failures are related to the latest Smithy release and we had a clean run yesterday with no functional changes.

nateprewitt (approved):

PR #327, opened 2024-10-24
Lock Smithy to a static version

JordonPhillips:
This updates our Smithy versioning to a single static version that is not synced via gradle properties. The version currently locked in is 1.52.0, which is not the latest, but 1.52.1 has a bug that prevents us from generating protocol tests.

Instead of relying on the version range to keep us up to date, we will now be relying on dependabot. It can also track all the different dependency file locations, which removes the need for the syncing via gradle properties.

### DIFF
--- a/codegen/gradle.properties
+++ b/codegen/gradle.properties
@@ -1,2 +1,0 @@
-smithyVersion=[1.48.0,2.0)
-smithyGradleVersion=1.0.0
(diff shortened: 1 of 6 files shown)


nateprewitt (approved):
~~Is this something dependabot can manage for us going forward or are we going to need to come back to update all these places?~~ I can read :)

PR #330, opened 2024-10-26
Add check for RequiredTrait

### LINKED ISSUE #329: Optional Field Incorrectly Generated for Required Parameter in Python Codegen
Steps to reproduce:

```
cd smithy-python/codegen
./gradlew smithy-python-codegen-test:build
```

This produces the file: `smithy-python/codegen/smithy-python-codegen-test/build/smithyprojections/smithy-python-codegen-test/source/python-client-codegen/weather/models.py`.

Looking at the `GetCityInput` structure you will see showing that city_id is not required because it allows None.

```
class GetCityInput:

    city_id: str | None = None
```

The smithy model for GetCityInput shows:
```
@http(method: "GET", uri: "/cities/{cityId}")
operation GetCity {
    input := {
        // "cityId" provides the identifier for the resource and
        // has to be marked as required.
        @required
        @httpLabel
        cityId: CityId
    }
```

The determination to see if a field is required is done in `StructureGenerator.java`.

```
        var required = new ArrayList<MemberShape>();
        var optional = new ArrayList<MemberShape>();
        var index = NullableIndex.of(context.model());
        for (MemberShape member: shape.members()) {
            if (index.isMemberNullable(member) || member.hasTrait(DefaultTrait.class)) {
                optional.add(member);
            } else {
                required.add(member);
            }
        }
```

isMemberNullable is returning True for the cityId, and therefore is marked as optional.

I'll PR adding an additional check to see if the `RequiredTrait.class` is set on the member.

When that is in place the structure then becomes:

```
class GetCityInput:

    city_id: str
```

Or maybe this is by design? https://github.com/smithy-lang/smithy-python/issues/182

rangerthegood:
*Issue #, if available:*
Resolves: #329

*Description of changes:*
Add an additional check for the trait RequiredTrait when making the determination if a field should be optional or not.

### DIFF
--- a/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
+++ b/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
@@ -40,6 +40,7 @@ import software.amazon.smithy.model.traits.DocumentationTrait;
 import software.amazon.smithy.model.traits.ErrorTrait;
 import software.amazon.smithy.model.traits.InputTrait;
 import software.amazon.smithy.model.traits.OutputTrait;
+import software.amazon.smithy.model.traits.RequiredTrait;
 import software.amazon.smithy.model.traits.SensitiveTrait;
 import software.amazon.smithy.python.codegen.generators.MemberDeserializerGenerator;
 import software.amazon.smithy.python.codegen.generators.MemberSerializerGenerator;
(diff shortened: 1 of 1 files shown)


JordonPhillips:
The [`NullableIndex`](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java) already checks this. But there's also a number of special cases where other traits take priority, which is why that knowledge index exists in the first place.

rangerthegood:
The code `index.isMemberNullable(member)` uses the CheckMode.CLIENT by default. I was thinking of using the generated models in a server. If I change the code to `index.isMemberNullable(member, CheckMode.SERVER)` this would also fix the issue I'm facing in #329.

PR #338, opened 2024-11-25
Flush AsyncBytesProvider on close by default

JordonPhillips:
This updates the default behavior of close on AsyncBytesProvider to flush already-written data. This should be the less surprising outcome, particularly when the class is used as a context manager.

### DIFF
--- a/python-packages/smithy-core/tests/unit/aio/test_types.py
+++ b/python-packages/smithy-core/tests/unit/aio/test_types.py
@@ -342,9 +342,9 @@ async def test_close_stops_writes() -> None:
         await provider.write(b"foo")
 
 
-async def test_close_deletes_buffered_data() -> None:
+async def test_close_without_flush_deletes_buffered_data() -> None:
     provider = AsyncBytesProvider(b"foo")
-    await provider.close()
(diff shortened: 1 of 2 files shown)


python-packages/smithy-core/tests/unit/aio/test_types.py
```diff
@@ -439,3 +439,29 @@ async def test_close_with_flush() -> None:
...
+        # it waits for the queue to drain.
+        result: list[bytes] = []
+        drain_task = asyncio.create_task(drain_provider(provider, result))
+
+    # The queue should have been read by this point.
+    assert result == [b"foo"]
+
+    # The draining task should be able to complete without errors. When next it
+    # tries to get a chunk, the provider's iterator will exit.
+    await drain_task
```
nateprewitt, line 467: Is there anything we can assert about the final state here or is this only to validate nothing is raised?

JordonPhillips: This is only to validate that it closes successfully and nothing is raised. The function doesn't return anything directly, so there's nothing else to check.

nateprewitt (approved):

PR #357, opened 2024-12-19
Buffer async inputs to the crt

JordonPhillips:
The "clever" workaround to the crt taking a sync stream doesn't work. This effectively reverts the crt binding to the previous state of buffering everything immediately. This unfortunately means that the input stream must be closed before anything can be read. This is not a tenable state for the future.

The ideal resolution is for the crt to support reading from an async stream. We might have to contribute that support. There may be other options, like using futures... or sleep... which might be necessary. But please no.

### DIFF
--- a/python-packages/smithy-http/smithy_http/aio/crt.py
+++ b/python-packages/smithy-http/smithy_http/aio/crt.py
@@ -25,7 +25,7 @@ except ImportError:
     HAS_CRT = False  # type: ignore
 
 from smithy_core import interfaces as core_interfaces
-from smithy_core.aio.types import AsyncBytesReader
+from smithy_core.aio import interfaces as core_aio_interfaces
 from smithy_core.exceptions import MissingDependencyException
 
 from .. import Field, Fields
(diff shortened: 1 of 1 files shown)


python-packages/smithy-http/smithy_http/aio/crt.py
```diff
@@ -301,23 +300,14 @@ async def _marshal_request(
             # If the body is already directly in memory, wrap in a BytesIO to hand
             # off to CRT.
             crt_body = BytesIO(body)
+        elif not self._is_sync_stream(body):
+            # If the body is an async stream.... read it all into memory. This is
+            # very unfortunate, but necessary because the CRT doesn't currently
+            # have the capability to read async. We will likely have to implment
+            # this capability into it ourselves, or implment a thread-based wrapper.
+            crt_body = BytesIO(await request.consume_body_async())
```
nateprewitt, line 308: I wonder if we can do anything like `readinto` with a buffer for the CRT to avoid having to do this all at once. I'll reach out to them to see if there's anything quick we can do here.

JordonPhillips: I mean we could ostensibly pass them an `InputStream` that does its own thing, but at the end of the day it's synchronous.

PR #361, opened 2025-01-07
Rewind bodies before retrying

JordonPhillips:
Bodies weren't being rewound before a retry, which was causing some issues. Notably asyncio would hang forever awaiting the number of bytes we'd already declared in the content-length. But in general it just meant retries didn't work at all.

This updates the serializer to use a seekable stream wherever possible for non-streaming bodies. For streaming blobs, this was already being done where necessary.

### DIFF
--- a/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -334,7 +334,8 @@ final class ClientGenerator implements Runnable {
                                 except SmithyRetryException:
                                     raise context_with_response.response
                                 await sleep(retry_token.retry_delay)
-                                if (seek := getattr(context_with_transport_request.transport_request.body, "seek")) is not None:
+                                current_body =  context_with_transport_request.transport_request.body
+                                if (seek := getattr(current_body, "seek", None)) is not None:
                                     await seek(0)
                             else:
                                 # Step 8: Invoke record_success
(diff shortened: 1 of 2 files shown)


codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -334,6 +334,8 @@ async def _handle_execution(
                                 except SmithyRetryException:
                                     raise context_with_response.response
                                 await sleep(retry_token.retry_delay)
+                                if (seek := getattr(context_with_transport_request.transport_request.body, "seek")) is not None:
```
nateprewitt, line 337: We need to specify the default for the attribute not existing.
```suggestion
                                current_body =  context_with_transport_request.transport_request.body
                                if (seek := getattr(current_body, "seek", None)) is not None:
```

nateprewitt (changes requested):

codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
nateprewitt: We also probably want to implement a `seek` wrapper on `AsyncBytesReader` _or_ change our tests to use `SeekableAsyncBytesReader` too.

JordonPhillips: `SeekableAsyncBytesReader` is a wrapper already, but applying it universally won't work for non-seekable streaming blobs that are really big. We'd have to buffer it all in memory. As for tests, I'm not sure how using it there helps. It's used in two places there:

* To wrap the actual request body and immediately read it all as `bytes`.
* To wrap the actual response body and immediately read it all as `bytes`.

The actual client will still be using the seekable one. It's also worth noting that we don't do retries in the protocol tests.

nateprewitt (approved):

PR #362, opened 2025-01-10
Add serialization design doc

JordonPhillips:
This adds a design doc for schema-based serialization. In the future this will be expanded with information about deserialization, codecs, and protocols.

This is a fairly rough first draft, which I blame on having covid, but should nevertheless be understandable.

### DIFF
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -218,7 +218,7 @@ class ExampleStructure:
 ### Performing Serialization
 
 To serialize a shape, all that is needed is an instance of the shape and a
-serializer. The following shows how one might serilize a shape to JSON bytes:
+serializer. The following shows how one might serialize a shape to JSON bytes:
 
 ```python
 >>> shape = ExampleStructure(member=9)
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -123,10 +123,18 @@ class ShapeSerializer(Protocol):
         with self.begin_struct(schema=schema) as struct_serializer:
             struct.serialize_members(struct_serializer)
 
-    def begin_list(self, schema: "Schema") -> AbstractContextManager["ShapeSerializer"]:
+    def begin_list(
+        self,
+        schema: "Schema",
+        size: int,
+    ) -> AbstractContextManager["ShapeSerializer"]:
         ...
 
-    def begin_map(self, schema: "Schema") -> AbstractContextManager["MapSerializer"]:
+    def begin_map(
+        self,
+        schema: "Schema",
+        size: int,
+    ) -> AbstractContextManager["MapSerializer"]:
         ...
 
     def write_null(self, schema: "Schema") -> None:
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -19,6 +19,44 @@ based on information from a Smithy model.
   feed well-formed data from any source into a protocol and have it transform
   that data with no side-effects.
 
+## Terminology - `Protocol` vs protcol
+
+In Smithy, a "protocol" is a method of communicating with a service over a
+particular transport using a particular format. For example, the
+`aws.protocols#RestJson1` protocol is a protocol that communicates over the an
+HTTP transport that makes use of REST bindings and formats structured HTTP
+payloads in JSON.
+
+In Python, a
+[`Protocol`](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols)
+is a type that is used to define structural subtyping. For example, the
+following shows a `Protocol` and two valid implementations of it:
+
+```python
+class ExampleProtocol(Protocol):
+    def greet(self, name: str) -> str:
+        return f"Hello {name}!"
+
+class ExplicitImplementation(ExampleProtocol):
+    pass
+
+class ImplicitImplementation:
+    def greet(self, name: str) -> str:
+        return f"Good day to you {name}."
+```
+
+Since this is *structural* subtyping, it isn't required that implmentations
+actual inheret from the `Protocol` or otherwise declare that they're
+implementing it. But they *can* to make it more explicit or to inherit a default
+implementation. The `Protocol` class itself cannot be instantiated, however.
+
+This overlapping of terms clearly can cause confusion. To hopefully avoid that,
+implementations of Python's `Protocol` type will referred to using the literal
+`Protocol` or the general term "interface". (A protocol *isn't* quite the same
+thing as an interface in other programming languages, but for our purposes it's
+close enough.) Smithy protocols will be referred to simply as "protocol"s or by
+their specific protocol names (e.g. restJson1).
+
 ## Schemas
 
 The basic building block of Smithy is the "shape", a representation of data of a
@@ -108,7 +146,7 @@ then be the method by which all serialization is performed.
 Using open interfaces in this way allows for great flexibility in the generated
 Python code, which will be discussed more later.
 
-In Python, these interfaces will be represented as `Protocol`s, as shown below:
+In Python these interfaces will be represented as shown below:
 
 ```python
 @runtime_checkable
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -1,0 +1,296 @@
+# Protocol Serialization and Deserialization
+
+This document will describe how objects are serialized and deserialized
+according to some protocol, such as
+[AWS RestJson1](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html),
+based on information from a Smithy model.
+
+## Goals
+
+* Shared - Protocols should be implemented as part of a shared library. If two
+  clients using the same protocol are installed, they should use a shared
+  implementation. These implementations should be as compact as possible while
+  still being robust.
+* Hot-swappable - Implementations should be flexible enough to be swapped at
+  runtime if necessary. If a service supports more than one protocol, it should
+  be trivially easy to swap between them, even at runtime.
+* Flexible - Implementations should be useable for purposes other than as a
+  component of making a request to a web service. Customers should be able to
+  feed well-formed data from any source into a protocol and have it transform
+  that data with no side-effects.
+
+## Schemas
+
+The basic building block of Smithy is the "shape", a representation of data of a
+given type with known properties called "members", additional constraints and
+metadata called "traits", and an identifier.
+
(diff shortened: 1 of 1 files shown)


designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+
+```python
+EXAMPLE_STRUCTURE_SCHEMA = Schema.collection(
+    id=ShapeID("com.example#ExampleStructure"),
+    members={
+        "member": {
+            "target": INTEGER,
+            "index": 0,
+            "traits": [
+                Trait(id=ShapeID("smithy.api#default"), value=0),
```
mtdowling, line 87: Document values don't need any kind of wrapping type?

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+
+A `SerializeableShape` is a class that has a `serialize` method that takes a
+`ShapeSerializer` and calls the relevant methods needed to serialize it. All
+generated shapes will implement the `SerializeableShape` interface, which will
+then be the method by which all serialization is performed.
+
+Using open interfaces in this way allows for great flexibility in the generated
+Python code, which will be discussed more later.
+
+In Python, these interfaces will be represented as `Protocol`s, as shown below:
```
mtdowling, line 111: Heh, I got all tripped up because Protocol in Smithy world means something specific, but in Python it seems to be kind of interface construct.

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+@dataclass(kw_only=True)
+class ExampleStructure:
+    member: int = 0
+
+    def serialize(self, serializer: ShapeSerializer):
+        serializer.write_struct(EXAMPLE_STRUCTURE_SCHEMA, self)
+
+    def serialize_members(self, serializer: ShapeSerializer):
+        serializer.write_integer(
+            EXAMPLE_STRUCTURE_SCHEMA.members["member"], self.member
```
mtdowling, line 214: To avoid any hashmap/hashing overhead, I would either use the member index or also generate constants that refer to each member

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+        )
+```
+
+### Performing Serialization
+
+To serialize a shape, all that is needed is an instance of the shape and a
+serializer. The following shows how one might serilize a shape to JSON bytes:
```
mtdowling, line 221:
```suggestion
serializer. The following shows how one might serialize a shape to JSON bytes:
```

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+    def begin_struct(
+        self, schema: "Schema"
+    ) -> AbstractContextManager["ShapeSerializer"]:
+        ...
+
+    def write_struct(self, schema: "Schema", struct: "SerializeableStruct") -> None:
+        with self.begin_struct(schema=schema) as struct_serializer:
+            struct.serialize_members(struct_serializer)
+
+    def begin_list(self, schema: "Schema") -> AbstractContextManager["ShapeSerializer"]:
```
mtdowling, line 126: Not sure if it applies here, or if you account for it somehow elsewhere, but in smithy-java, we ensure to pass in a kind of state value to avoid needing to rely on things like state capturing lambdas (i.e., closing over the outer scope in a lambda may require an allocation, so we instead pass in a generic value to thread through the serialization process). For example:

```java
    /**
     * Begin a list and write zero or more values into it using the provided serializer.
     *
     * @param schema    List schema.
     * @param listState State to pass into the consumer.
     * @param size      Number of elements in the list, or -1 if unknown.
     * @param consumer  Received in the context of the list and writes zero or more values.
     */
    <T> void writeList(Schema schema, T listState, int size, BiConsumer<T, ShapeSerializer> consumer);
```

This would apply generally to any nested type serde.

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+
+```python
+@dataclass(kw_only=True, frozen=True)
+class Schema:
+    id: ShapeID
+    shape_type: ShapeType
+    traits: dict[ShapeID, "Trait"] = field(default_factory=dict)
+    members: dict[str, "Schema"] = field(default_factory=dict)
+    member_target: "Schema | None" = None
+    member_index: int | None = None
```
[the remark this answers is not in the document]
JordonPhillips: No, it's meta-knowledge only the code generator knows about and only the generated deserialize method uses. In most cases it will match the ordering in the members dictionary of its parent, but not always (e.g. in the case of recursive members that get inserted later). I'm thinking of getting rid of this anyway - it's a performance optimization in Java but I don't know that it makes a real difference in Python.

designs/serialization.md
JordonPhillips: `DocumentValue` is the unwrapped json-like representation

designs/serialization.md
JordonPhillips: Yeah, `Protocol`s are duck-typed interfaces essentially. You can inherit from them so they can provide default implementations, but you can't instantiate them.

designs/serialization.md
JordonPhillips: The member index can't be used as it refers to the order present in the model at generation time, which is not the same as the order in the members dict. Generating constants for each member would add massive bloat to the artifact size, which I think offsets whatever gains you might get in performance.

designs/serialization.md
JordonPhillips: This method returns a context manager, which is responsible for maintaining the state of the list. The context manager presents itself as a `ShapeSerializer` and you add elements to the list by calling the normal serializer methods. Then any necessary finalization happens when the context manager's scope ends. For example, here's how you might write a list (taken from tests):

```python
with serializer.begin_list(schema) as ls:
    for element in self.list_member:
        ls.write_string(target_schema, element)
```

Here's an implementation of the context manager for JSON lists.

```python
class JSONListSerializer(InterceptingSerializer):
    _stream: "StreamingJSONEncoder"
    _parent: JSONShapeSerializer
    _is_first_entry = True

    def __init__(
        self, stream: "StreamingJSONEncoder", parent: JSONShapeSerializer
    ) -> None:
        self._stream = stream
        self._parent = parent

    def __enter__(self) -> Self:
        self._stream.write_array_start()
        return self

    def __exit__(
        self,
        exc_type: type[BaseException] | None,
        exc_value: BaseException | None,
        traceback: TracebackType | None,
    ) -> None:
        if not exc_value:
            self._stream.write_array_end()

    def before(self, schema: "Schema") -> ShapeSerializer:
        if self._is_first_entry:
            self._is_first_entry = False
        else:
            self._stream.write_more()

        return self._parent
```

When the context manager begins, it writes `[`, then before each element aside from the first it writes `,`, and finally when the context exits without error it writes `]`.

I don't think this is the same situation as a state-capturing lambda in Java

JordonPhillips: Another element in the Java version is size hints - I'm not sure how valuable those would be. You can't pre-allocate lists in Python so I don't think it helps at all unless we have a protocol that needs to declare list size ahead of time on the wire.

designs/serialization.md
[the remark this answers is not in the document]
mtdowling: Gotcha. It's handled internally. I think it might still end up as a performance boost even in Python, assuming array indexing is faster than hashmaps. When deserializing a type, the codec does the work of identifying which member to deserialize, and then hands that to the function used to build up the type. That function needs to determine what member schema it's given to know how to set the right value on the structure. I wonder if you can use the array index somehow to better handle dispatching in a wayt that doesn't require comparing strings (which is going to be slow). If this implies double-dispatch than that has its own perf hit, but just throwing out there that we want to avoid string hashing here probably.

designs/serialization.md
mtdowling: Even with CBOR, knowing the size before serializing a list allows you to encode the list as a finite list rather than an indefinite list.

designs/serialization.md
```diff
@@ -0,0 +1,296 @@
...
+namespace com.example
+
+structure ExampleStructure {
+    member: Integer = 0
+}
+```
+
+```python
+EXAMPLE_STRUCTURE_SCHEMA = Schema.collection(
+    id=ShapeID("com.example#ExampleStructure"),
```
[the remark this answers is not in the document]
JordonPhillips: There's no cache currently, but it can be easily done. Honestly I could probably just add `@cache` to the constructor and it should work fine.

As for prelude traits, there's currently not centralized constants. There ARE centralized constants for prelude shapes (e.g. `smithy.api#String`)

designs/serialization.md
JordonPhillips: You have to add, at minimum, two extra lines of code for *every member* - one to define and and one to import it. We might end up switching to using a wildcard import though, which would eliminate the lines bit.

designs/serialization.md
jonathan343: I was also confused by this for a second when I didn't see a `Protocol` class defined in the doc. Maybe we should clarify `Protocol` comes from the `typing` library in Python and maybe link to one of the following resources:
* https://docs.python.org/3/library/typing.html#typing.Protocol
* https://peps.python.org/pep-0544/

designs/serialization.md
```diff
@@ -0,0 +1,304 @@
...
+            "index": 0,
+            "traits": [
+                Trait(id=ShapeID("smithy.api#default"), value=0),
+            ],
+        },
+    },
+)
+```
+
+## Shape Serializers and Serializeable Shapes
```
jonathan343, line 94: In botocore, we perform some validations on the input parameters before attempting to serialize the request. Can you help me understand how this translates in the smithy world? I'm assuming the schema would have all the information we need for validation. Does this happen in the `write_` methods or somewhere else? 

For example, if an integer's minimum value is `0` and `-1` is provided. Does this get validated? If so, when and where?

jonathan343 (dismissed):

designs/serialization.md
JordonPhillips: There is enough information to do this and the capability to do so. Java land has very similar patterns and is doing extensive validation for server components.

That said, client side validation is a waste of time and an active barrier to forwards compatibility. I see no reason to implement it in the client. I'm honestly surprised botocore still does it - I could have sworn we got rid of that around the time EC2 increased the size of instance IDs.

jonathan343 (approved):

PR #363, opened 2025-01-14
Add deserialization design doc

JordonPhillips:
This adds the deserialization design component to the doc introduced in #362 

### DIFF
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -351,7 +351,7 @@ methods to extract typed data from it when given a schema. For example, a
 `JSONShapeDeserializer` could be written that is constructed with JSON bytes and
 allows a caller to convert it to a shape.
 
-A `SerializeableShape` is a class that has a `deserialize` method that takes a
+A `DeserializeableShape` is a class that has a `deserialize` method that takes a
 `ShapeDeserializer` and calls the relevant methods needed to deserialize it. All
 generated shapes will implement the `DeserializeableShape` interface, which will
 then be the method by which all deserialization is performed.
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -340,3 +340,199 @@ class HTTPHeaderSerializer(SpecificShapeSerializer):
 
     [...]
 ```
+
+## Shape Deserializers and Deserializeable Shapes
+
+Deserialization will function very similarly to serialization, through the
+interaction of two interfaces: `ShapeDeserializer` and `DeserializeableShape`.
+
+A `ShapeDeserializer` is a class that is given a data source and provides
+methods to extract typed data from it when given a schema. For example, a
+`JSONShapeDeserializer` could be written that is constructed with JSON bytes and
+allows a caller to convert it to a shape.
+
+A `SerializeableShape` is a class that has a `deserialize` method that takes a
+`ShapeDeserializer` and calls the relevant methods needed to deserialize it. All
+generated shapes will implement the `DeserializeableShape` interface, which will
+then be the method by which all deserialization is performed.
+
+In Python these interfaces will be represented as shown below:
+
+```python
+@runtime_checkable
+class ShapeDeserializer(Protocol):
+
+    def read_struct(
+        self,
+        schema: "Schema",
+        state: dict[str, Any],
+        consumer: Callable[["Schema", "ShapeDeserializer", dict[str, Any]], None],
+    ) -> None:
+        ...
+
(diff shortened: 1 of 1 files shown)


jonathan343 (dismissed):

designs/serialization.md
```diff
@@ -340,3 +340,186 @@ class HTTPHeaderSerializer(SpecificShapeSerializer):
...
+which uses relatively simple integer comparision instead of the comparatively
+more expensive string comparison needed to compare based on the member name.
+Further testing needs to be done in Python to determine whether the performance
+impact justifies the extra artifact size. In Java, the compiler is also capable
+of turning an integer switch into a jump table, which CPython does not do. This
+futher reduces the potential performance benefit of the approach.
+
+It is important to note that the general approach of dealing with members
+differs from serialization. No callback functions are needed in serialization,
+but they are needed for deserialization. The reason is that deserializers must
```
mtdowling, line 496: Is there any nuance to how callables are handled in Python? e.g., are references to top-level stateless functions faster and require no allocations vs inline closures that close over external state? If so, then do the callables have enough state passed to them to make them stateless?

JordonPhillips: Callables are very generic, but how you supply them can change things in sometimes subtle ways. Anonymous functions like in the example above do require an allocation. There are alternatives though. For example, you could use a partial:


```python
@dataclass(kw_only=True)
class ExampleStructure:
    member: int = 0

    @classmethod
    def deserialize(cls, deserializer: ShapeDeserializer) -> Self:
        return cls(**cls.deserialize_kwargs(deserializer))

    @classmethod
    def deserialize_kwargs(cls, deserializer: ShapeDeserializer) -> dict[str, Any]:
        kwargs: dict[str, Any] = {}
        deserializer.read_struct(_SCHEMA_CLIENT_OPTIONAL_DEFAULTS, consumer=partial(kwargs=kwargs))
        return kwargs

    def _consumer(self, schema: Schema, de: ShapeDeserializer, kwargs: dict[str, Any]) -> None:
        match schema.expect_member_index():
            case 0:
                kwargs["member"] = de.read_integer(
                    _SCHEMA_CLIENT_OPTIONAL_DEFAULTS.members["member"]
                )
    
            case _:
                logger.debug(f"Unexpected member schema: {schema}")
```

But that still needs to allocate a `Partial` object that stores the state. It might be faster, but it's probably about the same.

The other alternative of course is to pass the `kwargs` dict through the `read_struct` instead of using a partial. I think it's probably worth doing that (iirc that's what Java is doing), but it'll be a big PR. I'd rather have this design represent the current reality and then update it after. I can put that in as a follow-up inside this document though in a "future work" section

JordonPhillips: Actually went ahead and just altered the design to pass state around. If this gets approved I'll add a ticket to implement it.

mtdowling (dismissed):

designs/serialization.md
```diff
@@ -340,3 +340,199 @@ class HTTPHeaderSerializer(SpecificShapeSerializer):
...
+
+Deserialization will function very similarly to serialization, through the
+interaction of two interfaces: `ShapeDeserializer` and `DeserializeableShape`.
+
+A `ShapeDeserializer` is a class that is given a data source and provides
+methods to extract typed data from it when given a schema. For example, a
+`JSONShapeDeserializer` could be written that is constructed with JSON bytes and
+allows a caller to convert it to a shape.
+
+A `SerializeableShape` is a class that has a `deserialize` method that takes a
```
jonathan343, line 354:
```suggestion
A `DeserializeableShape` is a class that has a `deserialize` method that takes a
```

This should be `DeserializeableShape` right?

designs/serialization.md
```diff
@@ -340,3 +340,199 @@ class HTTPHeaderSerializer(SpecificShapeSerializer):
...
+In Python these interfaces will be represented as shown below:
+
+```python
+@runtime_checkable
+class ShapeDeserializer(Protocol):
+
+    def read_struct(
+        self,
+        schema: "Schema",
+        state: dict[str, Any],
```
jonathan343, line 368: I don't understand what role `state` plays here. I don't see it referenced anywhere in the doc. Can you help me understand this better?

jonathan343: Resolving. Just read the resolved thread below with you and Michael that explains this.

jonathan343 (approved):
🚀

PR #366, opened 2025-01-17
Add design for codecs and client protocols

JordonPhillips:
This adds in the design for the higher level uses of serializers and deserializers - namely codecs and client protocols.

### DIFF
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -536,3 +536,165 @@ client call to a service. A service could, for example, model its event
 structures and include them in their client. A customer could then use the
 generated `DeserializeableShape`s to deserialize those events into Python types
 when they're received without having to do so manually.
+
+## Codecs
+
+Serializers and deserializers are never truly disconnected - where there's one,
+there's always the other. They need to be tied together in a way that makes
+sense, is portable, and which provides extra utility for common use cases.
+
+One such use case is the serialization and deserialization to and from discrete
+bytes of a common format represented by a media type such as `application/json`.
+These will be represented by the `Codec` interface:
+
+```python
+@runtime_checkable
+class Codec(Protocol):
+
+    def create_serializer(self, sink: BytesWriter) -> ShapeSerializer:
+        ...
+
+    def create_deserializer(self, source: bytes | BytesReader) -> ShapeDeserializer:
+        ...
+
+    def serialize(self, shape: SerializeableShape) -> bytes:
+        ... # A default implementation will be provided
+
+    def deserialize[S: DeserializeableShape](
+        self, source: bytes | BytesReader,
+        shape: type[S],
+    ) -> S:
+        ... # A default implementation will be provided
+```
+
+This interface provides a layer on top of serializers and deserializers that lets
+them be interacted with in a bytes-in, bytes-out way. This allows them to be used
+generically in places like HTTP message bodies. The following shows how one could
+use a JSON codec:
+
+```python
+>>> codec = JSONCodec()
+>>> deserialized = codec.deserialize(b'{"member":9}', ExampleStructure)
+>>> print(deserialized)
+ExampleStructure(member=9)
+>>> print(codec.serialize(deserialized))
+b'{"member":9}'
+```
+
+Combining them this way also allows for sharing configuration. In JSON, for
+example, there could be a configuration option to represent number types that
+can't fit in am IEEE 754 double as a string, since many JSON implementations
+(including JavaScript's) treat them as such.
+
+`Codec`s also provides opportunities for minor optimizations, such as caching
+serializers and deserializers where possible.
+
+## Client Protocols
+
+`Codec`s aren't sufficient to fully represent a protocol, however, as there is
+also a transport layer that must be created and support data binding. An HTTP
+request, for example, can have operation members bound to headers, the query
+string, the response code, etc. Such transports generally operate by interacting
+`Request` and `Response` objects rather than raw bytes, so the bytes-based
+interfaces of `Codec` aren't sufficient by themselves.
+
+```python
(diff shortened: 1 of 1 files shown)


designs/serialization.md
```diff
@@ -523,3 +523,169 @@ client call to a service. A service could, for example, model its event
...
+One such use case is the serialization and deserialization to and from discrete
+bytes of a common format represented by a media type such as `application/json`.
+These will be represented by the `Codec` interface:
+
+```python
+@runtime_checkable
+class Codec(Protocol):
+
+    @property
+    def media_type(self) -> str:
```
mtdowling, line 542: FWIW, we removed media_type from codecs in our internal design docs. It's really the protocol that dictates this. e.g., a codec might be for JSON, but the protocol actually requires an AWS JSON media type.

designs/serialization.md
```diff
@@ -523,3 +523,169 @@ client call to a service. A service could, for example, model its event
...
+        ... # A default implementation will be provided
+
+    def deserialize[S: DeserializeableShape](
+        self, source: bytes | BytesReader,
+        shape: type[S],
+    ) -> S:
+        ... # A default implementation will be provided
+```
+
+This interface provides a layer on top of serializers and deserializers that lets
```
mtdowling, line 561: This grouping also allows for adding custom configuration to the codec (useJsonNameTrait, defaultTimestampFormat, useTimestampFormatTrait, serviceNamespace, etc.) Codecs that create documents that are later used to deserialize into a typed document will need to know the namespace of the original service in cases where the document's discriminator uses a relative shape ID with no namespace.

designs/serialization.md
```diff
@@ -523,3 +523,169 @@ client call to a service. A service could, for example, model its event
...
+        return self.protocol.deserialize_response(
+            operation=EXAMPLE_OPERATION_SCHEMA,
+            error_registry=EXAMPLE_OPERATION_REGISTRY,
+            request=transport_request,
+            response=transport_response,
+            context=context,
+        )
+```
+
+As you can see, this makes the protocol and trasnport configurable at runtime.
```
mtdowling, line 684:
```suggestion
As you can see, this makes the protocol and transport configurable at runtime.
```

mtdowling (changes requested):

designs/serialization.md
JordonPhillips: Shared configuration is mentioned later

mtdowling (dismissed):

designs/serialization.md
```diff
@@ -536,3 +536,165 @@ client call to a service. A service could, for example, model its event
...
+b'{"member":9}'
+```
+
+Combining them this way also allows for sharing configuration. In JSON, for
+example, there could be a configuration option to represent number types that
+can't fit in am IEEE 754 double as a string, since many JSON implementations
+(including JavaScript's) treat them as such.
+
+`Codec`s also provides opportunities for minor optimizations, such as caching
+serializers and deserializers where possible.
```
jonathan343, line 590: I'm curious about the cases when we would/wouldn't be able to cache serializers? I see in our current implementation we create a new serializer every time we call `serialize`.

designs/serialization.md
```diff
@@ -536,3 +536,165 @@ client call to a service. A service could, for example, model its event
...
+
+`Codec`s aren't sufficient to fully represent a protocol, however, as there is
+also a transport layer that must be created and support data binding. An HTTP
+request, for example, can have operation members bound to headers, the query
+string, the response code, etc. Such transports generally operate by interacting
+`Request` and `Response` objects rather than raw bytes, so the bytes-based
+interfaces of `Codec` aren't sufficient by themselves.
+
+```python
+class ClientProtocol[Request, Response](Protocol):
```
jonathan343, line 602: It's not clear to me in this example how `ClientProtocol` interacts with the `Codec`.

designs/serialization.md
```diff
@@ -536,3 +536,165 @@ client call to a service. A service could, for example, model its event
...
+request pipeline. Deserialization takes much of the same as well as a
+`TypeRegistry` that allows it to map errors it encounters to the generated
+exception classes.
+
+In most cases these `ClientProtocol`s will be constructed with a `Codec` used to
+(de)serialize part of the request, such as the HTTP message body. Since that
+aspect is separate, it allows for flexibility through composition. Two Smithy
+protocols that support HTTP bindings but use a different body media type could
+share most of a `ClientProtocol` implementation with the `Codec` being swapped
+out to support the appropriate media type.
```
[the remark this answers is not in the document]
JordonPhillips: They certainly could be implemented that way. I'm fairly certain that we could do that as I don't *think* there's any differences in how rest-xml and rest-json handle the rest bindings.

designs/serialization.md
JordonPhillips: `create_serializer` and `create_deserializer` probably never cache because the top-level serializers and deserializers need to hold onto their data sources and that can't be shared. The `serialize` and `deserialize` probably could cache though, as they have the opportunity to reset after each call. Wouldn't be thread safe though.

designs/serialization.md
JordonPhillips: This isn't a usage example, but rather a proposed interface. A codec isn't *required*. Imagine some crazy protocol that bound everything to headers for example. Or really anything that couldn't be categorized as a media type. Since it's not required, I've not put it in the interface. I do mention down below how they generally interact though

jonathan343 (dismissed):

JordonPhillips:
Had to rebase

mtdowling (approved):

jonathan343 (approved):

PR #373, opened 2025-01-29
Enable shapes-only codegen

JordonPhillips:
This adds an alternative generation mode that only generates "data shapes". It's a widely requested feature for both this code generator (I can't tell you how many DMs about this I've received) and for other Smithy generators.

These generated types could still be used in a rich way thanks to the schemas backing them - they could still be plugged into any `ShapeSerializer` or `ShapeDeserializer`


nateprewitt (dismissed):

nateprewitt:
Changes look good, looks like we have a test failing with the changes though.

>E   ImportError: cannot import name 'ALL_QUERY_STRING_TYPES_INPUT' from 'restjson._private.schemas' (/home/runner/work/smithy-python/smithy-python/codegen/smithy-python-protocol-test/build/smithyprojections/smithy-python-protocol-test/rest-json-1/python-client-codegen/restjson/_private/schemas.py)

codegen/smithy-python-codegen-test/smithy-build.json
```diff
@@ -5,6 +5,11 @@
             "service": "example.weather#Weather",
             "module": "weather",
             "moduleVersion": "0.0.1"
+        },
+        "python-shape-codegen": {
```
haydenbaker, line 9: To be consistent with smithy-java, let's call it `python-type-codegen`?

JordonPhillips:
Note that that number of added lines is scarier than reality as I had to copy over a bunch of smithy IDL files for testing.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/PythonFormatter.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private void format(FileManifest fileManifest) {
+        try {
+            CodegenUtils.runCommand("python3 -m black -h", fileManifest.getBaseDir());
+        } catch (CodegenException e) {
+            LOGGER.warning("Unable to find the python package black. Skipping formatting.");
+            return;
+        }
+        LOGGER.info("Running code formatter on generated code");
+        CodegenUtils.runCommand("python3 -m black . --exclude \"\"", fileManifest.getBaseDir());
```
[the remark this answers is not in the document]
haydenbaker: This should still work fine, but yeah, I can't see why we wouldn't transition this over.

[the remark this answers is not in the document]
JordonPhillips: Yeah I think it would be good to move to ruff here too, but since this is just moving code I'd rather that be a separate PR. We also will need to add ruff config to the pyproject

README.md
```diff
@@ -175,6 +175,153 @@ Only for now. Once the generator has been published, the Smithy CLI will be able
...
+For example, to generate only shapes within the `com.example` namespace:
+
+```json
+{
+    "version": "1.0",
+    "sources": ["model"],
+    "maven": {
+        "dependencies": [
+            "software.amazon.smithy:smithy-model:[1.54.0,2.0)",
+            "software.amazon.smithy.python:smithy-python-codegen:0.1.0"
```
haydenbaker, line 229:
```suggestion
            "software.amazon.smithy.python.codegen.plugins:types:0.1.0"
```

README.md
```diff
@@ -175,6 +175,153 @@ Only for now. Once the generator has been published, the Smithy CLI will be able
...
+        }
+    }
+}
+```
+
+Input and output shapes (shapes with the `@input` or `@output` traits and
+operation inputs / outputs created as part of an operation definition) are not
+generated by default. To generate these shapes anyway, remove the traits with
+the
+[`excludeTraits` transform](https://smithy.io/2.0/guides/smithy-build-json.html#excludetraits):
```
haydenbaker, line 257: Hmmm.. is this something where there could be a footgun later if they are using a (hypothetical) server plugin or the client plugin? Using type codegen in conjunction with client/server codegen will be common.

README.md
haydenbaker: We also probably need to change our versions to 0.0.1

README.md
haydenbaker: I'm just calling it out, because, while it would always be possible, I'm not sure we want to point this capability out if it can lead to problems later.

haydenbaker: Also, Smithy Java generates inputs and outputs without needing the exclusion. We should be consistent.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SchemaGenerator.java
haydenbaker, line 1: If you run the generation for the weather service example, pyright throws some errors:
```
  examples/weather/build/smithy/types/python-type-codegen/weather/_private/schemas.py:569:1 - error: "STRING_LIST" is constant (because it is uppercase) and cannot be redefined (reportConstantRedefinition)
  examples/weather/build/smithy/types/python-type-codegen/weather/_private/schemas.py:1139:1 - error: "STRING_LIST" is constant (because it is uppercase) and cannot be redefined (reportConstantRedefinition)
  examples/weather/build/smithy/types/python-type-codegen/weather/_private/schemas.py:1493:1 - error: "NON_EMPTY_STRING" is constant (because it is uppercase) and cannot be redefined (reportConstantRedefinition)
  examples/weather/build/smithy/types/python-type-codegen/weather/_private/schemas.py:1507:1 - error: "NON_EMPTY_STRING_LIST" is constant (because it is uppercase) and cannot be redefined (reportConstantRedefinition)
```

This is because the same constant is being produced in the same file:
```
STRING_LIST = Schema.collection(
    id=ShapeID("aws.auth#StringList"),
    shape_type=ShapeType.LIST,
    ...
)
...
STRING_LIST = Schema.collection(
    id=ShapeID("aws.protocols#StringList"),
    shape_type=ShapeType.LIST,
    ...
)
```

Smithy Java has the same problem, but it's not checked by anything:
```
    static final Schema STRING_LIST = Schema.listBuilder(ShapeId.from("aws.auth#StringList"))
        .putMember("member", PreludeSchemas.STRING)
        .build();
    ...
    static final Schema STRING_LIST = Schema.listBuilder(ShapeId.from("smithy.test#StringList"))
        .putMember("member", PreludeSchemas.STRING)
        .build();
```

Whether or not this is *actually* problematic... if the underlying schema is always the same for a given constant, then it's not a problem.

codegen/plugins/types/src/it/java/software/amazon/smithy/python/codegen/types/PythonTypeCodegenTest.java
```diff
@@ -0,0 +1,64 @@
...
+                .unwrap();
+        PluginContext context = PluginContext.builder()
+                .fileManifest(FileManifest.create(tempDir))
+                .settings(
+                        ObjectNode.builder()
+                                .withMember("selector", """
+                                        :test([id|namespace = 'example.weather'],
+                                              [id|namespace = 'example.weather.nested'],
+                                              [id|namespace = 'example.weather.nested.more'])
+                                        """)
```
haydenbaker, line 56: realistically, users probably won't add a selector, so I'd add a test (or modify this one) without the selector. You'll get the build error from pyright throwing `reportConstantRedefinition` and `reportRedeclaration`

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SchemaGenerator.java
haydenbaker: Same root issue, the serde methods are duplicated in `models.py`, leading to `reportRedeclaration`:
```
    models.py:892:5 - error: Function declaration "_serialize_string_list" is obscured by a declaration of the same name (reportRedeclaration)
    models.py:901:5 - error: Function declaration "_deserialize_string_list" is obscured by a declaration of the same name (reportRedeclaration)
    models.py:1123:5 - error: Function declaration "_serialize_string_list" is obscured by a declaration of the same name (reportRedeclaration)
    models.py:1132:5 - error: Function declaration "_deserialize_string_list" is obscured by a declaration of the same name (reportRedeclaration)

```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/DirectedPythonClientCodegen.java
```diff
@@ -113,31 +102,17 @@ private ProtocolGenerator resolveProtocolGenerator(
 
     @Override
     public void customizeBeforeShapeGeneration(CustomizeDirective<GenerationContext, PythonSettings> directive) {
```
haydenbaker, line 104: SJ has schemas generated in the `customizeBeforeIntegrations`, but I guess there's no practical difference with type codegen? Just something to point out.

README.md
JordonPhillips: I'd rather they don't generate them tbh. I can make some improvments in this area though

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SchemaGenerator.java
JordonPhillips: Yes that's because tons of dependencies have those types. This is why the integ test only generates local shapes. There's not really a great workaround other than raising a conflict exception that's a bit more explicit.

codegen/plugins/types/src/it/java/software/amazon/smithy/python/codegen/types/PythonTypeCodegenTest.java
JordonPhillips: Eh I disagree. I think most people aren't gonna want to generate every shape in `smithy.aws` or whatever random trait package they have. The only ways around that are selectors or creating your own service shape.

README.md
JordonPhillips: I made it a simple toggle in the settings.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SchemaGenerator.java
JordonPhillips: As it turns out, we weren't running validation! I forgot that ModelTransformer doesn't validate by itself. All those string lists were private, so they shouldn't have been pulled in in the first place.

codegen/plugins/types/src/main/resources/META-INF/services/software.amazon.smithy.build.SmithyBuildPlugin
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+software.amazon.smithy.python.codegen.PythonTypeCodegenPlugin
```
haydenbaker, line 6:
```suggestion
software.amazon.smithy.python.codegen.types.PythonTypeCodegenPlugin
```

haydenbaker (dismissed):

haydenbaker (approved):

PR #378, opened 2025-02-12
Add document type design

JordonPhillips:
This adds a full design document for document types.

Note that this is a bit of a rough draft at the moment but I wanted to get it up sooner rather than later, and it nevertheless should convey all of the necessary information if not in an ideal order.

### DIFF
--- a/designs/documents.md
+++ b/designs/documents.md
@@ -1,0 +1,446 @@
+# Document Types
+
+Smithy's [document type](https://smithy.io/2.0/spec/simple-types.html#document)
+represents protocol-agnostic data of any type that can be expressed by the
+Smithy data model. In other clients, documents were represented essentially as
+JSON, but this fails to achieve a number of goals we have for the type.
+
+This specification describes how documents are defined and used in
+smithy-python.
+
+## Goals
+
+* Extend documents to fully support the Smithy data model.
+* Integrate documents with schema-based serialization and deserialization.
+* Abstract away the underlying media type of documents.
+* Allow documents to carry structured values.
+* Make documents easily convertible to/from plain Python.
+
+## Specification
+
+The `Document` type in Python is a class with typed accessors for Smithy
+data types that implements `SerializableStruct` and `DeserializableShape`.
+Below is an example of what the base `Document` type will look like at the high
+level, subsequent sections will discuss details of each component.
+
+```python
+type DocumentValue = (
+    Mapping[str, DocumentValue]
+    | Sequence[DocumentValue]
+    | str
+    | int
+    | float
+    | Decimal
+    | bool
+    | None
+    | bytes
+    | datetime.datetime
+)
+
+
+class Document:
+    _schema: Schema
+
+    def __init__(
+        self,
+        value: DocumentValue | dict[str, "Document"] | list["Document"] = None,
+        *,
+        schema: Schema = _DOCUMENT,
+    ) -> None:
+        ...
+
+    @property
+    def shape_type(self) -> ShapeType:
+        ...
+
+    @property
+    def discriminator(self) -> ShapeID:
+        ...
+
+    def is_none(self) -> bool:
+        ...
+
+    def as_blob(self) -> bytes:
+        ...
+
+    def as_boolean(self) -> bool:
+        ...
+
+    def as_string(self) -> str:
+        ...
+
+    def as_timestamp(self) -> datetime.datetime:
+        ...
+
+    def as_integer(self) -> int:
+        ...
+
+    def as_float(self) -> float:
+        ...
+
+    def as_decimal(self) -> Decimal:
+        ...
+
+    def as_list(self) -> list["Document"]:
+        ...
+
+    def as_map(self) -> dict[str, "Document"]:
+        ...
+
+    def as_value(self) -> DocumentValue:
+        ...
+
+    def as_shape[S: DeserializableShape](self, shape_class: type[S]) -> S:
+        ...
+
+    def serialize(self, serializer: ShapeSerializer) -> None:
+        serializer.write_document(self._schema, self)
+
+    def serialize_contents(self, serializer: ShapeSerializer) -> None:
+        ...
+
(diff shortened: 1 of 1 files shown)


designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+A `TypeRegistry` would then be used to get the concrete class, which is then used
+to convert the `Document` into the target class.
+
+```python
+>>> document.as_shape(registry.get(document.discriminator))
+```
+
+##### Type Registries
```
JordonPhillips, line 371: Design for these probably goes in the serialization document, but I wanted to have it here for now for the sake of making the review easier.

mtdowling (changes requested):

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+specifically will be supported and what shape types they apply to.
+
+* `__len__` - Valid for lists, maps, structures, and unions.
+* `__getitem__` - Valid for lists, maps, structures, and unions.
+* `get` - Valid for maps, structures, and unions. This is equivalent to `dict`'s
+  `get` method which allows for a default value to be specified.
+* `__setitem__` - Valid for lists, maps, and structures, and unions. Unions and
+  structures may only use keys representing members that are in their schema.
+  Unions may only replace the current value.
+* `__delitem__` - Valid for lists, maps, and structures. Notably not valid for
```
jonathan343, line 287: Can't unions have additional null members? Should that case be considered here?

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+        sub_registry: "TypeRegistry | None" = None
+    ):
+        self._types = types
+        self._sub_registry = sub_registry
+
+    def get(self, shape: ShapeID) -> DeserializeableShape:
+        ...
+```
+
+#### Botocore Interop
```
jonathan343, line 391: This seems potentially very valuable for customers familiar with boto3 and not too concerned with performance. We should definitely explore/discuss this more!

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+Currently the base `Document` implementation lazily converts between the two
+representations and caches the results. This is done for the sake of speed, but
+has an unfortunate space cost. This behavior is not part of the API, however,
+and MAY be changed without backwards compatibility implications.
+
+```python
+>>> document = Document({"foo": "bar"})
+>>> document.shape_type
+ShapeType.DOCUMENT
+>>> document.as_value()
```
SamRemis, line 228: What is the distinction here between a "value" and a python dictionary?

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Structures and Unions
+
+Smithy structures and unions are generated into their own individual classes
+in Python, and so it isn't possible to have a unique accessor for each of them.
+Instead, `Document`s have a generic `as_shape` method that supports converting
+the `Document` to any `DeserialzeableShape`.
+
+Similarly, a `from_shape` method allows for converting any `SerializeableShape`
+into a document. When a `Document` is constructed this way, it will retain the
```
SamRemis, line 242: What is the expected behavior for when the input value does not fit into the specified shape?  I'm assuming that will raise an error based on the contents here, but I'd like to see that behavior recorded in this doc.

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+
+Smithy's [document type](https://smithy.io/2.0/spec/simple-types.html#document)
+represents protocol-agnostic data of any type that can be expressed by the
+Smithy data model. In other clients this has beeen represented essentially as
+JSON, but this fails to achieve a number of goals we have for the type.
+
+This specification will describe how documents will be defined and used in
+smithy-python.
+
+## Goals
```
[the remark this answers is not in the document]
JordonPhillips: "structured" fits better I think since they've already inherently got a type

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+    "foo": "spam",
+    "bar": "eggs"
+}
+```
+
+A `TypeRegistry` would then be used to get the concrete class, which is then used
+to convert the `Document` into the target class.
+
+```python
+>>> document.as_shape(registry.get(document.discriminator))
```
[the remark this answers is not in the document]
JordonPhillips: > it would also be nice to abstract this in the type registry itself as a function

Sure

> which would also allow you to completely customize how documents are handled (e.g., if you need to unwrap some kind of envelope)

Would that not just be handled by the protocol document class? You already need them to extract the discriminator in the first place.

designs/documents.md
JordonPhillips: They may show up that way on the wire, but that's an underlying protocol behavior that should not be exposed.

designs/documents.md
JordonPhillips: I don't even think the performance hit is that significant. I would guess that the large majority of customers would not be bothered.

designs/documents.md
JordonPhillips: It's not always a dictionary. Like JSON, there is nothing requiring a document to be rooted with a mapping or list. `True` is a valid `DocumentValue` but not a valid `dict` for example.

designs/documents.md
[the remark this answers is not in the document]
mtdowling: A document / protocol wouldn't be able to handle any kind of custom enveloping, whereas the registry could.

designs/documents.md
```diff
@@ -0,0 +1,424 @@
...
+    def discriminator(self) -> ShapeID:
+        ...
+
+    def is_none(self) -> bool:
+        ...
+
+    def as_blob(self) -> bytes:
+        ...
+
+    def as_boolean(self) -> bool:
```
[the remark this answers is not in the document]
JordonPhillips: They return the expected type. If the underlying data isn't compatible, an error is thrown. This can include things like converting a base64 string into bytes for a JSON document when `as_bytes` is called.

jonathan343 (approved):

mtdowling (approved):

PR #381, opened 2025-02-20
Re-organize and tidy up codegen packages

haydenbaker:
*Description of changes:*
- Changes the codegen package structure, adds buildSrc plugins to clean up gradle configs
  - New packages will be easy to add with these changes
  - Removed aws package since it was effectively empty, can be re-added later
  - codegen-test package moved into the integration tests of core
- Adds spotless, formatted codegen
- Annotated all classes not labeled unstable as internal
- Adds instructions for setting up virtualenv with pants to ease development
- Updates Makefile to correctly display help info
```
Usage:
  make <target>

Targets:
  help                       Show this help.
  ...
  check-py                   Runs checkers for the python packages.
  ...
  clean                      Clean up generated code and artifacts.

```


codegen/smithy-aws-python-codegen/src/main/java/software/amazon/smithy/aws/python/codegen/AwsAuthIntegration.java
```diff
@@ -1,16 +0,0 @@
-/*
```
JordonPhillips, line 1: I get that these do nothing right now, but they were there so that the Python team didn't have to set up all this, so I'd appreciate you keeping them around in whatever package name aligns with what sj is doing

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/ImportDeclarations.java
```diff
@@ -1,34 +1,26 @@
...
+import software.amazon.smithy.python.codegen.SymbolProperties;
 import software.amazon.smithy.utils.SmithyInternalApi;
 import software.amazon.smithy.utils.StringUtils;
 
 /**
  * Internal class used for aggregating imports of a file.
  */
 @SmithyInternalApi
-final class ImportDeclarations implements ImportContainer {
+public final class ImportDeclarations implements ImportContainer {
```
JordonPhillips, line 23: Why does this need to be public?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonDelegator.java
```diff
@@ -1,42 +1,34 @@
...
 
 /**
  * Manages writers for Python files.
  */
-final class PythonDelegator extends WriterDelegator<PythonWriter> {
+@SmithyInternalApi
+public final class PythonDelegator extends WriterDelegator<PythonWriter> {
 
-    PythonDelegator(
+    public PythonDelegator(
```
JordonPhillips, line 23: Why does this need to be public?

JordonPhillips (changes requested):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/ImportDeclarations.java
haydenbaker: Used by `GenerationContext`

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonDelegator.java
haydenbaker: Used by `GenerationContext`

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/ImportDeclarations.java
haydenbaker: It doesn't, changing.

haydenbaker: Used in `ConfigGenerator` a la `writer.getImportContainer().addImport`

JordonPhillips (approved):
Needs to be rebased, but looks great! Thanks for doing this

PR #382, opened 2025-02-23
Add size hits to list and map serializers

jonathan343:
This PR add the `size: int` parameter to the `begin_list` and `begin_map` methods in classes that implement the `ShapeSerializer` protocol.

This follows the interface described in the [shape serializer design](https://github.com/smithy-lang/smithy-python/blob/develop/designs/serialization.md#shape-serializers-and-serializeable-shapes).

---

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ListGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ListGenerator.java
@@ -44,7 +44,7 @@ public final class ListGenerator implements Runnable {
         writer.write("""
                 def $1L(serializer: ShapeSerializer, schema: Schema, value: $2T) -> None:
                     member_schema = schema.members["member"]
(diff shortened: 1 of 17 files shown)


jonathan343:
**Open Question**
The [shape serializer design](https://github.com/smithy-lang/smithy-python/blob/develop/designs/serialization.md#shape-serializers-and-serializeable-shapes) defines the `size` parameter as a *required* integer value which this PR implements. I'm wondering if this parameter should be considered optional since it only seems applicable to protocols like CBOR which are serialized more efficiently with definite sizes.

JordonPhillips:
> I'm wondering if this parameter should be considered optional since it only seems applicable to protocols like CBOR which are serialized more efficiently with definite sizes.

Callers to these methods generally don't know what concrete implementation they're calling out to, so they always have to set it. `len` is static time generally so it shouldn't be a performance issue

JordonPhillips (dismissed):

JordonPhillips (approved):

PR #385, opened 2025-02-24
Replace pants with uv and simplify development workflow

haydenbaker:
*Description of changes:*
- Pants is over-complicated and can be frustrating to work with
- On the other hand, uv is a breath of fresh air and makes this project a breeze to work on
- This PR replaces the usage of pants with uv, and simplifies the overall development experience (including CI) 
  - Here is a link to a CI job that ran on my fork: https://github.com/haydenbaker/smithy-python/actions/runs/13503738796/job/37728351740?pr=8

*TODOs*:
- There are few TODOs after this PR is merged
  - We can add more linting rules as we see fit, I added a list near the TODO in the root pyproject.toml file
  - Update the README with a small guide on uv
  - Update the README with a section on the makefile


.github/workflows/ci.yml
```diff
@@ -1,115 +1,57 @@
...
+
 jobs:
-  test:
+  ci:
     runs-on: ubuntu-latest
-
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: ["3.12"]
```
JordonPhillips, line 18: We need to keep the matrix. Actually we should be adding 3.13 to it, but not as part of this CR

packages/aws-event-stream/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[project]
+name = "aws-event-stream"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+authors = [
+    { name = "Hayden Baker", email = "haydebak@amazon.com" }
+]
```
JordonPhillips, line 8: Bold, but not what we want here. iirc this is optional so best to remove it

packages/aws-event-stream/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[project]
+name = "aws-event-stream"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "Implementations for AWS event stream framing."
```

packages/smithy-aws-core/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "smithy-aws-core"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "Core Smithy components for AWS services and protocols."
```

packages/smithy-core/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "smithy_core"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "Core components for implementing Smithy tooling in Python."
```

packages/smithy-event-stream/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "smithy-event-stream"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "Smithy event stream interfaces and core components."
```

packages/smithy-http/pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[project]
+name = "smithy_http"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "HTTP components for Smithy tooling."
```

packages/smithy-json/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[project]
+name = "smithy-json"
+version = "0.1.0"
+description = "Add your description here"
```
JordonPhillips, line 4:
```suggestion
description = "JSON serialization and deserialization support for Smithy tooling."
```

.github/workflows/lint.yml
```diff
@@ -1,34 +0,0 @@
-name: Lint code
```
JordonPhillips, line 1: These have been removed, but it's not clear that they've been replaced. Have they?

JordonPhillips (changes requested):

packages/aws-event-stream/pyproject.toml
haydenbaker: Do we want to add any other fields from the previous toml files? e.g. `license`, `keywords`, `classifiers`, etc?

.github/workflows/lint.yml
haydenbaker: The main ci workflow does everything now. This has the benefit of only requiring a single action, and only needing to set up the environment once. It runs fairly quickly (~3 minutes, most of which is spent initializing gradle and building java).

haydenbaker: We _could_ separate out building codegen as a separate job within in the same action, but let's see if the action run time becomes an issue first.

JordonPhillips: I'm not really concerned about runtime, i'm more concerned about separation of concerns. It's useful to see at a glance what's failing and why without having to dig into some logs. This isn't a blocker though - we can make that change later if we want

packages/aws-event-stream/pyproject.toml
JordonPhillips: Yeah let's add those, but as a separate PR so that this can get merged

packages/aws-event-stream/tests/unit/test_events.py
```diff
@@ -37,7 +37,7 @@
         b"\x00\x00\x00\x10"  # total length
         b"\x00\x00\x00\x00"  # headers length
         b"\x05\xc2\x48\xeb"  # prelude crc
-        b"\x7D\x98\xc8\xff"  # message crc
+        b"\x7d\x98\xc8\xff"  # message crc
```
JordonPhillips, line 40: Dang, ruff is really this pedantic huh

JordonPhillips (approved):

PR #386, opened 2025-02-25
Add event stream design

JordonPhillips:

### DIFF
--- a/designs/event-streams.md
+++ b/designs/event-streams.md
@@ -1,0 +1,325 @@
+# Event Streams
+
+Event streams represent a behavioral difference in Smithy operations. Most
+operations work philosophically like functions in python - you provide some
+parameters once, and get results once. Event streams, on the other hand,
+represent a continual exchange of data which may be flow in one direction
+or in both directions (a.k.a. a "bidirectional" or "duplex" stream).
+
+To facilitate these different usage scenarios, the return type event stream
+operations are altered to provide customers with persistent stream objects
+that they can write or read to.
+
+## Event Publishers
+
+An `AsyncEventPublisher` is used to send events to a service.
+
+```python
+class AsyncEventPublisher[E: SerializableShape](Protocol):
+    async def send(self, event: E) -> None:
+        ...
+
+    async def close(self) -> None:
+        pass
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc_value: Any, traceback: Any):
+        await self.close()
+```
+
+Publishers expose a `send` method that takes an event class which implements
+`SerializableShape`. It then passes that shape to an internal `ShapeSerializer`
+and sends it over the connection. (Note that these `ShapeSerializer`s and
+connection types are internal, and so are not part of the interface shown
+above.)
+
+The `ShapeSerializer`s work in exactly the same way as they do for other use
+cases. They are ultimately driven by each `SerializableShape`'s `serialize`
+method.
+
+Publishers also expose a few Python standard methods. `close` can be used to
+clean up any long-running resources, such as an HTTP connection or open file
+handle. The async context manager magic methods are also supported, and by
+default they just serve to autoatically call `close` on exit. It is important
+however that implementations of `AsyncEventPublisher` MUST NOT require
+`__aenter__` or any other method to be called prior to `send`. These publishers
+are intended to be immediately useful and so any setup SHOULD take place while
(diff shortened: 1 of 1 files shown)


designs/event-streams.md
```diff
@@ -0,0 +1,275 @@
...
+lack of an input stream means that the service has nothing to wait on from the
+client before sending responses.
+
+```python
+with await client.output_operation() as stream:
+    for event in output_stream:
+        handle_event(event)
+```
+
+## FAQ
```
alextwoods, line 268: A couple of general questions (not sure where they belong)

How are error and unknown events handled?  And how are errors in the initial response (esp for Input/Duplex stream) handled?

How/where is event signing done - is this part of event publishers?

designs/event-streams.md
```diff
@@ -0,0 +1,275 @@
...
+Event streams represent a behavioral difference in Smithy operations. Most
+operations work philosophically like functions in python - you provide some
+parameters once, and get results once. Event streams, on the other hand,
+represent a continual exchange of data which may be flow in one direction
+or in both directions (a.k.a. a "bidirectional" or "duplex" stream).
+
+To facilitate these different usage scenarios, the return type event stream
+operations are altered to provide customers with persistent stream objects
+that they can write or read to.
+
```
alextwoods, line 12: Might help to show customer usage example before diving into the components.

JordonPhillips: I personally prefer examples to be at the end, but maybe I'm misunderstanding what you're asking for.

designs/event-streams.md
```diff
@@ -0,0 +1,325 @@
...
+        await self.close()
+```
+
+Publishers expose a `send` method that takes an event class which implements
+`SerializableShape`. It then passes that shape to an internal `ShapeSerializer`
+and sends it over the connection. (Note that these `ShapeSerializer`s and
+connection types are internal, and so are not part of the interface shown
+above.)
+
+The `ShapeSerializer`s work in exactly the same way as they do for other use
```
alextwoods, line 38: Is the shape serializer responsible for both serializing the shape and encoding the event stream message? In general for event streams, I had thought about these as two separate steps / responsibilities - a shape serializer generally would serialize into an event message and then a separate encoder would handle encoding to the binary wire format (and then a separate signer signs that message and re-encodes it).

JordonPhillips: > Is the shape serializer responsible for both serializing the shape and encoding the event stream message?

It can be if that makes the most sense in a given scenario. Generally it would be better to do it that way because then you avoid having to allocate an intermediate representation. The need to sign events makes that impractical, however, because the signer needs to add headers.

Deserialization similarly would be better to stream off the wire, but it runs into another issue - async. ShapeDeserializers are synchronous, so we need to read the whole event and then pass it to the deserializer so that it can all be standard.

> I had thought about these as two separate steps / responsibilities

This is some space I think could use a touch up in the implementation. Right now it's a bit of both ways. It works, but could be more efficient.

designs/event-streams.md
```diff
@@ -0,0 +1,325 @@
...
+cases. They are ultimately driven by each `SerializableShape`'s `serialize`
+method.
+
+Publishers also expose a few Python standard methods. `close` can be used to
+clean up any long-running resources, such as an HTTP connection or open file
+handle. The async context manager magic methods are also supported, and by
+default they just serve to autoatically call `close` on exit. It is important
```
jonathan343, line 45:
```suggestion
default they just serve to automatically call `close` on exit. It is important
```

jonathan343 (dismissed):

designs/event-streams.md
```diff
@@ -0,0 +1,325 @@
...
+
+An `AsyncEventPublisher` is used to send events to a service.
+
+```python
+class AsyncEventPublisher[E: SerializableShape](Protocol):
+    async def send(self, event: E) -> None:
+        ...
+
+    async def close(self) -> None:
+        pass
```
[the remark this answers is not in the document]
JordonPhillips: Correct. IIRC one of the type checkers enforces this sort of behavior.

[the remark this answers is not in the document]
jonathan343: This caught my eye because our implementation in `packages/smithy-event-stream/src/smithy_event_stream/aio/interfaces.py` uses `...` instead of `pass`. 

We should probably keep this consistent, but this is also just a design doc so not a big deal.

jonathan343 (approved):

PR #387, opened 2025-02-25
User agent

alextwoods:
*Description of changes:*
Add UserAgent header to requests in generated clients using runtime plugins with separate generic and AWS specific logic.

This change adds a generic (non-aws) plugin that adds a base UserAgent object to the interceptor context that allows other interceptors to add data to the user agent.  This is required because the user agent string MUST follow a defined order so we cannot simply append values to the header.  The generic interceptor also adds a `modify_before_sign` interceptor which sets the user agent header from the UserAgent object on the context.

The AWS specific plugin must be generated per service client to give us access to the service library's version and the service ID.  The interceptor just adds data to the UserAgent context object.

The `UserAgent` class logic is largely borrowed from botocore's [UserAgent](https://github.com/boto/botocore/blob/develop/botocore/useragent.py).

The advantages of this approach:
* UserAgent header MUST follow a defined order.  Adding a 'user-agent' object on context allows different interceptors to easily add data to the user agent without breaking the order or having to parse/rebuild the string.  
* This provides easy implementation of business metrics down the road.
* Uses existing extension points (interceptor/plugin) to modify requests rather than adding additional logic to the core execution pipeline.

Cons/Issues:
* We need access to certain config values during the request.  The interceptor context does not give access to config.  We can access a version of config when the plugin is called, but other plugins (including operation level plugins) could modify it after.


packages/smithy-aws-core/src/smithy_aws_core/interceptors/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
```
haydenbaker, line 12: Use the short form of the license header:
```
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0
```

alextwoods: Oh nice - much shorter!  Looks like theres a lot of existing files that use the long form that I was copying from.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
...
+    "ios",
+    "watchos",
+    "tvos",
+    "other",
+)
+_USERAGENT_PLATFORM_NAME_MAPPINGS = {"darwin": "macos"}
+_USERAGENT_SDK_NAME = "aws-sdk-python"
+
+
+class UserAgentComponent(NamedTuple):
```
JordonPhillips, line 27: This seems more appropriate to use a dataclass with. botocore probably couldn't due to versioning constraints. Doing something like:

```python
@dataclass(frozen=True, slots=True)
class UserAgentComponent:
    prefix: str
    name: str
    value: str | None = None
```

Would keep the usage patterns the same and shouldn't have a notable impact on speed. The one thing you don't get is unpacking or integer indexing, but I don't see that being done here anyway.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
...
+    representation these are combined in the format ``prefix/name#value``.
+
+    This class is considered private and is subject to abrupt breaking changes.
+    """
+
+    prefix: str
+    name: str
+    value: Optional[str] = None
+
+    def to_string(self):
```
JordonPhillips, line 40: Why isn't this just `__str__`? A limitation of `NamedTuple`?

JordonPhillips (changes requested):
You should run `pyupgrade --py312-plus` on all this, it'll resolve most of the typing and syntax issues I've called out.

As far as this being in AWS, I don't agree. I think almost all of this can and should be generic. The only real exception would be the app id and execution env, but those could be appended in a separate interceptor

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+
+    def __init__(self, value: str):
+        self._value = value
+
+    def to_string(self) -> str:
+        return self._value
+
+
+_UAComponent = Union[UserAgentComponent, RawStringUserAgentComponent]
```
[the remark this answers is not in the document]
haydenbaker: I'll add the linters from my TODOs and put up a PR. Pyupgrade is included in that list (ruff has built-in support): https://docs.astral.sh/ruff/rules/#pyupgrade-up

You can also run it like: `uv run ruff check --select UP --fix`

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
...
+    def with_config(
+        self,
+        ua_suffix: str | None,
+        ua_app_id: str | None,
+        sdk_version: str | None,
+    ) -> Self:
+        self._user_agent_suffix = ua_suffix
+        self._user_agent_app_id = ua_app_id
+        self._sdk_version = sdk_version
+        return self
```
nateprewitt, line 117: I'll need to look a bit closer but are we going to have any concurrency issues with this paradigm? I'm mostly curious about interceptors that might touch this.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def _get_crt_version() -> str | None:
+    """This function is considered private and is subject to abrupt breaking changes."""
+    try:
+        import awscrt
+
+        return awscrt.__version__
+    except AttributeError:
+        return None
```
nateprewitt, line 269: I'm wondering if we fold the `HAS_CRT` check into this call so it does the None/"Unknown" work for us.

nateprewitt (changes requested):

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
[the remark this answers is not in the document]
alextwoods: Nice! Wasn't aware of these.  Running the ruff fix changes a number of other files - to keep this PR isolated, I'll just run the pyupgrade command on files in this PR.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
alextwoods: Good call, switched to __str__

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
alextwoods: The `UserAgent` object is constructed for each request when the interceptor is run, so I don't think there are concurrency issues with this specific code. 

However - you are correct that there is an issue with the way the Interceptor is constructed by the plugin with config that I called out i the PR description.  Since config isn't accessible (probably correctly?) in interceptor context, the only way for us to access those values is when the plugin is called with the config.  However, that config may be mutated by other client plugins or by operation plugins, meaning that the values we had extracted from the Config in the interceptor may differ from actual... see: [plugin impl](https://github.com/smithy-lang/smithy-python/blob/203193392b4f21891e6a6194f81c66ca7b0e0a54/packages/smithy-aws-core/src/smithy_aws_core/plugins/__init__.py#L12-L15).

I'm not quite sure how to work around this fundamentally.  The easiest solution would be to have allow access to config (or some parts of it) through the interceptor context, but we may not want to expose that as part of that interface.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
```diff
@@ -0,0 +1,269 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#  pyright: reportMissingTypeStubs=false,reportUnknownMemberType=false
+
+import os
+import platform
+from string import ascii_letters, digits
+from typing import NamedTuple, Optional, Self, Union, List
+
+from smithy_http.aio.crt import HAS_CRT
```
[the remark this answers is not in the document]
JordonPhillips: 2 seems reasonable

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -289,7 +289,19 @@ private void writeInterceptorsType(PythonWriter writer) {
...
+            configProtocolSymbol = Symbol.builder()
+                    .name("HttpConfig")
+                    .namespace("smithy_http.interfaces.config", ".")
+                    .build();
+        } else {
+            configProtocolSymbol = Symbol.builder()
+                    .name("Config")
+                    .namespace("smithy_core.interfaces.config", ".")
+                    .build();
+        }
```
JordonPhillips, line 304: I think I might have made a confusing statement before. We don't need to centrally define `Protocol`s here for the config. Rather, things that care about certain config properties being present should define `Protocol`s for what they expect. For example:

```python
class InterceptorConfig(Protocol):
    interceptors: list[Interceptor[Any, Any, Any, Any]]
    
def report_interceptors(config: InterceptorConfig) -> None:
    print(f"Found {len(config.interceptors)} interceptors")
```

This works because `Protocol`s are *not* interfaces like in Java or other programming language. They can be used for many of the same purposes, but in reality they're a formalization of duck-typing. Implementations *do not* have to implement them for type checking to work.

Explicitly implementing them *can* be helpful to ensure that you didn't get something wrong, and doing so does get you any available default implementations. But in this case I don't think it gives us anything.

JordonPhillips (changes requested):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
alextwoods: I was thinking that having the explicitly defined Config would be useful for other cases that defined generic (non service specific) plugins. 

I'm happy to drop it and use a locally defined Protocol to satisfy the typing.  I'm in the middle of a refactor to make the user agent generic and address issues in ordering of the fields.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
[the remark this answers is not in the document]
alextwoods: Edit:  This isn't related to the http client that is used, its just availability of awscrt in the environment and is part of the sdk metadata (at least if we follow the patterns used in boto3 and other sdks for this)I.  think we do still potentially want to add data about which http_client is used, but that can be added down the road.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
JordonPhillips: We could have those protocols and have them be public, but it's not necessary to actually explicitly declare them on the config object. I don't think it scales. Not unless we changed the way we generate config to be based on them, which might be a good idea (but out of scope for sure here). For example:

```python
@dataclass(kw_only=True)
class BaseConfig:
    interceptors: list[Interceptor[Any, Any, Any, Any]] = field(default_factory=list)
    retry_strategy: RetryStrategy = field(default_factory=SimpleRetryStrategy)
    
    
@dataclass(kw_only=True)
class HTTPConfig:
    endpoint_uri: str | URI | None = None
    ...
    
 @dataclass(kw_only=True)
 class Config(BaseConfig, HTTPConfig):
    # Probably need to manually call all the __post_init__
    pass
```

I kinda like that actually. It wouldn't require Java to set up. Hmm....

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
```diff
@@ -0,0 +1,104 @@
...
+                                                                        sdk_version=$3T,
+                                                                        service_id='$4L'
+                                                                    )
+                                                                )
+                                                            """,
+                                                            CodegenUtils.getConfigSymbol(c.settings()),
+                                                            userAgentInterceptor,
+                                                            versionSymbol,
+                                                            c.settings().service().getName()
+                                                    );
```
JordonPhillips, line 93: Why did you move this into Java?

JordonPhillips (changes requested):

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
alextwoods: To get the client library version and service id - since both of those are specific to the generated client.

The alternative would be to add that information to the interceptor context (or potentially to use some python metaprogramming hacks?). 

Do you have thoughts or preferences between those? (or maybe another obvious way I'm missing to get client library version and service id)

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
alextwoods: Yeah - I do like that definition outside of Java and I think its likely useful for anyone trying to author generic plugins.  Theres likely some common aws specific config variants as well.

packages/smithy-aws-core/src/smithy_aws_core/user_agent.py
nateprewitt: Yeah, after thinking about it some more, I'm not sure either. Given this is still in an experimental state, I think I'm fine deferring this for right now. We may be trying to control for a non-issue in practice.

packages/smithy-aws-core/src/smithy_aws_core/__init__.py
```diff
@@ -1,12 +1,4 @@
...
-#     http://aws.amazon.com/apache2.0/
-#
-# or in the "license" file accompanying this file. This file is
-# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
-# ANY KIND, either express or implied. See the License for the specific
-# language governing permissions and limitations under the License.
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+
+__version__: str = "0.1.0"
```
nateprewitt, line 4: Is this 0.1.0 or 0.0.1? I think we've been using the latter everywhere else.

packages/smithy-core/src/smithy_core/__init__.py
```diff
@@ -8,6 +8,8 @@
 from . import interfaces, rfc3986
 from .exceptions import SmithyException
 
+__version__: str = "0.1.0"
```
nateprewitt, line 11: Same note here, are we bumping the version? I would think we do our first minor on the initial release.

packages/smithy-http/src/smithy_http/interceptors/user_agent.py
```diff
@@ -0,0 +1,151 @@
...
+
+class UserAgentBuilder:
+    def __init__(
+        self,
+        *,
+        platform_name: str | None = None,
+        platform_version: str | None = None,
+        platform_machine: str | None = None,
+        python_version: str | None = None,
+        python_implementation: str | None = None,
```
nateprewitt, line 53: I think the initial idea of having all of these required but nullable is to force the implementer to make a decision about information gathering. If you forget something it throws an error. We can make it optional if there are usability concerns, just want to make sure we're consciously making that tradeoff.

packages/smithy-http/src/smithy_http/interceptors/user_agent.py
```diff
@@ -0,0 +1,151 @@
...
+        returned. Otherwise, one component "os/other" is returned and a second
+        with prefix "md" and the raw platform name.
+
+        String representations of example return values:
+         * ``os/macos#10.13.6``
+         * ``os/linux``
+         * ``os/other``
+         * ``os/other md/foobar#1.2.3``
+        """
+        if self._platform_name is None or self._platform_name == "":
```
nateprewitt, line 105: nit; we can consolidate this into:
```suggestion
        if self._platform_name in (None, ""):
```

nateprewitt (changes requested):
I think this is probably my last round of comments, otherwise things look good!

nateprewitt:
Also looks like the test_environment tests may not have everything properly mocked:

```
AssertionError: assert 'md/arch#arm64' in 'python/1.2.3 ua/2.1 os/linux#5.4.228-131.415.AMZN2.X86_64 md/arch#x86_64 lang/python#4.3.2 md/pyimpl#Cpython'
```

It looks like you tested arm64 locally, but GHA is running on x86_64.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
```diff
@@ -58,10 +72,17 @@ public List<RuntimeClientPlugin> getClientPlugins(GenerationContext context) {
...
             final SymbolReference versionSymbol = SymbolReference.builder()
                     .symbol(Symbol.builder()
-                        .namespace(moduleName, ".")
-                        .name("__version__")
-                        .build()
-                    ).build();
+                            .namespace(moduleName, ".")
+                            .name("__version__")
+                            .build())
+                    .build();
```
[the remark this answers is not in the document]
nateprewitt: This is from Friday, I think we already discussed this, I just missed it was already queued.

packages/smithy-http/tests/unit/interceptors/test_user_agent.py
```diff
@@ -0,0 +1,71 @@
...
+
+import smithy_core
+from smithy_http.interceptors.user_agent import UserAgentBuilder
+
+
+def test_from_environment(monkeypatch):  # type: ignore
+    monkeypatch.setattr(platform, "system", lambda: "Linux")  # type: ignore
+    monkeypatch.setattr(platform, "release", lambda: "5.4.228-131.415.AMZN2.X86_64")  # type: ignore
+    monkeypatch.setattr(platform, "python_version", lambda: "4.3.2")  # type: ignore
+    monkeypatch.setattr(platform, "python_implementation", lambda: "Cpython")  # type: ignore
```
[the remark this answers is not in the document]
alextwoods: Good call - I did steal this from the boto tests so that should potentially be updated as well: https://github.com/boto/botocore/blob/develop/tests/unit/test_useragent.py#L157

packages/smithy-aws-core/src/smithy_aws_core/__init__.py
alextwoods: I took this from the pyproject: https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-aws-core/pyproject.toml#L3

Should those match?  Is there a convention in Python to keep these two values in sync?  (In Ruby for example the version is often written to a single file that both the code and library specification read from)

packages/smithy-core/src/smithy_core/__init__.py
alextwoods: Ditto on pyproject version here: https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/pyproject.toml#L3

packages/smithy-http/tests/unit/interceptors/test_user_agent.py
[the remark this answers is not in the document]
nateprewitt: Good call, https://github.com/boto/botocore/pull/3414.

packages/smithy-aws-core/src/smithy_aws_core/__init__.py
nateprewitt: Jonthan actually just brought this up, apparently we randomly bumped the version recently. I'm looking at getting it fixed, I don't think that was intentional. I'll follow up shortly.

packages/smithy-http/src/smithy_http/interceptors/user_agent.py
```diff
@@ -42,24 +42,23 @@ def modify_before_signing(
 _USERAGENT_SDK_NAME = "python"
 
 
-class _UserAgentBuilder:
+class UserAgentBuilder:
```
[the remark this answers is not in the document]
alextwoods: Edit: I went back and forth on that a little bit - I don't think we'd actually want users to subclass or use it and it should probably remain an internal detail.  I'll add the _ back.

packages/smithy-http/src/smithy_http/interceptors/user_agent.py
alextwoods: It seems generically useful to be able to construct with limited information... however, I think we shouldn't consider this as a generic class.  I've updated it to private again and I'll remove these defaults to force initialization with all values.

packages/smithy-aws-core/src/smithy_aws_core/__init__.py
haydenbaker: Can we use the [hatch version build hook](https://hatch.pypa.io/1.13/plugins/build-hook/version/) instead of maintaining version strings in init files?

nateprewitt (dismissed):

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
JordonPhillips: Ah yeah the service id isn't available right now because we're not writing out service schemas, and thus those traits aren't present.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/UserAgentIntegration.java
JordonPhillips, line 1: The formatter has truly failed here... oh well

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -324,7 +324,7 @@ private void generateConfig(GenerationContext context, PythonWriter writer) {
         writer.addStdlibImport("dataclasses", "dataclass");
         writer.write("""
                 @dataclass(init=False)
-                class $L:
+                class $T:
```
JordonPhillips, line 327: `T` is intended for references, not declarations. Why was this change made?

JordonPhillips (dismissed):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/DirectedPythonCodegen.java
```diff
@@ -273,9 +274,39 @@ public void generateIntEnumShape(GenerateIntEnumDirective<GenerationContext, Pyt
...
+                    runtimeClientPlugin.writeAdditionalFiles(context);
+                }
+            }
+        }
+    }
+
+    /**
+     * Creates top level __init__.py file.
+     */
+    private void generateServiceModuleInit(CustomizeDirective<GenerationContext, PythonSettings> directive) {
```
JordonPhillips, line 299: This could be done directly in `generateInits`. I'm refactoring this in a separate PR though so it's fine to have it here for now.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/UserAgentIntegration.java
alextwoods: Yeah - I had a bit of a fight with it trying to make this a little less garbage.  Made it slightly better, but still don't love it.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
alextwoods: Reverted to $L for the decleration - Had done it without thinking since its using the symbol as an argument.

nateprewitt (approved):

JordonPhillips (approved):

PR #392, opened 2025-02-26
Add new ruff lint rules

haydenbaker:
*Description of changes:*
- Adds a batch of new rules - I, T, UP, C4, PIE, and RUF
- There is a commit for each rule with its changes if you want to see the specific changes a rule made
  - some changes were made by hand (i.e. they didn't have autofix) 

### DIFF
--- a/packages/aws-event-stream/src/aws_event_stream/events.py
+++ b/packages/aws-event-stream/src/aws_event_stream/events.py
@@ -15,6 +15,7 @@ from collections.abc import Callable, Iterator, Mapping
 from dataclasses import dataclass
 from io import BytesIO
 from struct import pack, unpack
+from types import MappingProxyType
 from typing import Literal, Self
 
 from smithy_core.aio.interfaces import AsyncByteStream
@@ -610,11 +611,11 @@ class _DecodeUtils:
     INT64_BYTE_FORMAT = "!q"
 
     # uint byte size to unpack format
-    UINT_BYTE_FORMAT: dict[_ArraySize, str] = {
+    UINT_BYTE_FORMAT: Mapping[_ArraySize, str] = MappingProxyType({
         1: UINT8_BYTE_FORMAT,
         2: UINT16_BYTE_FORMAT,
         4: UINT32_BYTE_FORMAT,
-    }
+    })
 
     @staticmethod
     def unpack_uint8(data: BytesLike) -> tuple[int, int]:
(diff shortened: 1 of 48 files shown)


packages/smithy-core/src/smithy_core/interceptors.py
```diff
@@ -148,7 +148,6 @@ def read_before_execution(
         the `response`. If multiple `read_before_execution` methods throw exceptions,
         the latest will be used and earlier ones will be logged and dropped.
         """
-        pass
```
nateprewitt, line 151: Should these be raising NotImplementedError rather than just passing? Having the functions be no-ops seems odd.

nateprewitt (commented):
Minor formatting comments, I think we may be missing configuration for ruff in most of the projects.

packages/smithy-core/src/smithy_core/interceptors.py
JordonPhillips: They very specifically need to be no-ops. The usage is something like

```python
for interceptor in interceptors:
    interceptor.read_before_execution(...)
```

But not every implementation will implement all of these hooks. So if we had them throwing not implemented then we'd have to catch and ignore that, adding a bunch of useless thrashing

packages/aws-event-stream/src/aws_event_stream/events.py
```diff
@@ -15,6 +15,7 @@
 from dataclasses import dataclass
 from io import BytesIO
 from struct import pack, unpack
+from types import MappingProxyType
```
JordonPhillips, line 18: TIL

packages/smithy-core/src/smithy_core/types.py
```diff
@@ -16,7 +16,7 @@
     serialize_rfc3339,
 )
 
-Document: TypeAlias = (
+type Document = (
```
JordonPhillips, line 19: I forgot this was still here - probably needs to be deleted

packages/aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -21,8 +22,6 @@
 from smithy_core.shapes import ShapeID, ShapeType
 from smithy_core.traits import Trait
 
-from aws_event_stream.events import Byte, EventMessage, Long, Short
-
```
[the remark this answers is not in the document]
haydenbaker: done

JordonPhillips (approved):

packages/smithy-core/src/smithy_core/interceptors.py
nateprewitt: Should this explicitly be `return None` then as a default if we're not expecting all of these to be implemented? Having a completely bare function is abnormal.

nateprewitt: Apparently this has already been a hotly debated topic that ruff has taken an opinionated stance on. It feels weird, but we can go with it if people feel strongly about it.

JordonPhillips: Yeah iirc there are some that prefer `pass` some that prefer `...` and so on. Functions have to have some sort of statement to be valid, but that statement can be the doc string. I really don't care so long as it doesn't impact performance.

PR #393, opened 2025-02-27
Add initial aws-sdk-signers package

nateprewitt:
## Overview
The PR adds a `aws-sdk-signers` package with basic SigV4 signing capabilities. This will need to be wired up to our clients in a follow up PR and then added to sign events for the aws-eventstream implementation.

The goal is this package can live inside the smithy-python repo long term while also being able to be distributed independently of any of the smithy-* libraries. There is still some work needed to fully decouple typing for that which will be done in a future PR.

## Description of changes:

* Added SigV4Signer to sign arbitrary requests sychronously.
* Added AsyncSigV4Signer to sign arbitrary requests asychronously.


nateprewitt:
I'll take a look at the new docformatter tomorrow, some of these suggestions are definitely not what we want. We may need to tweak the config.

packages/aws-sdk-signers/requirements.txt
JordonPhillips, line 1: Is this necessary?

packages/smithy-http/src/smithy_http/aio/aiohttp.py
```diff
@@ -80,7 +80,7 @@ async def send(
         headers_list = list(
             chain.from_iterable(
                 fld.as_tuples()
-                for fld in request.fields.get_by_type(FieldPosition.HEADER)
+                for fld in request.fields
```
JordonPhillips, line 83: Why was this change made?

packages/aws-sdk-signers/src/aws_sdk_signers/_io.py
JordonPhillips, line 1: Maybe we should extract a dependency-free `smithy-io` package since these are generally useful

JordonPhillips (dismissed):
Make sure we've got a task to resolve these shared code issues. The IO stuff is easy enough to extract. Having an HTTP package with just protocols might be a bit weird, but probably reasonable?

packages/aws-sdk-signers/src/aws_sdk_signers/_io.py
nateprewitt: Yeah, I originally started with a more abstract `smithy-types` that would be a lighter lift to rely on that `smithy-core`/`smithy-http`. That has its own problems with imports so I'm fine trying to just constrain it to io for now.

packages/smithy-http/src/smithy_http/aio/aiohttp.py
nateprewitt: We're not actually adding the FieldPosition correctly for headers so we end up with an empty list here. I have a task to go back and work out what's failing. This was a carry over from the original client POC that I still need to sort out.

packages/aws-sdk-signers/requirements.txt
nateprewitt: Nope, this was a carryover from pants that didn't get cleaned up when I moved to Hayden's new layout. Will delete.

JordonPhillips (approved):

PR #394, opened 2025-02-27
Generate Regional endpoints

alextwoods:
*Description of changes:*
Adds a hook for integrations to register an endpoints generator that is responsible for generating client configuration and resolution logic.  

Adds an Aws Standard Regional Endpoints generation path that adds region to client config and generates logic to use a simple pattern based regional endpoint resolver.

The generic hook for endpoint generation can be used in the future to implement Endpoints2.0 both generically or in Aws flavors (specifically accounting for AWS builtins / std lib functions). 


packages/smithy-aws-core/src/smithy_aws_core/endpoints/standard_regional.py
```diff
@@ -0,0 +1,61 @@
...
+
+    async def resolve_endpoint(self, params: RegionalEndpointParameters) -> Endpoint:
+        if params.sdk_endpoint is not None:
+            # If it's not a string, it's already a parsed URI so just pass it along.
+            if not isinstance(params.sdk_endpoint, str):
+                return Endpoint(uri=params.sdk_endpoint)
+
+            # Does crt have implementations of these parsing methods? Using the standard
+            # library is probably fine.
+            parsed = urlparse(params.sdk_endpoint)
```
nateprewitt, line 34: The CRT doesn't have any way to parse it in their public bindings to my knowledge. This should be fine for now.

packages/smithy-aws-core/tests/unit/endpoints/test_standard_regional.py
```diff
@@ -0,0 +1,62 @@
...
+
+    assert endpoint.uri.host == "service.us-west-2.amazonaws.com"
+
+
+async def test_resolve_endpoint_with_no_sdk_endpoint_or_region():
+    resolver = StandardRegionalEndpointsResolver(endpoint_prefix="service")
+    params = RegionalEndpointParameters(sdk_endpoint=None, region=None)
+
+    with pytest.raises(EndpointResolutionError):
+        await resolver.resolve_endpoint(params)
```
nateprewitt, line 62: There's the inverse of this where both are set that I was gonna ask about in the implementation above. Conceptually I guess this is the same as using an endpoint override for precedence but it may be surprising if the user doesn't realize they set both. A quick debug log above when both are set and having a test ensuring the endpoint is always used is probably useful for catching regressions.

nateprewitt (commented):
Two minor comments, otherwise looks good!

packages/smithy-aws-core/tests/unit/endpoints/test_standard_regional.py
alextwoods: Yeah this is a good point.  In our standard endpoint rules we don't raise or log when both the custom endpoint and region are set.  Because region is *also* used as a signing configuration, for sigv4 services, customers often need to specify both (ie, if you are setting custom endpoint you almost always will need to set a region to ensure signing works correctly as well).  

I can definintely add a debug log, but we'll loose that behavior when moving to endpoints2.0.

Either way - good call, will add a test case.

nateprewitt (dismissed):

packages/smithy-aws-core/tests/unit/endpoints/test_standard_regional.py
nateprewitt: Yeah, that's fine to skip the log. We've got the test and that's fine for now. Thanks!

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -433,20 +433,13 @@ async def _handle_attempt(
...
-                            if config.endpoint_uri is None:
-                                raise $1T(
-                                    "No endpoint_uri found on the operation config."
-                                )
+            writer.write("# Step 7f: Invoke endpoint_resolver.resolve_endpoint");
 
-                            endpoint = await config.endpoint_resolver.resolve_endpoint(
-                                StaticEndpointParams(uri=config.endpoint_uri)
-                            )
+            context.endpointsGenerator().renderEndpointParameterConstruction(context, writer);
```
JordonPhillips, line 439: This is a prime use case for a [`CodeSection`](https://github.com/smithy-lang/smithy/blob/main/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeSection.java). You would create a section [like this](https://github.com/smithy-lang/smithy-python/blob/develop/codegen/core/src/main/java/software/amazon/smithy/python/codegen/sections/ConfigSection.java) called `EndpointParamSection` or something like that with the necessary context, and then a `PythonIntegration` would return a [`CodeInterceptor`](https://github.com/smithy-lang/smithy/blob/main/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeInterceptor.java) in its [`interceptors`](https://github.com/smithy-lang/smithy/blob/main/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/SmithyIntegration.java#L176) method.

Then here you'd have something like:

```java
writer.pushState(new EndpointParamsSection(context));
// Default implementation. Interceptors can be blow it away.
writer.write("""
    endpoint_parameters = StaticEndpointParams(uri=config.endpoint_uri)
""");
writer.popState();
```

You can then get rid of the `EndpointsGenerator` entirely.

smithy-python doesn't use sections or interceptors much currently, but it should.

JordonPhillips (changes requested):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
alextwoods: For an implementation of an approach using CodeSections and CodeInterceptor see: https://github.com/smithy-lang/smithy-python/pull/407

JordonPhillips:
I prefer option 2

PR #395, opened 2025-02-28
Update traits to allow for dynamically discoverable implementations

JordonPhillips:
This updates the design for traits from being simple bags of loosely typed data. Now traits may instead have concrete implementations that are discoverable dynamically simply by implementing the relevant subclass.

### DIFF
--- a/designs/serialization.md
+++ b/designs/serialization.md
@@ -80,10 +80,17 @@ implementation and/or additional helper methods.
 class Schema:
     id: ShapeID
     shape_type: ShapeType
-    traits: dict[ShapeID, "Trait"] = field(default_factory=dict)
+    traits: dict[ShapeID, "Trait | DynamicTrait"] = field(default_factory=dict)
     members: dict[str, "Schema"] = field(default_factory=dict)
     member_target: "Schema | None" = None
     member_index: int | None = None
+
+    @overload
+    def get_trait[T: "Trait"](self, t: type[T]) -> T | None: ...
+    @overload
+    def get_trait(self, t: ShapeID) -> "Trait | DynamicTrait | None": ...
+    def get_trait(self, t: "type[Trait] | ShapeID") -> "Trait | DynamicTrait | None":\
+        return self.traits.get(t if isinstance(t, ShapeID) else t.id)
 
     @classmethod
     def collection(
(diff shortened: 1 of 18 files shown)


nateprewitt (dismissed):

packages/aws-event-stream/src/aws_event_stream/_private/deserializers.py
```diff
@@ -23,7 +23,7 @@
     INITIAL_RESPONSE_EVENT_TYPE,
     get_payload_member,
 )
-from .traits import EVENT_HEADER_TRAIT
+from smithy_core.traits import EventHeaderTrait
```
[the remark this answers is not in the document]
JordonPhillips: This is only ever used by event streams, yes. But it could be used if we had an alternative event stream format or something like that. I've kept all the traits in core because so far we're only using traits from Smithy's default set of traits. But we could conceivably split them up too.

JordonPhillips:
had to rebase

designs/serialization.md
```diff
@@ -122,13 +123,112 @@ EXAMPLE_STRUCTURE_SCHEMA = Schema.collection(
...
+represents traits that have no known associated Python class. Traits not defined
+by Smithy itself may be unknown, for example, but still need representation.
+
+The other kind of trait inherits from the `Trait` class. This represents known
+traits, such as those defined by Smithy itself or those defined externally but
+made available in Python. Since these are concrete classes, they may be more
+comfortable to use, providing better typed accessors to data or even relevant
+utility functions.
+
+Both kinds of traits implement an inherent `Protocol` - they both have the `id`
```
alextwoods, line 152: Nice!

packages/smithy-core/src/smithy_core/schemas.py
```diff
@@ -121,13 +123,70 @@ def expect_member_index(self) -> int:
...
+        return result
+
+    @overload
+    def expect_trait[T: "Trait"](self, t: type[T]) -> T: ...
+
+    @overload
+    def expect_trait(self, t: ShapeID) -> "Trait | DynamicTrait": ...
+
+    def expect_trait(self, t: "type[Trait] | ShapeID") -> "Trait | DynamicTrait":
+        """Get a trait based on it's ShapeID or class.
```
nateprewitt, line 158:
```suggestion
        """Get a trait based on its ShapeID or class.
```

nateprewitt (approved):

PR #396, opened 2025-02-28
Add basic logging

jonathan343:
### Overview
This PR adds the following:
- Basic logging for request and response information.
- Transitions to lazy string formatting to prevent eager evaluation. This approach defers the construction of log messages until they are actually needed, improving performance and consistency.
- Updates codegen logic for determining if a structure's member is sensitive. We currently check if the member itself has a sensitive trait however, the `sensitive` trait can only target _"Any shape that is not a service, operation, resource, or member."_. We instead now check the targeted shape for the `sensitive` trait.

### Why avoid f-strings in logging?
Using f-strings require Python to evaluate their expressions immediately, meaning that the log message is fully formatted regardless of whether it will be output (e.g., if the log level filters it out). Customers who don't configure logs would still experience the performance hit that comes with evaluating log statements.

---

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -327,7 +327,7 @@ final class ClientGenerator implements Runnable {
                                 except SmithyRetryException:
                                     raise context_with_response.response
                                 logger.debug(
-                                    "Retry needed. Attempting request #%s in %s seconds.",
+                                    "Retry needed. Attempting request #%s in %.4f seconds.",
                                     retry_token.retry_count + 1,
                                     retry_token.retry_delay
                                 )
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -229,7 +229,7 @@ final class ClientGenerator implements Runnable {
                     event_response_deserializer: DeserializeableShape | None = None,
                     ${/hasEventStream}
                 ) -> Output:
-                    logger.debug(f"Making request for operation {operation_name} with parameters: {input}")
+                    logger.debug('Making request for operation "%s" with parameters: %s', operation_name, input)
                     context: InterceptorContext[Input, None, None, None] = InterceptorContext(
                         request=input,
                         response=None,
@@ -326,17 +326,26 @@ final class ClientGenerator implements Runnable {
                                     )
                                 except SmithyRetryException:
                                     raise context_with_response.response
+                                logger.debug(
+                                    "Retry needed. Attempting request #%s in %s seconds.",
+                                    retry_token.retry_count + 1,
+                                    retry_token.retry_delay
+                                )
                                 await sleep(retry_token.retry_delay)
                                 current_body =  context_with_transport_request.transport_request.body
                                 if (seek := getattr(current_body, "seek", None)) is not None:
                                     await seek(0)
                             else:
                                 # Step 8: Invoke record_success
+                                logger.debug(
+                                    "Attempt %s succeeded. Not retrying request.",
+                                    retry_token.retry_count + 1
+                                )
                                 retry_strategy.record_success(token=retry_token)
                                 break
                     except Exception as e:
                         if context.response is not None:
-                            logger.exception(f"Exception occurred while handling: {context.response}")
+                            logger.exception("Exception occurred while handling: %s", context.response)
                             pass
                         context._response = e
 
@@ -443,10 +452,12 @@ final class ClientGenerator implements Runnable {
                                 raise $1T(
                                     "No endpoint_uri found on the operation config."
                                 )
-
+                            endpoint_resolver_parameters = StaticEndpointParams(uri=config.endpoint_uri)
+                            logger.debug("Calling endpoint resolver with parameters: %s", endpoint_resolver_parameters)
                             endpoint = await config.endpoint_resolver.resolve_endpoint(
-                                StaticEndpointParams(uri=config.endpoint_uri)
+                                endpoint_resolver_parameters
                             )
+                            logger.debug("Endpoint resolver result: %s", endpoint)
                             if not endpoint.uri.path:
                                 path = ""
                             elif endpoint.uri.path.endswith("/"):
@@ -484,6 +495,11 @@ final class ClientGenerator implements Runnable {
             writer.write("""
                             # Step 7i: sign the request
                             if auth_option and signer:
+                                logger.debug("HTTP request to sign: %s", context.transport_request)
+                                logger.debug(
+                                    "Signer properties: %s",
+                                    auth_option.signer_properties
+                                )
                                 context._transport_request = await signer.sign(
                                     http_request=context.transport_request,
                                     identity=identity,
@@ -518,10 +534,13 @@ final class ClientGenerator implements Runnable {
                             context_with_response = cast(
                                 InterceptorContext[Input, None, $1T, $2T], context
                             )
+                            logger.debug("HTTP request config: %s", request_config)
+                            logger.debug("Sending HTTP request: %s", context_with_response.transport_request)
                             context_with_response._transport_response = await config.http_client.send(
                                 request=context_with_response.transport_request,
                                 request_config=request_config,
                             )
+                            logger.debug("Received HTTP response: %s", context_with_response.transport_response)
 
                     """, transportRequest, transportResponse);
         }
@@ -556,7 +575,7 @@ final class ClientGenerator implements Runnable {
                             interceptor.read_after_deserialization(context_with_output)
                     except Exception as e:
                         if context.response is not None:
-                            logger.exception(f"Exception occurred while handling: {context.response}")
+                            logger.exception("Exception occurred while handling: %s", context.response)
                             pass
                         context._response = e
 
@@ -582,7 +601,7 @@ final class ClientGenerator implements Runnable {
                             )
                     except Exception as e:
                         if context.response is not None:
-                            logger.exception(f"Exception occurred while handling: {context.response}")
+                            logger.exception("Exception occurred while handling: %s", context.response)
                             pass
                         context._response = e
 
@@ -592,7 +611,7 @@ final class ClientGenerator implements Runnable {
                             interceptor.read_after_attempt(context)
                         except Exception as e:
                             if context.response is not None:
-                                logger.exception(f"Exception occurred while handling: {context.response}")
+                                logger.exception("Exception occurred while handling: %s", context.response)
                                 pass
                             context._response = e
 
@@ -613,11 +632,11 @@ final class ClientGenerator implements Runnable {
                             pass
                         except Exception as e:
                             # log and ignore exceptions
-                            logger.exception(f"Exception occurred while dispatching trace events: {e}")
+                            logger.exception("Exception occurred while dispatching trace events: %s", e)
                             pass
                     except Exception as e:
                         if context.response is not None:
-                            logger.exception(f"Exception occurred while handling: {context.response}")
+                            logger.exception("Exception occurred while handling: %s", context.response)
                             pass
                         context._response = e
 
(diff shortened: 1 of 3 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -326,17 +326,26 @@ async def _handle_execution(
                                     )
                                 except SmithyRetryException:
                                     raise context_with_response.response
+                                logger.debug(
+                                    "Retry needed. Attempting request #%s in %s seconds.",
```
nateprewitt, line 330: Does retry_count start at 0 or 1 on our first call? i.e. is the first log message request 1 or 2?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -326,17 +326,26 @@ async def _handle_execution(
...
+                                    retry_token.retry_delay
+                                )
                                 await sleep(retry_token.retry_delay)
                                 current_body =  context_with_transport_request.transport_request.body
                                 if (seek := getattr(current_body, "seek", None)) is not None:
                                     await seek(0)
                             else:
                                 # Step 8: Invoke record_success
+                                logger.debug(
+                                    "Attempt %s succeeded. Not retrying request.",
```
nateprewitt, line 341: We should probably quantify what attempt succeeded, I don't know if there's enough context here to be immediately clear.

I'm also not sure if we need to log this at all as it may produce more noise than value. Generally we want to record exceptional cases rather than noting everything that happened.
```suggestion
                                    "Request attempt %s succeeded. Not retrying request.",
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -326,17 +326,26 @@ async def _handle_execution(
...
+                                    retry_token.retry_count + 1
+                                )
                                 retry_strategy.record_success(token=retry_token)
                                 break
                     except Exception as e:
                         if context.response is not None:
-                            logger.exception(f"Exception occurred while handling: {context.response}")
+                            logger.exception("Exception occurred while handling: %s", context.response)
```
nateprewitt, line 348: What gets logged with this? Do we have an example response payload?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -443,10 +452,12 @@ async def _handle_attempt(
                                 raise $1T(
                                     "No endpoint_uri found on the operation config."
                                 )
-
+                            endpoint_resolver_parameters = StaticEndpointParams(uri=config.endpoint_uri)
+                            logger.debug("Calling endpoint resolver with parameters: %s", endpoint_resolver_parameters)
```
nateprewitt, line 456: We may need to rationalize this against Alex's other Endpoint work. I hope we don't need to add different log statements for each endpoint resolver.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -484,6 +495,11 @@ async def _handle_attempt(
             writer.write("""
                             # Step 7i: sign the request
                             if auth_option and signer:
+                                logger.debug("HTTP request to sign: %s", context.transport_request)
+                                logger.debug(
+                                    "Signer properties: %s",
+                                    auth_option.signer_properties
+                                )
```
nateprewitt, line 502: I can build this to check but if you have a quick answer, what does the output of these look like? I haven't checked our repr's for them.

nateprewitt (commented):
Looking good so far, just a few initial questions

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
alextwoods: Yeah - I'm planning on doing a small refactor of the endpoint generator so that this code wouldn't be duplicated between them. (IE - the only thing that should be different between different endpoint generators is the parameter classes and how they are constructed)

alextwoods: See refactoring: https://github.com/smithy-lang/smithy-python/pull/394/commits/2a4e6f09218b852372074369a374a758b05fbe4a

This should play nicely with the logging above.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
nateprewitt: Ok if the first log starts at 2, then this probably fine. I would say let's constrain the formatting a little though for seconds. Something like `%.3f` (or 4) should work to give us enough relevant information. I don't think we need full precision for this.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: Yup, here is an example response:
```
ERROR:bedrock_runtime.client:Exception occurred while handling: Missing Authentication Token
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -229,7 +229,7 @@ async def _handle_execution(
                     event_response_deserializer: DeserializeableShape | None = None,
                     ${/hasEventStream}
                 ) -> Output:
-                    logger.debug(f"Making request for operation {operation_name} with parameters: {input}")
+                    logger.debug('Making request for operation "%s" with parameters: %s', operation_name, input)
```
jonathan343, line 232: Example:
```
DEBUG:bedrock_runtime.client:Making request for operation "Converse" with parameters: ConverseInput(model_id='amazon.titan-text-express-v1', messages=[Message(role='user', content=[ContentBlockText(value='Create a list of 3 best songs from the 1990s')])], system=None, inference_config=None, tool_config=None, guardrail_config=None, additional_model_request_fields=None, prompt_variables=None, additional_model_response_field_paths=None, request_metadata=None, performance_config=None)
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -518,10 +534,13 @@ async def _handle_attempt(
                             context_with_response = cast(
                                 InterceptorContext[Input, None, $1T, $2T], context
                             )
+                            logger.debug("HTTP request config: %s", request_config)
+                            logger.debug("Sending HTTP request: %s", context_with_response.transport_request)
```
jonathan343, line 538: Example:
```
DEBUG:bedrock_runtime.client:HTTP request config: HTTPRequestConfiguration(read_timeout=None)
DEBUG:bedrock_runtime.client:Sending HTTP request: HTTPRequest(destination=URI(scheme='https', username=None, password=None, host='bedrock-runtime.us-east-1.amazonaws.com', port=None, path='/model/amazon.titan-text-express-v1/converse', query='', fragment=None), method='POST', fields=Fields(OrderedDict({'content-type': Field(name='Content-Type', value=['application/json'], kind=<FieldPosition.HEADER: 0>), 'content-length': Field(name='Content-Length', value=['98'], kind=<FieldPosition.HEADER: 0>)})))

```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -518,10 +534,13 @@ async def _handle_attempt(
...
                             context_with_response._transport_response = await config.http_client.send(
                                 request=context_with_response.transport_request,
                                 request_config=request_config,
                             )
+                            logger.debug("Received HTTP response: %s", context_with_response.transport_response)
```
jonathan343, line 543: Example:
```
DEBUG:bedrock_runtime.client:Received HTTP response: HTTPResponse(status=403, fields=Fields(OrderedDict({'date': Field(name='Date', value=['Fri, 28 Feb 2025 18:54:22 GMT'], kind=<FieldPosition.HEADER: 0>), 'content-type': Field(name='Content-Type', value=['application/json'], kind=<FieldPosition.HEADER: 0>), 'content-length': Field(name='Content-Length', value=['42'], kind=<FieldPosition.HEADER: 0>), 'connection': Field(name='Connection', value=['keep-alive'], kind=<FieldPosition.HEADER: 0>), 'x-amzn-requestid': Field(name='x-amzn-RequestId', value=['af2aa5a0-aeae-406b-8082-c131293508e3'], kind=<FieldPosition.HEADER: 0>), 'x-amzn-errortype': Field(name='x-amzn-ErrorType', value=['MissingAuthenticationTokenException:http://internal.amazon.com/coral/com.amazon.coral.service/'], kind=<FieldPosition.HEADER: 0>)})), reason='Forbidden')

```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
nateprewitt: I may need to tweak our reprs for fields and URIs so they're more readable. What we have now isn't great for logging.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: Removed this in 7d90ac5ff71221af5b51649ea9b9a12adb739193

We can always add it back if we want to.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: I went with `%.4f`. This was added in 36b28c71fa60e109aa90b1bf5088509a8ecd6ca6

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: Why modify the `__repr__` and not `__str__`? Based off the [docs for `__repr__`](__repr__), what we have now is preferred:

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).
> ...
> This is typically used for debugging, so it is important that the representation is information-rich and unambiguous.

Seems like [`__str__`](https://docs.python.org/3/reference/datamodel.html#object.__str__) is what we want to modify:
> Called by [str(object)](https://docs.python.org/3/library/stdtypes.html#str), the default [__format__()](https://docs.python.org/3/reference/datamodel.html#object.__format__) implementation, and the built-in function [print()](https://docs.python.org/3/library/functions.html#print), to compute the “informal” or nicely printable string representation of an object. The return value must be a [str](https://docs.python.org/3/library/stdtypes.html#textseq) object.

> This method differs from [object.__repr__()](https://docs.python.org/3/reference/datamodel.html#object.__repr__) in that there is no expectation that [__str__()](https://docs.python.org/3/reference/datamodel.html#object.__str__) return a valid Python expression: a more convenient or concise representation can be used.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: The `signer_properties` value would print like a normal python dictionary.

`auth_option` is either `HTTPAuthOption | None` and at this point we've checked `auth_option` isn't None. I've copied over the `HTTPAuthOption` class below for convenience:
```py
@dataclass(kw_only=True)
class HTTPAuthOption:
    """Auth scheme used for signing and identity resolution."""

    # The ID of the scheme to use. This string matches the one returned by
    # HttpAuthScheme.scheme_id
    scheme_id: str

    # Parameters to pass to IdentityResolver.get_identity.
    identity_properties: dict[str, Any]

    # Parameters to pass to HttpSigner.sign.
    signer_properties: dict[str, Any]
```

JordonPhillips (approved):

PR #400, opened 2025-03-04
Add HTTP Binding serializers and deserializers

JordonPhillips:
This adds ShapeSerializers and ShapeDeserializers for HTTP bindings. It's a draft that still needs tests but it should work, more or less.

### DIFF
--- a/packages/smithy-core/src/smithy_core/deserializers.py
+++ b/packages/smithy-core/src/smithy_core/deserializers.py
@@ -3,11 +3,12 @@ from collections.abc import Callable
 from decimal import Decimal
 from typing import TYPE_CHECKING, Never, Protocol, Self, runtime_checkable
 
-from .exceptions import SmithyException
+from .exceptions import SmithyException, UnsupportedStreamException
 
 if TYPE_CHECKING:
     from .documents import Document
     from .schemas import Schema
+    from .aio.interfaces import StreamingBlob as _Stream
 
 
 @runtime_checkable
@@ -171,6 +172,22 @@ class ShapeDeserializer(Protocol):
         """
         ...
 
+    def read_data_stream(self, schema: "Schema") -> "_Stream":
+        """Read a data stream from the underlying data.
+
+        The data itself MUST NOT be read by this method. The value returned is intended
+        to be read later by the consumer. In an HTTP implementation, for example, this
+        would directly return the HTTP body stream. The stream MAY be wrapped to provide
+        a more consistent interface or to avoid exposing implementation details.
+
+        Data streams are only supported at the top-level input and output for
+        operations.
+
+        :param schema: The shape's schema.
+        :returns: A data stream derived from the underlying data.
+        """
+        raise UnsupportedStreamException()
+
 
 class SpecificShapeDeserializer(ShapeDeserializer):
     """Expects to deserialize a specific kind of shape, failing if other shapes are
@@ -245,6 +262,9 @@ class SpecificShapeDeserializer(ShapeDeserializer):
         self._invalid_state(schema)
 
     def read_timestamp(self, schema: "Schema") -> datetime.datetime:
+        self._invalid_state(schema)
+
+    def read_data_stream(self, schema: "Schema") -> "_Stream":
         self._invalid_state(schema)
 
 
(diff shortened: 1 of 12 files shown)


JordonPhillips:
I hate that you can't convert to draft after the fact

nateprewitt:
You can, the UX is just a little out of the way. Under reviewers press this: 
Image of Convert to Draft button

packages/smithy-core/src/smithy_core/deserializers.py
```diff
@@ -247,6 +264,9 @@ def read_document(self, schema: "Schema") -> "Document":
     def read_timestamp(self, schema: "Schema") -> datetime.datetime:
         self._invalid_state(schema)
 
+    def read_data_stream(self, schema: "Schema") -> "_Stream":
```
nateprewitt, line 267: What's the reason we're importing `StreamingBlob` as a private variable? It seems odd that we're returning a private type from a public interface.

packages/smithy-core/src/smithy_core/interfaces/__init__.py
```diff
@@ -58,6 +59,27 @@ class BytesReader(Protocol):
...
+    )
+
+
 # A union of all acceptable streaming blob types. Deserialized payloads will
 # always return a ByteStream, or AsyncByteStream if async is enabled.
 type StreamingBlob = BytesReader | bytes | bytearray
+
+
+def is_streaming_blob(obj: Any) -> TypeGuard[StreamingBlob]:
+    """Determines wheter the given object is a StreamingBlob.
```
nateprewitt, line 81:
```suggestion
    """Determines whether the given object is a StreamingBlob.
```

packages/smithy-core/src/smithy_core/types.py
```diff
@@ -111,3 +115,41 @@ def deserialize(self, value: str | float) -> datetime:
                 return ensure_utc(parsedate_to_datetime(expect_type(str, value)))
             case TimestampFormat.DATE_TIME:
                 return ensure_utc(datetime.fromisoformat(expect_type(str, value)))
+
+
+@dataclass(init=False, frozen=True)
+class PathPattern:
+    """A formattable URI path pattern.
+
+    The pattern may contain formattlable labels, which may be normal labels or greedy
```
nateprewitt, line 124:
```suggestion
    The pattern may contain formattable labels, which may be normal labels or greedy
```

packages/smithy-http/tests/unit/conftest.py
haydenbaker, line 1: Did you need to add this for some reason or did you just add it for defining fixtures later

packages/smithy-core/src/smithy_core/deserializers.py
JordonPhillips: I really don't want these being re-exported since there's two symbols of the same name in different namespaces. It's not a major deal though. From a type checking perspective it'll expand the union instead of referring to the local symbol name anyway.

packages/smithy-http/tests/unit/conftest.py
JordonPhillips: I was trying to define all this test setup as fixtures but was unable to make it work well with typing.

packages/smithy-core/src/smithy_core/deserializers.py
JordonPhillips: Oh wait they're in TYPE_CHECKING so won't be exported in any capacity

packages/smithy-core/src/smithy_core/traits.py
```diff
@@ -193,3 +193,117 @@ def __post_init__(self):
     @property
     def value(self) -> str:
         return self.document_value  # type: ignore
+
+
+# TODO: Get all this moved over to the http package
```
haydenbaker, line 198: _Does_ it need to be? I don't think we need to put traits in their "respective" packages

haydenbaker (approved):

packages/smithy-core/src/smithy_core/traits.py
nateprewitt: I'd tend to agree with this, I think the other PR we moved everything into the centralized location. I like that unless we have a technical limitation of keeping them together. Is the concern that we may include traits that aren't needed if the smithy-http package isn't present, or are we hitting typing problems?

nateprewitt (approved):

haydenbaker:
Looks like you need to run the gradle build before committing, `ClientGenerator` needs re-formatting.

packages/smithy-core/src/smithy_core/traits.py
JordonPhillips: It's more about fighting against package size inflation, though practically for HTTP it's not much of an issue.

PR #405, opened 2025-03-05
Add Support for Python 3.13

jonathan343:
### Summary
This PR adds Python 3.13 to the list of versions our CI runs on. Additionally, it adds the `Programming Language :: Python :: 3.12` classifier to existing classifier lists. This matches our supported python version range (`requires-python = ">=3.12"`). 

> [!NOTE]  
> Note: As of the time of this PR, the `Programming Language :: Python :: 3.14` classifier is available, however, 3.14 is still in the alpha stage and not scheduled to be released until October 2025.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.12"]
+        python-version: ["3.12", "3.13"]
 
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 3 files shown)


nateprewitt:
If we're going to claim 3.13 support, we need to be running the CI on that version first.

jonathan343:
> If we're going to claim 3.13 support, we need to be running the CI on that version first.

Got it. I don't know of anything preventing us from supporting 3.13. I'll look into getting that added when I get some time.

JordonPhillips:
You should just have to add it to the version matrix in the github action yaml [here](https://github.com/smithy-lang/smithy-python/blob/develop/.github/workflows/ci.yml#L17)

nateprewitt (approved):

PR #406, opened 2025-03-05
Sigv4 AuthScheme + Static/Environment Credentials Providers

alextwoods:
*Description of changes:*
Adds the Sigv4 AuthScheme to the AWS AuthIntegration and wires up the identity provider/auth params configs.

*Generated Examples:*

config.py:
```python
@dataclass(init=False)
class Config:
    """Configuration for AmazonBedrockFrontendService."""
     # .... other properties
    aws_credentials_identity_resolver: (
        IdentityResolver[AWSCredentialIdentity, IdentityProperties] | None
    )
    region: str | None
```

auth.py:
```python
@dataclass
class HTTPAuthParams:
    operation: str
    region: str | None


class HTTPAuthSchemeResolver:
    def resolve_auth_scheme(
        self, auth_parameters: HTTPAuthParams
    ) -> list[HTTPAuthOption]:
        auth_options: list[HTTPAuthOption] = []

        if (option := _generate_sigv4_option(auth_parameters)) is not None:
            auth_options.append(option)

        return auth_options


def _generate_sigv4_option(auth_params: HTTPAuthParams) -> HTTPAuthOption | None:
    return HTTPAuthOption(
        scheme_id="aws.auth#sigv4",
        identity_properties={},
        signer_properties={"service": "bedrock", "region": auth_params.region},
    )

```

client.py:
```python
auth_parameters: HTTPAuthParams = HTTPAuthParams(
                operation=operation_name,
                region=config.region,
            )
```


packages/smithy-aws-core/src/smithy_aws_core/auth/sigv4.py
```diff
@@ -0,0 +1,54 @@
...
+        signer: HTTPSigner[AWSCredentialsIdentity, SigV4SigningProperties]
+        | None = None,
+    ) -> None:
+        """Constructor.
+
+        :param identity_resolver: The identity resolver to extract the api key identity.
+        :param signer: The signer used to sign the request.
+        """
+        self.scheme_id = "aws.auth#sigv4"
+        # TODO: There are type mismatches in the signature of the "sign" method.
```
nateprewitt, line 43: Do you know what the mismatches were? I can take a look at those?

alextwoods: I think its primarily a naming, http_request vs request, see: https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-http/src/smithy_http/aio/interfaces/auth.py#L39

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/static_credentials_resolver.py
jonathan343, line 1: nit - Do we need to repeat `credentials_resolver` in the file name if it's already under the `credentials_resolvers` directory?

jonathan343: I'm less worried about this after seeing you added the following in `__init__.py`:

```py
__all__ = ("EnvironmentCredentialsResolver", "StaticCredentialsResolver")
```

alextwoods: We probably don't need to repeat it either way - I'm not sure what python conventions are on this.  Is it convention for the file name to match the class name?

jonathan343: >  Is it convention for the file name to match the class name?

No, the only formal recommendation I know of is in [PEP 8](https://peps.python.org/pep-0008/#package-and-module-names):
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

I think something like the structure below makes sense:
```
└── smithy_aws_core
    ├── __init__.py
    ├── credentials_resolvers
    │   ├── __init__.py
    │   ├── environment.py
    │   └── static.py

```

nateprewitt: modules in python (the .py files) are not inherently tied to a class implementation like Java or other languages. They're just logical groupings of things without a lot of hard rules. Generally, we've tended to treat them as hierarchies.

If implementation are smaller, we may group them all into a single file like `signers.py`, especially if they have shared components. If there's a lot of complexity, you may consider the directory strategy where you have a file per implementation to avoid leaking things between them/for general readability.

packages/smithy-aws-core/src/smithy_aws_core/auth/sigv4.py
```diff
@@ -0,0 +1,54 @@
...
+        self.signer = signer or SigV4Signer()  # type: ignore
```
jonathan343, line 44: Why is `SigV4Signer` being used here and not `AsyncSigV4Signer`?

jonathan343: Maybe a better question is why the generated clients are currently producing:
```py
self.http_auth_schemes = http_auth_schemes or {
    "aws.auth#sigv4": SigV4AuthScheme(),
}
```
as opposed to:
```py
self.http_auth_schemes = http_auth_schemes or {
    "aws.auth#sigv4": SigV4AuthScheme(singer=AsyncSigV4Signer()),
}
```

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/static_credentials_resolver.py
alextwoods: For now I renamed to shorter names - environment.py and static.py under `credentials_resolvers` (though the __init__.py still makes them available through the module, and I didn't rename the classes themselves).  I think generically for credentials resolvers they can get fairly complex, so keeping them in their own files is probably better for long term maintainability.

packages/smithy-aws-core/src/smithy_aws_core/auth/sigv4.py
alextwoods: Yeah - it should be the AsyncSigV4Signer for sure.  I thought I had changed that earlier.  Anyway, I pushed a fix for that.

As to why the generated config has just `SigV4AuthScheme()` - I think in general auth schemes should be able to be constructed without arguments (ie, in this case we know a good default signer to use), and should only need to be provided with arguments when you're overriding behavior - and the codegen is setup for this, [AuthScheme](https://github.com/smithy-lang/smithy-python/blob/develop/codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/AuthScheme.java#L61)  only allows specifying the symbol for the auth scheme (and doesn't provide a mechanism to provide default arguments to it).  

But - we could change that interface to allow specifying default parameters... are there any cases or behavior that helps us cover?

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAuthIntegration.java
```diff
@@ -4,11 +4,159 @@
...
+        if (!hasSigV4Auth(context)) {
+            return;
+        }
+        var trait = context.settings().service(context.model()).expectTrait(SigV4Trait.class);
+        var params = CodegenUtils.getHttpAuthParamsSymbol(context.settings());
+        var resolver = CodegenUtils.getHttpAuthSchemeResolverSymbol(context.settings());
+
+        // Add a function that generates the http auth option for api key auth.
+        // This needs to be generated because there's modeled parameters that
+        // must be accounted for.
```
JordonPhillips, line 85: Tragic

packages/smithy-aws-core/src/smithy_aws_core/auth/sigv4.py
jonathan343: Cool, thanks.

And sorry, my wording was confusing. My intention was to suggest using `AsyncSigV4Signer()` by default and suggest the override if there was a reason we didn't want to change it default. Looks like this was just a miss though and there isn't a reason to not change the default.

JordonPhillips (changes requested):

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAuthIntegration.java
alextwoods: I think the alternative would be to add signingService as an auth param.  We would then need a way to wire that up during the request.  It could be an undocumented config property or have it somewhere else.

packages/smithy-aws-core/src/smithy_aws_core/auth/sigv4.py
```diff
@@ -0,0 +1,54 @@
...
+
+@dataclass(init=False)
+class SigV4AuthScheme(
+    HTTPAuthScheme[
+        AWSCredentialsIdentity, SigV4Config, IdentityProperties, SigV4SigningProperties
+    ]
+):
+    """SigV4 AuthScheme."""
+
+    scheme_id: str
```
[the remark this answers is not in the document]
alextwoods: Yeah, I think we'd need to update the HttpAuthScheme protocol as well for that.  I've just updated to Final for now.

packages/smithy-aws-core/tests/unit/credentials_resolvers/test_environment_credentials_resolver.py
```diff
@@ -0,0 +1,68 @@
...
+
+async def test_no_values_set():
+    with pytest.raises(SmithyIdentityException):
+        await EnvironmentCredentialsResolver().get_identity(
+            identity_properties=IdentityProperties()
+        )
+
+
+async def test_required_values_missing(monkeypatch):  # type: ignore
+    monkeypatch.setenv("AWS_ACCOUNT_ID", "123456789012")  # type: ignore
```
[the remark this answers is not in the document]
alextwoods: Doh - well that was obvious.  I spent too long trying to get the typing to work and somehow missed this, haha.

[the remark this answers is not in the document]
JordonPhillips: Their docs don't include typing and all their classes are indirect so you kinda have to go code diving to get the type hints, which is unfortunate

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/static.py
```diff
@@ -0,0 +1,19 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from smithy_aws_core.identity import AWSCredentialsIdentity
+from smithy_core.aio.interfaces.identity import IdentityResolver
+from smithy_core.interfaces.identity import IdentityProperties
+
+
+class StaticCredentialsResolver(
+    IdentityResolver[AWSCredentialsIdentity, IdentityProperties]
```
JordonPhillips, line 9: Did changing `AWSCredentialsResolver` to a `type` make it unusable here? Wondering at the change.

JordonPhillips (approved):

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/static.py
alextwoods: Yeah - type aliases can't be used as base classes, which makes sense.  I think the type alias is still useful though.

jonathan343 (approved):

PR #407, opened 2025-03-06
Generate Regional Endpoints (OPTION 2: CodeSection/Interceptor)

alextwoods:
*Description of changes:*
Based on [suggestion](https://github.com/smithy-lang/smithy-python/pull/394#discussion_r1981277883) in #394, this PR implements StandardRegionalEndpoints using CodeSections + interceptors.

Similar to the `HttpAuthGenerator`, we add a standard `EndpointGenerator` which always renders `endpoints.py` which must define the service's `EndpointParameters` and `EndpointResolver[EndpointParameters]` and which adds code sections for both.

The AWSStandardRegionalIntegation adds the required config properties through a runtime plugin and adds interceptors that completely rewrite the endpoint param and resolver sections.  For both the base and standard regional cases, we simply import generic (non service-specific) parameters and resolvers and export them.  For endpoints2.0 we would generate both in place.


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
JordonPhillips, line 1: I hate that spotless reformatted all of this. It was fairly deliberately formatted to keep everything at a level so it could be read with minimal head scratching. But we're getting rid of these soon anyway so oh well.

packages/smithy-http/src/smithy_http/endpoints.py
```diff
@@ -13,11 +16,23 @@ class Endpoint(interfaces.Endpoint):
     headers: interfaces.Fields = field(default_factory=Fields)
 
 
+class EndpointResolutionError(SmithyException):
+    """Exception type for all exceptions raised by endpoint resolution."""
```
JordonPhillips, line 20: nit: I'd prefer that all exceptions go in a central `exceptions.py`

packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
```diff
@@ -12,14 +12,23 @@
     HTTPRequestConfiguration,
 )
 
+Config = TypeVar("Config", contravariant=True)
+
+
+class EndpointParametersProtocol(Protocol[Config]):
+    @classmethod
+    def build(cls, config: Config) -> Self:
+        raise NotImplementedError()
```
JordonPhillips, line 21: As of python 3.12 we have proper generic declarations and inferred variance, so `TypeVar` is not needed.

```suggestion
class EndpointParameters[C](Protocol):
    @classmethod
    def build(cls, config: C) -> Self:
        raise NotImplementedError()
```

packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
```diff
@@ -12,14 +12,23 @@
...
+
 # EndpointParams are defined in the generated client, so we use a TypeVar here.
 # More specific EndpointParams implementations are subtypes of less specific ones. But
 # consumers of less specific EndpointParams implementations are subtypes of consumers
 # of more specific ones.
 EndpointParams = TypeVar("EndpointParams", contravariant=True)
 
 
-class EndpointResolver(Protocol[EndpointParams]):
+class EndpointResolverProtocol(Protocol[EndpointParams]):
```
JordonPhillips, line 31: Adding type information in the name of the type isn't a thing we should be doing. Especially in Python, where you can just alias the name on import to avoid conflicts.

JordonPhillips (changes requested):

packages/smithy-http/src/smithy_http/endpoints.py
alextwoods: Makes sense.  Moved.

JordonPhillips (approved):

PR #409, opened 2025-03-06
Default to CRT HTTP Client for all EventStreams

alextwoods:
*Description of changes:*
Based on internal discussion, use the CRT HTTP Client as the default for all event streams - single and bi-directional.  This provides better forwards compatibility when a service moves from single directional event streams to bi-directional, but is not potentially the right long term solution.

This PR also adds the default HTTP client as an explicit dependency to improve the out of the box experience.

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
@@ -138,7 +138,8 @@ public final class ConfigGenerator implements Runnable {
         if (usesHttp2(context)) {
             clientBuilder
                     .initialize(writer -> {
-                        writer.addDependency(SmithyPythonDependency.SMITHY_HTTP.withOptionalDependencies("awscrt"));
+                        writer.addDependency(SmithyPythonDependency.SMITHY_HTTP);
+                        writer.addDependency(SmithyPythonDependency.AWS_CRT);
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
:shipit:

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -138,15 +138,17 @@ private static List<ConfigProperty> getHttpProperties(GenerationContext context)
...
                         writer.addImport("smithy_http.aio.crt", "AWSCRTHTTPClient");
                         writer.write("self.http_client = http_client or AWSCRTHTTPClient()");
                     });
 
         } else {
             clientBuilder
                     .initialize(writer -> {
-                        writer.addDependency(SmithyPythonDependency.SMITHY_HTTP.withOptionalDependencies("aiohttp"));
+                        writer.addDependency(SmithyPythonDependency.SMITHY_HTTP);
+                        writer.addDependency(SmithyPythonDependency.AIO_HTTP);
```
nateprewitt, line 151: rereading this, I think this is slightly incorrect, but assuming the other conditional triggers now isn't a hard blocker for right now.

What we want this to look like eventually is:

`smithy-http[awscrt]`
or
`smithy-http[aiohttp]`

where the http client version is declared in smithy-http and we're not taking additional dependencies. I think we can track that as follow up wort, I'm not sure `.withOptionalDependencies("aiohttp")` was actually working before.

PR #414, opened 2025-03-09
Fix AsyncBytesReader, read when bytes are already available

alextwoods:
*Description of changes:*
Output event streams were hanging when trying to read bytes from the AsyncBytesReader when `self._remainder` already had sufficient bytes to satisfy the read.

### DIFF
--- a/packages/smithy-core/src/smithy_core/aio/types.py
+++ b/packages/smithy-core/src/smithy_core/aio/types.py
@@ -72,10 +72,11 @@ class AsyncBytesReader:
             self._remainder = b""
             return result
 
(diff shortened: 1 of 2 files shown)


nateprewitt (dismissed):
This is a behavior we should probably have a test for to make sure it continues to work like we expect. Doesn't necessarily need to be done here but we should have an item to come back and write up tests for this `read()` impl.

JordonPhillips (dismissed):
+1 to adding a test

JordonPhillips (approved):

nateprewitt (approved):

PR #416, opened 2025-03-10
Create client transport interface

JordonPhillips:
This adds a generic interface to represent client transports, such as HTTP clients. It additionally makes the HTTP client implement that interface explicitly.

A change to the HTTP client interface as a result of this is that `request` is no longer a keyword-only argument. While I understand and agree with the general desire to restrict positional arguments, I think adhering to that too dogmatically can make things less usable to no gain.


haydenbaker (approved):

haydenbaker (changes requested):
Protocol test generation needs to be updated looks like

haydenbaker (approved):
Protocol tests generation needs to be updated

PR #417, opened 2025-03-10
Use ruff formatter in generated clients

JordonPhillips:
This updates the formatter used for generated clients to match the one we use for the static python code.


haydenbaker (approved):
🙌

nateprewitt (approved):

PR #418, opened 2025-03-10
Add sphinx doc generation 

SamRemis:
This PR adds generation of sphinx docs.  There are a number of changes that are necessary to accomplish this:

1) Converting the HTML provided by service teams to RST for sphinx (`AwsDocConverter.java`)
2) Updating the linewrapping to respect RST formatting (`PythonWriter.java`)
3) Add the core sphinx files to be generated during the initial repo setup (`SetupGenerator.java`)
4) Add events during client generation for when we are generating a symbol that needs to be documented (many files- most of the `*Generator.java` files as well as the `*Section.java` files)
5) Add more necessary boiler plate code- requirements, references to the new interceptors, etc.

This is still in draft mode pending the following action items:
1) Add tests


codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/StructureGenerator.java
```diff
@@ -309,7 +309,7 @@ private boolean hasDocs() {
     private void writeMemberDocs(MemberShape member) {
         member.getMemberTrait(model, DocumentationTrait.class).ifPresent(trait -> {
             String memberName = symbolProvider.toMemberName(member);
-            String docs = writer.formatDocs(String.format(":param %s: %s", memberName, trait.getValue()));
+            String docs = writer.formatDocs(String.format(":param %s: %s", memberName, trait.getValue())).replace("\n", "\n\t");
```
nateprewitt, line 312: Does this add a literal tab or 4 spaces? We should try to get everything in the generator to spaces, I'm not sure if the code formatter is handling that for us.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDocConverter.java
```diff
@@ -0,0 +1,146 @@
...
+import software.amazon.smithy.model.transform.ModelTransformer;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.nodes.Node;
+import org.jsoup.nodes.TextNode;
+import org.jsoup.select.NodeVisitor;
+
+/**
+ * Adds a runtime plugin to set user agent.
```
SamRemis, line 22: Need to update this comment, and add a few more; already have it locally, but will add it to a future revision

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
```diff
@@ -125,9 +130,12 @@ raise NotImplementedError()
 
                 """, unknownSymbol.getName());
         memberNames.add(unknownSymbol.getName());
+        writer.popState();
 
-        writer.write("type $L = $L\n", parentName, String.join(" | ", memberNames));
+        writer.pushState(new UnionSection(shape, parentName, memberNames));
+        writer.write("$L = Union[$L]\n", parentName, String.join(" | ", memberNames));
```
SamRemis, line 136: Need to revert to the older union syntax; autodoc doesn't natively support the new syntax yet: https://github.com/sphinx-doc/sphinx/issues/10785

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
```diff
@@ -255,4 +250,193 @@ private static void writeReadme(
...
+    ) {
+        String version = settings.moduleVersion();
+        context.writerDelegator().useFileWriter("docs/conf.py", "", writer -> {
+            writer.write("""
+                    import os
+                    import sys
+                    sys.path.insert(0, os.path.abspath('..'))
+
+                    project = 'AWS SDK for Python'
+                    author = 'Boto'
```
nateprewitt, line 283:
```suggestion
                    author = 'Amazon Web Services'
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
```diff
@@ -4,11 +4,10 @@
  */
 package software.amazon.smithy.python.codegen.writer;
 
-import java.util.Map;
```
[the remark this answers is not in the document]
nateprewitt: Sorry, not sure I understand this change. Why are we changing the imports? All of these are used, right?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
```diff
@@ -141,9 +150,78 @@ public PythonWriter writeDocs(String docs) {
...
+        }
+        wrappedText.append(indentStr).append(line).append("\n");
+    }
+
+    private static int findWrapPosition(String line, int maxLineLength) {
+        // Find the last space before maxLineLength
+        int wrapAt = line.lastIndexOf(' ', maxLineLength);
+        if (wrapAt == -1) {
+            // If no space found, force wrap at maxLineLength
+            wrapAt = maxLineLength;
```
nateprewitt, line 187: How does this work for things like links and code that may stretch beyond the limit. Is this going to create rendering issues?

pyproject.toml
```diff
@@ -10,11 +10,13 @@ dependencies = []
 dev = [
     "black>=25.1.0",
     "docformatter>=1.7.5",
+    "pydata-sphinx-theme>=0.16.1",
```
nateprewitt, line 13: Do these belong at the top level here or should these be added into the generated projects dependencies?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
SamRemis: It does create some; with the current docs there is only one link I can find with this issue.  

What is the ideal behavior here?  Should we allow these docs to exceed the max line length?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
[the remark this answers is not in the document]
SamRemis: Something weird may have happened during rebase.  It looks like they are used now; they will be added back on a later revision.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
nateprewitt: If it's a link, probably. That only affects rendering in the code. There's a question of if we even want the links in the docstrings, but I don't think we need to solve that now.

codegen/aws/core/build.gradle.kts
```diff
@@ -12,4 +12,5 @@ extra["moduleName"] = "software.amazon.smithy.python.aws.codegen"
 dependencies {
     implementation(project(":core"))
     implementation(libs.smithy.aws.traits)
+    implementation("org.jsoup:jsoup:1.19.1")
```
JordonPhillips, line 15: The version should go in [here](https://github.com/smithy-lang/smithy-python/blob/develop/codegen/gradle/libs.versions.toml). Docs on how that works are [here](https://docs.gradle.org/current/userguide/version_catalogs.html).

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDocConverter.java
```diff
@@ -0,0 +1,162 @@
...
+
+    String convertHtmlToRst(String html) {
+        Document document = Jsoup.parse(html);
+        RstNodeVisitor visitor = new RstNodeVisitor();
+        document.body().traverse(visitor);
+        return "\n" + visitor;
+    }
+
+    private static class RstNodeVisitor implements NodeVisitor {
+        private final StringBuilder sb = new StringBuilder();
```
JordonPhillips, line 52: It'll probably be better to use a [SimpleCodeWriter](https://github.com/smithy-lang/smithy/blob/main/smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleCodeWriter.java) for this. There's a few major benefits. For starters, it will make indentation management easier. Right now any list values that happen to have a newline will break the indentation. Same for anything in any sort of block.

A secondary benefit is that you can then create a named section that can be pushed in the opening tag and popped in the closing tag. That'll let you intercept those sections later down the line if you need to do customizations. It'll also let you support ordered lists since you can put context into the writer to say which list element you're on. Since context is scoped to sections, you won't have to worry about overriding some parent.

Given time constraints we have right now, this can be a fast follow.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDocConverter.java
```diff
@@ -0,0 +1,162 @@
...
+@SmithyInternalApi
+public class AwsDocConverter implements PythonIntegration {
+    @Override
+    public Model preprocessModel(Model model, PythonSettings settings) {
+        return ModelTransformer.create().mapShapes(model, shape -> {
+            if (shape.hasTrait(DocumentationTrait.class)) {
+                DocumentationTrait docTrait = shape.getTrait(DocumentationTrait.class).get();
+                String html = docTrait.getValue();
+                String rst = convertHtmlToRst(html);
+                DocumentationTrait newDocTrait = new DocumentationTrait(rst);
```
JordonPhillips, line 34: I'd rather run this in `PythonWrtier` where we write documentation. The format of the documentation trait is commonmark, so putting RST in there isn't really correct.

Speaking of commonmark, you need to account for that. The easiest way is to just render the whole thing to HTML like [this](https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/DocumentationConverter.java#L73). commonmark can contain HTML (and most of the docs we get from AWS are just HTML) but there's more to it.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsRstDocFileGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+     * Utility method to generate a header for documentation files.
+     *
+     * @param title The title of the section.
+     * @return A formatted header string.
+     */
+    private static String generateHeader(String title) {
+        return String.format("%s%n%s%n%n", title, "=".repeat(title.length()));
+    }
+
+    private static final class OperationGenerationInterceptor
```
JordonPhillips, line 43: What's the advantage of doing this over just autoclassing the client itself?

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDocConverter.java
JordonPhillips, line 1: This doesn't need to be AWS specific, nor does the rst doc file generator

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
```diff
@@ -255,4 +262,199 @@ private static void writeReadme(
             writer.popState();
         });
     }
+
+    /**
+     * Write the files required for sphinx doc generation
+     */
+    private static void writeDocsSkeleton(
```
JordonPhillips, line 269: We don't need to do this now, but lets add a todo to make this configurable. We should be able to turn off generating all the sphinx setup files and select a theme.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
```diff
@@ -255,4 +262,199 @@ private static void writeReadme(
...
+            GenerationContext context
+    ) {
+        String version = settings.moduleVersion();
+        context.writerDelegator().useFileWriter("docs/conf.py", "", writer -> {
+            writer.write("""
+                    import os
+                    import sys
+                    sys.path.insert(0, os.path.abspath('..'))
+
+                    project = 'AWS SDK for Python'
```
JordonPhillips, line 301: This should reference the service title, falling back to the shape name. You could preprocess the model in the AWS side to set the title trait to be based on the SDK ID if you like.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
JordonPhillips: Can you add this reference as a comment in the code? That context will get lost I guarantee

JordonPhillips (changes requested):

JordonPhillips (dismissed):

JordonPhillips (dismissed):
Once this and that bug are fixed I'm approving

JordonPhillips (approved):

PR #421, opened 2025-03-10
Fix read/write order in AsyncBytesProvider

alextwoods:
*Description of changes:*
Previously the `AsyncBytesProvider` write was appending to a list and the async iter was `pop`'ing from the end of that list - effectively iterating through the chunks of data in reverse order.

This PR updates the `AsyncBytesProvider` to use a deque - write appends to it and the iterator uses `popleft` to return the fist item.

Also updates the existing `test_provider_reads_written_data` test to write multiple chunks of data.  Test fails without this change (returning [bar, foo] instead of [foo, bar]) and now passes.


nateprewitt (approved):
:shipit:

JordonPhillips (approved):

PR #423, opened 2025-03-11
Typed context

JordonPhillips:
This adds a typed context bag, which is essentially a dict with some extra type forwarding built in. It removes the need to pass around a bunch of `dict[str, Any]` all over the place without sacrificing the ease that doing so provides.

The intended use of this is to have statically defined typed properties, like so:

```python
SERVICE_ID = PropertyKey(key="service_id", value_type=str)
CONFIG_PROPERTY = PropertyKey(key="config", value=Config)

...

context.properties[SERVICE_ID] = "my-service"

...

# After plugins have modified config
context.properties[CONFIG] = config

...

service_id = context.properties[SERVICE_ID]
config = context.properties[CONFIG]
endpoint = endpoints.get(service_id, {}).get(config.region)
```

The property keys can then be shared so that nobody messes up the key or its typing, like we do in the code generator for symbol properties.


packages/smithy-core/src/smithy_core/interfaces/__init__.py
```diff
@@ -83,3 +93,91 @@ def is_streaming_blob(obj: Any) -> TypeGuard[StreamingBlob]:
...
+    Used with :py:class:`Context` to set and get typed values.
+
+    For a concrete implementation, see :py:class:`smithy_core.types.PropertyKey`.
+    """
+
+    key: str
+    """The string key used to access the value."""
+
+    value_type: type[T]
+    """The type that of the associated value in the properties bag."""
```
nateprewitt, line 111:
```suggestion
    """The type of the associated value in the properties bag."""
```

packages/smithy-core/src/smithy_core/types.py
```diff
@@ -153,3 +157,99 @@ def format(self, *args: object, **kwargs: str) -> str:
...
+
+@dataclass(kw_only=True, frozen=True, slots=True, init=False)
+class PropertyKey[T](_PropertyKey[T]):
+    """A typed property key."""
+
+    key: str
+    """The string key used to access the value."""
+
+    value_type: type[T]
+    """The type that of the associated value in the property bag."""
```
nateprewitt, line 170:
```suggestion
    """The type of the associated value in the property bag."""
```

haydenbaker (dismissed):

alextwoods (dismissed):
Nice! I like this.  I'll put out an update to the UserAgent to use a typed property once this is merged :-)

JordonPhillips:
Had to rebase

alextwoods (approved):

PR #424, opened 2025-03-11
De-duplicate Config (Fix duplicate Region)

alextwoods:
*Description of changes:*
Previously we collected `ConfigProperty` in a list, however, multiple integrations/runtime plugins may need to independently register and depend on the same configuration properties - an example of this is sigv4 auth and regional endpoints - they both depend on the `region` config and could potentially be used independently of each other, so should both register the config properties they need without trying to inspect or understand what other integrations/runtime plugins are being used.

This PR changes the data structure used to collect ConfigProperties during generation from a `List` to a `TreeSet` which uses the properties `name` to de-dupe it.  

Currently, the *first* registered config value with a specific name will be preserved, which does not allow runtime plugins to override base config properties (eg:  with this ordering, its not possible for a runtime plugin to override the the `http_client`). 


nateprewitt (approved):
🚀

PR #427, opened 2025-03-11
Add host header when unset to crt requests

alextwoods:
*Description of changes:*
We were previously not setting the `host` header which was causing requests to fail.  The underlying CRT client does not set it.  The aihttp client does set this automatically so our adapter didn't need to set it explicitly.

TODO: Needs added tests.


nateprewitt:
😬 wonderful, I think every Python client does this for you so the CRT is an outlier here. I think this is fine conceptually but I know some libraries don't play nicely with the end-user setting this. Did we validate `aiohttp` still works with this change?

nateprewitt (dismissed):

jonathan343 (approved):
LGTM

nateprewitt (approved):

PR #434, opened 2025-03-12
Pass new request into signer for canonical_request

nateprewitt:
*Description of changes:*
We shouldn't be mutating the base request during signing, the intended workflow is to only use the copied request.


alextwoods (commented):
I think we also need to update below: `signing_fields = await self._normalize_signing_fields(request=http_request)`

alextwoods (dismissed):

alextwoods (dismissed):

alextwoods (approved):

PR #436, opened 2025-03-12
Remove use of thread locks in crt bindings

JordonPhillips:
This primarily serves the purpose of removing thread-based locks from the CRT bindings, but it also ensures that the status code and headers have been received before returning from send.

Thread-based `Lock`s will deadlock if acquired twice by the same thread, so it's likely to happen in async code that runs multiple coroutines in one thread. And also, `deque`s are already thread-safe so there's no need to protect them further.

### DIFF
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -156,7 +156,6 @@ class CRTResponseFactory:
     def on_response(
         self, status_code: int, headers: list[tuple[str, str]], **kwargs: Any
     ) -> None:  # pragma: crt-callback
-        print("on-response")
         fields = Fields()
         for header_name, header_val in headers:
             try:
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -176,7 +176,6 @@ class CRTResponseFactory:
         )
 
     async def await_response(self) -> AWSCRTHTTPResponse:
-        print(f"Initial status: {self._response_future._state}")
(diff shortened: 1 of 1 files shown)


alextwoods (dismissed):

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -61,25 +60,103 @@ def _initialize_default_loop(self) -> "crt_io.ClientBootstrap":
...
+
+class CRTResponseFactory:
+    def __init__(self, body: CRTResponseBody) -> None:
+        self._body = body
+        self._response_future = ConcurrentFuture[AWSCRTHTTPResponse]()
+
+    def on_response(
         self, status_code: int, headers: list[tuple[str, str]], **kwargs: Any
     ) -> None:  # pragma: crt-callback
+        print("on-response")
```
nateprewitt, line 159:
```suggestion
```

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -90,76 +167,25 @@ def _on_headers(
...
+            )
+        )
 
-    @property
-    def reason(self) -> str | None:
-        """Optional string provided by the server explaining the status."""
-        # TODO: See how CRT exposes reason.
-        return None
+    async def await_response(self) -> AWSCRTHTTPResponse:
+        print(f"Initial status: {self._response_future._state}")
```
nateprewitt, line 180:
```suggestion
```

nateprewitt (changes requested):

nateprewitt (approved):

alextwoods (approved):

PR #438, opened 2025-03-13
Event Signing

nateprewitt:
This PR is a draft implementation of Event signing in `aws-sdk-signers` and some temporary scaffolding changes to enable their usage. We have a few spots right now where our implementations don't align nicely. I've added some shims to help test while we discuss a longer term solution.

----

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -571,6 +571,9 @@ final class ClientGenerator implements Runnable {
         writer.pushState(new SignRequestSection());
         if (context.applicationProtocol().isHttpProtocol() && supportsAuth) {
             writer.addStdlibImport("re");
+            writer.addStdlibImport("typing", "Any");
+            writer.addImport("smithy_core.interfaces.identity", "Identity");
+            writer.addImport("smithy_core.types", "PropertyKey");
             writer.write("""
                             # Step 7i: sign the request
                             if auth_option and signer:
@@ -594,8 +597,11 @@ final class ClientGenerator implements Runnable {
                                 auth_value = fields["Authorization"].as_string()  # type: ignore
                                 signature = re.split("Signature=", auth_value)[-1]  # type: ignore
                                 context.properties["signature"] = signature.encode('utf-8')
-                                context.properties["identity"] = identity
-                                context.properties["signer_properties"] = auth_option.signer_properties
+
+                                identity_key = PropertyKey(key="identity", value_type=Identity | None)
+                                sp_key = PropertyKey(key="signer_properties", value_type=dict)
+                                context.properties[identity_key] = identity
+                                context.properties[sp_key] = auth_option.signer_properties
                     """);
         }
         writer.popState();
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -598,8 +598,14 @@ final class ClientGenerator implements Runnable {
                                 signature = re.split("Signature=", auth_value)[-1]  # type: ignore
                                 context.properties["signature"] = signature.encode('utf-8')
 
-                                identity_key = PropertyKey(key="identity", value_type=Identity | None)
-                                sp_key = PropertyKey(key="signer_properties", value_type=dict)
+                                identity_key: PropertyKey[Identity | None] = PropertyKey(
+                                    key="identity",
+                                    value_type=Identity | None  # type: ignore
+                                )
+                                sp_key: PropertyKey[dict[str, Any]] = PropertyKey(
+                                    key="signer_properties",
+                                    value_type=dict[str, Any]  # type: ignore
+                                )
                                 context.properties[identity_key] = identity
                                 context.properties[sp_key] = auth_option.signer_properties
                     """);
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -570,6 +570,8 @@ final class ClientGenerator implements Runnable {
 
         writer.pushState(new SignRequestSection());
         if (context.applicationProtocol().isHttpProtocol() && supportsAuth) {
+            writer.addStdlibImport("binascii", "hexlify");
+            writer.addStdlibImport("re");
             writer.write("""
                             # Step 7i: sign the request
                             if auth_option and signer:
@@ -587,6 +589,14 @@ final class ClientGenerator implements Runnable {
                                     )
                                 )
                                 logger.debug("Signed HTTP request: %s", context.transport_request)
+
+                                # TODO - Move this to separate resolution/population function
+                                fields = context._transport_request.fields
+                                auth_value = fields["Authorization"].as_string()  # type: ignore
+                                signature = re.split("Signature=", auth_value)[-1]  # type: ignore
+                                context._properties["signature"] = hexlify(signature.encode('utf-8'))  # type: ignore
+                                context._properties["identity"] = identity
+                                context._properties["signer_properties"] = auth_option.signer_properties
                     """);
         }
         writer.popState();
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -570,7 +570,6 @@ final class ClientGenerator implements Runnable {
 
         writer.pushState(new SignRequestSection());
         if (context.applicationProtocol().isHttpProtocol() && supportsAuth) {
-            writer.addStdlibImport("binascii", "hexlify");
             writer.addStdlibImport("re");
             writer.write("""
                             # Step 7i: sign the request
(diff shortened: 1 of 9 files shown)


packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -348,7 +353,7 @@ def _should_sha256_sign_payload(
         if request.destination.scheme != "https":
             return True
 
-        return signing_properties.get("payload_signing_enabled", True)
+        return signing_properties.get("payload_signing_enabled", False)
```
alextwoods, line 356: I think this still needs to default to `True` for now.  `UNSIGNED-PAYLOAD` is only supported for operations with the [@unsignedPayload trait](https://smithy.io/2.0/aws/aws-auth.html#aws-auth-unsignedpayload-trait).

For now, none of the services we need to support have the unsignedPayload trait on any operations, so I think the best course is to revert this default to `True`.

Long term - the auth resolver will likely need to check the runtime model for the operation to inspect the traits and set this property correctly per operation. (Note - I've created an entry in our tracker for this).

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -553,6 +554,14 @@ await sleep(retry_token.retry_delay)
...
                                 )
                                 logger.debug("Signed HTTP request: %s", context.transport_request)
+
+                                # TODO - Move this to separate resolution/population function
+                                fields = context.transport_request.fields
+                                auth_value = fields["Authorization"].as_string()  # type: ignore
+                                signature = re.split("Signature=", auth_value)[-1]  # type: ignore
+                                context.properties["signature"] = signature.encode('utf-8')
+                                context.properties["identity"] = identity
+                                context.properties["signer_properties"] = auth_option.signer_properties
```
alextwoods, line 564: Nit: These should probably use the typed properties.

These are generic and I think something reasonable to always have after auth resolution, so I think we can leave them here.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -553,6 +554,14 @@ await sleep(retry_token.retry_delay)
                                     )
                                 )
                                 logger.debug("Signed HTTP request: %s", context.transport_request)
+
+                                # TODO - Move this to separate resolution/population function
+                                fields = context.transport_request.fields
+                                auth_value = fields["Authorization"].as_string()  # type: ignore
+                                signature = re.split("Signature=", auth_value)[-1]  # type: ignore
+                                context.properties["signature"] = signature.encode('utf-8')
```
alextwoods, line 562: The sigv4 specific bits here we can add a code interceptor that appends to the SignRequestSection into the AwsAuthIntegration.

Though I'm not sure how this fits into the pure python request pipeline.

packages/aws-event-stream/src/aws_event_stream/events.py
```diff
@@ -387,7 +387,7 @@ def get_result(self) -> bytes:
             raise InvalidHeadersLength(len(result))
         return result
 
-    def encode_headers(self, headers: HEADERS_DICT):
+    def encode_headers(self, headers: HEADERS_DICT) -> None:
```
alextwoods, line 390: I understand why this method builds an internal buffer - but the interface does seem weird.  It seems like we could either:
1. have this as a class method which would construct an instance, call _encode_headers and then return the result. or
2. Have a private class that handles the build/state, _EventHeaderEncoder which encode_headers would instantiate and return the result from.

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -789,6 +800,113 @@ async def _compute_payload_hash(
...
+        event_message: "EventMessage",
+        event_encoder_cls: type["EventHeaderEncoder"],
+    ) -> "EventMessage":
+        async with self._signing_lock:
+            # Copy and prepopulate any missing values in the
+            # signing properties.
+            new_signing_properties = SigV4SigningProperties(  # type: ignore
+                **self._signing_properties
+            )
+            # TODO: If date is in properties, parse a datetime from it.
```
alextwoods, line 828: We might want to fix this now since the other signing methods do this and I think we'd probably want it for testing.  I think we should be able to use `_normalize_signing_properties` to create the new_signing_properties and then parse date_obj from it.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
nateprewitt: I think we're entirely refactoring this with Jordon's upcoming work. Ideally we're not doing this at all going forward.

packages/aws-event-stream/src/aws_event_stream/events.py
nateprewitt: Yeah, I'd generally agree with this, the encode workflow here is a bit odd. I think we can look at a refactor as follow up, this PR is just fixing the typing issue from the original.

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
```diff
@@ -36,16 +36,23 @@
...
-type Signer = Callable[[EventMessage], EventMessage]
-"""A function that takes an event message and signs it, and returns it signed."""
+class EventSigner(Protocol):
+    """A signer to manage credentials and EventMessages for an Event Stream lifecyle."""
+
+    def sign_event(
+        self,
+        *,
+        event_message: EventMessage,
+        event_encoder_cls: type[EventHeaderEncoder],
```
JordonPhillips, line 46: This shouldn't be needed anymore right? It'll just get encoded later

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
```diff
@@ -36,16 +36,23 @@
 _DEFAULT_BLOB_CONTENT_TYPE = "application/octet-stream"
 
 
-type Signer = Callable[[EventMessage], EventMessage]
-"""A function that takes an event message and signs it, and returns it signed."""
+class EventSigner(Protocol):
+    """A signer to manage credentials and EventMessages for an Event Stream lifecyle."""
+
+    def sign_event(
```
JordonPhillips, line 42: I'd like to move this to the regular signer. We can create a central event dataclass or protocol to give it. That should solve the double initialization problem. That can be later though.

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -789,6 +800,113 @@ async def _compute_payload_hash(
...
+                    SIGV4_TIMESTAMP_FORMAT
+                )
+
+            timestamp = new_signing_properties["date"]
+            headers: dict[str, str | bytes | datetime.datetime] = {":date": date_obj}
+            encoder = event_encoder_cls()
+            encoder.encode_headers(headers)
+            encoded_headers = encoder.get_result()
+
+            payload = event_message.encode()
```
JordonPhillips, line 841: This doesn't seem right. You're signing the entire encoded event and then modifying it. Shouldn't this just be the payload?

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -789,6 +800,113 @@ async def _compute_payload_hash(
...
+            )
+            event_signature = await self._sign_event(
+                timestamp=timestamp,
+                string_to_sign=string_to_sign,
+                signing_properties=new_signing_properties,
+            )
+            headers[":chunk-signature"] = event_signature
+
+            event_message.headers = headers
+            event_message.payload = payload
```
JordonPhillips, line 858: This will only ever have two headers - the chunk signature and date. And then the payload is an entire previously encoded event. Am I missing something? This doubles the overhead of the encoding.

JordonPhillips (changes requested):

packages/aws-event-stream/src/aws_event_stream/events.py
JordonPhillips: Yeah this was intended for streaming out without an intermediary. Then we switched to have an intermediary

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
JordonPhillips: oh my god it is actually like this

alextwoods: This is correct, but weird.  When sigv4 signing an event, you encode the entire unsigned message as the payload to sign, and then the signed message will have only two headers: `:date` and `:chunk-signature`.

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
JordonPhillips: The only [docs we have for this](https://docs.aws.amazon.com/transcribe/latest/dg/streaming-setting-up.html) say to do it this way. That seems very inefficient.

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
alextwoods: I think its necessary to encode the new headers while building the string to sign.

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
alextwoods: We need to track signing state (previous signature) between events so theres some state management.  But we could also add previous signature as an input to sign_event and store it in the publisher.

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
JordonPhillips: Yeah this was before my Grand Revelation

packages/aws-event-stream/src/aws_event_stream/_private/serializers.py
nateprewitt: Yeah my original plan had been for this context to be in the Publisher rather than the signer. What we have implemented currently makes that a bit harder. I think raising it a level to let the publisher do the orchestration makes more sense long term though.

JordonPhillips: We can do that without muddying up the interface too much. Just make an `event_signer` method on the auth scheme class or whatever that's given the entire transport request and returns the signer to use. A stateless signer could just return itself.

JordonPhillips (approved):

PR #439, opened 2025-03-13
Return immediately on streaming inputs

JordonPhillips:
This adds (optional) request and response futures that are plumbed through the request pipeline. These are used to return event streams immediately when events are ready to be sent.

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -199,12 +199,12 @@ final class ClientGenerator implements Runnable {
                                 config: $4T,
                                 operation_name: str,
                             ) -> Any:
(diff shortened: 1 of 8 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -342,6 +407,10 @@ await seek(0)
                                     else:
                                         # Step 8: Invoke record_success
                                         retry_strategy.record_success(token=retry_token)
+                                        if response_future is not None:
+                                            response_future.set_result(
+                                                context_with_response.response,  # type: ignore
```
alextwoods, line 412: Should this be the transport response instead?  IE:
```
response_future.set_result(context_with_response.transport_response)
```

alextwoods (approved):

nateprewitt (approved):

PR #443, opened 2025-03-14
Ensure streams get closed and don't wait on concurrent futures in crt

JordonPhillips:

### DIFF
--- a/packages/aws-event-stream/src/aws_event_stream/_private/deserializers.py
+++ b/packages/aws-event-stream/src/aws_event_stream/_private/deserializers.py
@@ -40,9 +40,18 @@ class AWSAsyncEventReceiver[E: DeserializeableShape](AsyncEventReceiver[E]):
         self._source = source
         self._is_client_mode = is_client_mode
         self._deserializer = deserializer
+        self._closed = False
 
     async def receive(self) -> E | None:
-        event = await Event.decode_async(self._source)
+        if self._closed:
+            return None
+
+        try:
+            event = await Event.decode_async(self._source)
+        except Exception as e:
+            await self.close()
+            raise IOError("Failed to read from stream.") from e
+
         if event is None:
             return None
 
@@ -57,9 +66,17 @@ class AWSAsyncEventReceiver[E: DeserializeableShape](AsyncEventReceiver[E]):
         return result
 
     async def close(self) -> None:
+        if self._closed:
+            return
+        self._closed = True
+
         if (close := getattr(self._source, "close", None)) is not None:
             if asyncio.iscoroutine(result := close()):
                 await result
+
+    @property
+    def closed(self) -> bool:
+        return self._closed
 
 
 class EventDeserializer(SpecificShapeDeserializer):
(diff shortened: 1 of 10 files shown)


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -128,21 +133,21 @@ def on_body(self, chunk: bytes, **kwargs: Any) -> None:  # pragma: crt-callback
...
     async def next(self) -> bytes:
-        if self._stream is None:
+        if self._completion_future is None:
             raise SmithyHTTPException("Stream not set")
 
         # TODO: update backpressure window once CRT supports it
         if self._received_chunks:
             return self._received_chunks.popleft()
-        elif self._stream.completion_future.done():
+        elif self._completion_future.done():
```
JordonPhillips, line 142: Note that async futures return immediately on done, whereas concurrent futures can wait... forever

packages/aws-event-stream/src/aws_event_stream/_private/deserializers.py
```diff
@@ -40,9 +40,18 @@ def __init__(
...
     async def receive(self) -> E | None:
-        event = await Event.decode_async(self._source)
+        if self._closed:
+            return None
+
+        try:
+            event = await Event.decode_async(self._source)
+        except Exception as e:
+            await self.close()
+            raise IOError("Failed to read from stream.") from e
```
alextwoods, line 53: Exceptions could be raised here for other decode issues right (ie: like an invalid message), rather than just failing to read from the stream?

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -343,13 +357,11 @@ async def _marshal_request(
             crt_body = BytesIO(body)
         else:
             # If the body is async, or potentially very large, start up a task to read
-            # it into the BytesIO object that CRT needs. By using asyncio.create_task
-            # we'll start the coroutine without having to explicitly await it.
+            # it into the intermediate object that CRT needs. By using
+            # asyncio.create_task we'll start the coroutine without having to
+            # explicitly await it.
             crt_body = BufferableByteStream()
-            if not isinstance(body, AsyncIterable):
```
alextwoods, line 349: Is there a reason we're always wrapping the body here? If its already an async iterable, it seems like we don't need to wrap it.

JordonPhillips: So we can close it. `AsyncByteStream` can handle calling close on the source if it's available. I can unroll that though

alextwoods: Yeah, it looks like there is a bug in [AsyncBytesReader._read_from_iterable](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/src/smithy_core/aio/types.py#L64) - specifically:

```
        if len(result) < size:
            async for element in iterator:
                result += element
                if len(result) >= size:
                    break
```

The `async for element in iterator` will block until we've read at least a full chunk size of data rather than returning what data is available without waiting.  This can be an issue when say, an event that you are trying to send is smaller than the chunk size and you aren't sending another event immediately after - it will hang forever here.  The `AsyncBytesProvider` that is used for the body in event stream operations (and is what is getting wrapped in these cases), provides chunks as soon as they are available regardless of size.

alextwoods (approved):

PR #422, opened 2025-03-11
Add protocol implementations, operation shape generation, type registry

haydenbaker:
Very WIP, putting up just for visibillity

Several TODOs, organization still TBD

Example operation type:
```
@dataclass(kw_only=True, frozen=True)
class GetCity(APIOperation["GetCityInput", "GetCityOutput"]):
    input = GetCityInput
    output = GetCityOutput
    schema = _SCHEMA_GET_CITY
    input_schema = _SCHEMA_GET_CITY_INPUT
    output_schema = _SCHEMA_GET_CITY_OUTPUT
    error_registry = TypeRegistry(
        {
            ShapeID("example.weather#NoSuchResource"): NoSuchResource,
        }
    )
    effective_auth_schemes = [ShapeID("smithy.api#noAuth")]

```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #444, opened 2025-03-14
Add a type registry

haydenbaker:
*Description of changes:*
- Adds `TypeRegistry`, which will be mainly used for enabling error serde
- This is small subset of changes originally drafted in #422, which I am breaking up to make incremental


packages/smithy-core/src/smithy_core/documents.py
```diff
@@ -633,3 +638,24 @@ def read_document(self, schema: "Schema") -> Document:
     @override
     def read_timestamp(self, schema: "Schema") -> datetime.datetime:
         return self._value.as_timestamp()
+
+
+# A registry for on-demand deserialization of types by using a mapping of shape IDs to their deserializers.
+class TypeRegistry:
```
JordonPhillips, line 644:
```suggestion
class TypeRegistry:
    """A registry for on-demand deserialization of types by using a mapping of shape
    IDs to their deserializers.
    """
```

packages/smithy-core/src/smithy_core/documents.py
```diff
@@ -633,3 +638,24 @@ def read_document(self, schema: "Schema") -> Document:
...
+    def __init__(
+        self,
+        types: dict[ShapeID, type[DeserializeableShape]],
+        sub_registry: "TypeRegistry | None" = None,
+    ):
```
JordonPhillips, line 649:
```suggestion
    def __init__(
        self,
        types: Mapping[ShapeID, type[DeserializeableShape]],
        sub_registry: "TypeRegistry | None" = None,
    ):
        """Initialize a TypeRegistry.
        
        :param types: A mapping of ShapeID to the shapes they deserialize to.
        :param sub_registry: A registry to delegate to if an ID cannot be found in
            types.
        """
```

Or something like that

packages/smithy-core/src/smithy_core/documents.py
```diff
@@ -633,3 +638,24 @@ def read_document(self, schema: "Schema") -> Document:
...
+        self._types = types
+        self._sub_registry = sub_registry
+
+    def get(self, shape: ShapeID) -> type[DeserializeableShape]:
```
JordonPhillips, line 653: This class should probably also implement `__getitem__` and `__contains__`. Also these methods need docs.

JordonPhillips (changes requested):

packages/smithy-core/src/smithy_core/documents.py
```diff
@@ -640,22 +640,53 @@ def read_timestamp(self, schema: "Schema") -> datetime.datetime:
...
+    def __getitem__(self, shape: ShapeID):
+        """Get the deserializable shape for the given shape ID.
+
+        :param shape: The shape ID to get from the registry.
+        :returns: The corresponding deserializable shape.
+        :raises KeyError: If the shape ID is not found in the registry.
+        """
+        return self.get(shape)
+
+    def __contains__(self, shape: ShapeID):
```
JordonPhillips, line 682:
```suggestion
    def __contains__(self, item: object, /):
```

I really wish type checkers would check this. Dunder methods in general aren't well supported and I don't know why. It's very easy to get this wrong. In fact we only do it right in 2 of 6 usages. We also have one `Any` which is actually more wrong. And I'm almost certainly guilty of all or nearly all of those error cases.

In this particular case the implementation doesn't rely on the given item being a `ShapeID` so there's no actual runtime issue.

JordonPhillips (changes requested):
One tiny change

packages/smithy-core/src/smithy_core/documents.py
haydenbaker: Oh interesting, thanks for the explanation too.

JordonPhillips (dismissed):

JordonPhillips (approved):

PR #446, opened 2025-03-16
[do not merge] Event signing fixes

alextwoods:
*Description of changes:*
Builds off of #438  and #443.

Tested working roundtrip (multiple input and output) events with lex.

Summary of changes:
* In resjson, if operation is event stream input, set the content type and content sha256 correctly (Note, we should potentially check if the service has sigv4 auth before setting the content sha256)
* Use a timestamp for the :date header.  This will serialize it as an epoch ms.
* Encode the entire message to be signed and use that as the payload of the signed event.  The signed event then only has two headers (:date, and :chunk-signature). 
* Fixes to STS:
  * change first bit to + "AWS4-HMAC-SHA256-PAYLOAD\n"
  * remove hexlify on prior signature everywhere.
  * remove final \n

There is also possibly a bug in AsyncBytesReader.  This change removes the wrapping of the body in AsyncBytesReader in crt.py when constructing the request (more debugging is required there).

TODO:
* If date is in properties, parse a datetime from it.
* Debug AsyncBytesReader

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/RestJsonProtocolGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/RestJsonProtocolGenerator.java
@@ -390,48 +390,64 @@ public class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
     }
 
     @Override
-    public void wrapEventStream(GenerationContext context, PythonWriter writer) {
+    public void wrapInputStream(GenerationContext context, PythonWriter writer) {
         writer.addDependency(SmithyPythonDependency.SMITHY_JSON);
         writer.addDependency(SmithyPythonDependency.AWS_EVENT_STREAM);
-        writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
-        writer.addImports("aws_event_stream.aio",
-                Set.of(
-                        "AWSDuplexEventStream",
-                        "AWSInputEventStream",
-                        "AWSOutputEventStream"));
         writer.addImport("smithy_json", "JSONCodec");
         writer.addImport("smithy_core.aio.types", "AsyncBytesReader");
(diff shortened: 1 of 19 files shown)


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -126,6 +126,7 @@ def set_stream(self, stream: "crt_http.HttpClientStream") -> None:
 
     def on_body(self, chunk: bytes, **kwargs: Any) -> None:  # pragma: crt-callback
         # TODO: update back pressure window once CRT supports it
+        print("<-----", chunk)
```
alextwoods, line 129: Todo: need to remove these, but they are useful for (almost) wire logging.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/RestJsonProtocolGenerator.java
```diff
@@ -156,6 +157,21 @@ protected void serializeDocumentBody(
...
+    protected void writeDefaultHeaders(GenerationContext context, PythonWriter writer, OperationShape operation) {
+        var eventStreamIndex = EventStreamIndex.of(context.model());
+        if (eventStreamIndex.getInputInfo(operation).isPresent()) {
+            writer.addImport("smithy_http", "Field");
+            writer.write(
+                    "Field(name=\"Content-Type\", values=[$S]),",
+                    "application/vnd.amazon.eventstream");
+            writer.write(
+                    "Field(name=\"X-Amz-Content-SHA256\", values=[$S]),",
+                    "STREAMING-AWS4-HMAC-SHA256-EVENTS\"");
```
nateprewitt, line 170:
```suggestion
                    "STREAMING-AWS4-HMAC-SHA256-EVENTS");
```

nateprewitt:
I've finished porting this back to #438. Resolving this one now that cleanup is done.

PR #448, opened 2025-03-18
Add http and restjson1 client protocols

haydenbaker:
*Description of changes:*
- Adds the client protocol implementations
- Several TODOs, some not necessarily needed and some are blocked by other work (e.g. endpoint properties) 
- Codegen will use the protocol when setting up the client request pipeline
  - for restjson, this could hypothetically look like:
```
class Weather:

    pipeline: RequestPipeline[HTTPRequest, HTTPResponse] = RequestPipeline(RestJsonClientProtocol(), ...)

    async def get_city(...) -> GetCityOutput:
        ...
        call = ClientCall(...)
        result = pipeline(call)
        ...
 
```

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
@@ -1,7 +1,7 @@
 from typing import Literal
 
 from smithy_aws_core.traits import RestJson1Trait
-from smithy_core.protocols import HttpBindingClientProtocol
+from smithy_http.aio.protocols import HttpBindingClientProtocol
 from smithy_core.codecs import Codec
 from smithy_core.shapes import ShapeID
 from smithy_json import JSONCodec
--- a/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
@@ -1,4 +1,4 @@
-from typing import Literal
+from typing import Final
 
 from smithy_aws_core.traits import RestJson1Trait
 from smithy_http.aio.protocols import HttpBindingClientProtocol
@@ -12,7 +12,7 @@ class RestJsonClientProtocol(HttpBindingClientProtocol):
 
     _id: ShapeID = RestJson1Trait.id
     _codec: JSONCodec = JSONCodec()
-    _contentType: Literal["application/json"] = "application/json"
+    _contentType: Final = "application/json"
 
     @property
     def id(self) -> ShapeID:
--- a/packages/smithy-aws-core/src/smithy_aws_core/traits.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/traits.py
@@ -1,0 +1,41 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+
+# This ruff check warns against using the assert statement, which can be stripped out
+# when running Python with certain (common) optimization settings. Assert is used here
+# for trait values. Since these are always generated, we can be fairly confident that
+# they're correct regardless, so it's okay if the checks are stripped out.
+# ruff: noqa: S101
+
+from dataclasses import dataclass, field
+from typing import Mapping, Sequence
+
+from smithy_core.shapes import ShapeID
+from smithy_core.traits import Trait, DocumentValue, DynamicTrait
+
+
+@dataclass(init=False, frozen=True)
+class RestJson1Trait(Trait, id=ShapeID("aws.protocols#restJson1")):
+    http: set[str] = field(repr=False, hash=False, compare=False, default_factory=set)
+    eventStreamHttp: set[str] = field(
+        repr=False, hash=False, compare=False, default_factory=set
+    )
+
+    def __init__(self, value: DocumentValue | DynamicTrait = None):
+        super().__init__(value)
+        assert isinstance(self.document_value, Mapping)
+
+        assert isinstance(self.document_value["http"], Sequence)
+        for val in self.document_value["http"]:
+            assert isinstance(val, str)
+            self.http.add(val)
+
+        if vals := self.document_value.get("eventStreamHttp") is None:
+            object.__setattr__(self, "eventStreamHttp", self.http)
+        else:
+            # check that eventStreamHttp is a subset of http
+            assert isinstance(vals, Sequence)
+            for val in self.document_value["eventStreamHttp"]:
+                assert val in self.http
+                assert isinstance(val, str)
+                self.eventStreamHttp.add(val)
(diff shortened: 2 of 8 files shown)


packages/smithy-aws-core/src/smithy_aws_core/traits.py
```diff
@@ -0,0 +1,41 @@
...
+
+        if vals := self.document_value.get("eventStreamHttp") is None:
+            object.__setattr__(self, "eventStreamHttp", self.http)
+        else:
+            # check that eventStreamHttp is a subset of http
+            assert isinstance(vals, Sequence)
+            for val in self.document_value["eventStreamHttp"]:
+                assert val in self.http
+                assert isinstance(val, str)
+                self.eventStreamHttp.add(val)
```
haydenbaker, line 41: We already have validators to validate this kind of constraint model-side, but this would be good time to discuss whether we do any kind of validation in our trait definitions here. Thoughts?

packages/smithy-core/src/smithy_core/protocols.py
```diff
@@ -0,0 +1,113 @@
...
+            uri_builder.path = os.path.join(uri.path, uri_builder.path or "")
+        # TODO: merge headers from the endpoint properties bag
+        return request
+
+
+class HttpBindingClientProtocol(HttpClientProtocol):
+    """An HTTP-based protocol that uses HTTP binding traits."""
+
+    @property
+    def codec(self) -> Codec:
```
JordonPhillips, line 46:
```suggestion
    def payload_codec(self) -> Codec:
```

packages/smithy-core/src/smithy_core/protocols.py
```diff
@@ -0,0 +1,113 @@
...
+        *,
+        operation: APIOperation[OperationInput, OperationOutput],
+        input: OperationInput,
+        endpoint: URI,
+        context: TypedProperties,
+    ) -> HTTPRequest:
+        # TODO: request binding cache like done in SJ
+        serializer = HTTPRequestSerializer(
+            payload_codec=self.codec,
+            http_trait=operation.schema.expect_trait(HTTPTrait),  # TODO
```
JordonPhillips, line 69: What's this todo?

packages/smithy-core/src/smithy_core/protocols.py
```diff
@@ -0,0 +1,113 @@
...
+            endpoint_trait=operation.schema.get_trait(EndpointTrait),
+        )
+
+        input.serialize(serializer=serializer)
+        request = serializer.result
+
+        if request is None:
+            raise ValueError("Request is None")  # TODO
+
+        request.fields["content-type"].add(self.content_type)
```
JordonPhillips, line 79: This is going to depend on the payload. It's probably best to put this into the request serializer and just pass this in as the payload codec or something

packages/smithy-core/src/smithy_core/protocols.py
```diff
@@ -0,0 +1,113 @@
...
+        if not (200 <= response.status <= 299):  # TODO: extract to utility
+            # TODO: implement error serde from type registry
+            raise NotImplementedError
+
+        body = response.body
+        # TODO: extract to utility, seems common
+        if (read := getattr(body, "read", None)) is not None and iscoroutinefunction(
+            read
+        ):
+            body = BytesIO(await read())
```
JordonPhillips, line 103: This shouldn't be done unconditionally. It should only be for cases where there's no `streaming` member and the body itself isn't already a synchronous stream (i.e. something with a sync `read`)

JordonPhillips (changes requested):

packages/smithy-http/src/smithy_http/serializers.py
```diff
@@ -112,6 +112,10 @@ def begin_struct(self, schema: Schema) -> Iterator[ShapeSerializer]:
         ) is not None and not iscoroutinefunction(seek):
             seek(0)
 
+        # TODO: conditional on empty-ness and a param of the protocol?
+        headers = binding_serializer.header_serializer.headers
+        headers.append(("content-type", self._payload_codec.media_type))
```
JordonPhillips, line 117: If there is a payload member, it'll be the content-type of said member (`application/octet-stream`, `text/plain`, `MediaType` trait value, event stream content type, or codec content type). If there's no payload member, it's just the codec type.

The emptyness thing can be done later for sure. We might have to add some tracking to see if anything was serialized since inspecting the payload isn't a good idea.

JordonPhillips (changes requested):

packages/smithy-http/src/smithy_http/serializers.py
```diff
@@ -83,22 +85,33 @@ def begin_struct(self, schema: Schema) -> Iterator[ShapeSerializer]:
...
+                if (media_type := payload_member.get_trait(MediaTypeTrait)) is not None:
+                    content_type = media_type.value
                 payload = BytesIO()
                 payload_serializer = self._payload_codec.create_serializer(payload)
                 binding_serializer = HTTPRequestBindingSerializer(
                     payload_serializer, self._http_trait.path, host_prefix
                 )
                 yield binding_serializer
         else:
+            if self._get_eventstreaming_member(schema) is not None:
```
haydenbaker, line 113: Might be handy (and more efficient) if we can embed this kind of information in the schemas themselves when we code-generate them, instead of having to do checks like this.

JordonPhillips: `APIOperation.input_stream_member`

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -0,0 +1,27 @@
...
+from smithy_core.shapes import ShapeID
+from smithy_json import JSONCodec
+
+
+class RestJsonClientProtocol(HttpBindingClientProtocol):
+    """An implementation of the aws.protocols#restJson1 protocol."""
+
+    _id: ShapeID = RestJson1Trait.id
+    _codec: JSONCodec = JSONCodec()
+    _contentType: Literal["application/json"] = "application/json"
```
JordonPhillips, line 15: This isn't necessary, but pro tip:

```suggestion
    _contentType: Final = "application/json"
```

That makes it a `Literal` without having to type it twice.

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if uri.scheme:
+            uri_builder.scheme = uri.scheme
+        if uri.host:
+            uri_builder.host = uri.host
+        if uri.port and uri.port > -1:
+            uri_builder.port = uri.port
+        if uri.path:
+            uri_builder.path = os.path.join(uri.path, uri_builder.path or "")
+        # TODO: merge headers from the endpoint properties bag
```
JordonPhillips, line 39: It'll be under `headers` in the properties bag. Not necessary right now though.

JordonPhillips (dismissed):

JordonPhillips (approved):

PR #449, opened 2025-03-18
Update interceptors

JordonPhillips:
This updates interceptors to be dataclasses with stronger typing and updates the generated request pipeline to use them. The pipeline updates are fairly extensive, but it has the advantage of removing most of the casting that was littered all over the place. This also ensures the immutability of the explicit properties.


nateprewitt (approved):
Conceptually this all looks good, I need to go take a look at this compared to the other inflight work to see if we're breaking anything though.

alextwoods (approved):

PR #450, opened 2025-03-18
Centralize event streams

JordonPhillips:
To summarize what this PR does:

* smithy-event-stream package was removed, and its contents were moved into core.
* The event stream classes were changed from protocols to concrete classes. The expectation is that protocols implement the publisher and receiver, then those are combined into one of the three stream types.
* Methods were added to the `ClientProtocol` interface to create the publishers and receivers.
* code gen was updated to use the new types


designs/event-streams.md
```diff
@@ -236,11 +232,14 @@ class InputEventStream[I: SerializableShape, R](Protocol):
...
 
 
-class OutputEventStream[O: DeserializableShape, R](Protocol):
+class OutputEventStream[OE: DeserializeableShape, O: DeserializeableShape]:
 
-    output_stream: AsyncEventReceiver[O]
-
-    response: R
+    output_stream: EventReceiver[OE]
+    response: O
```
haydenbaker, line 238: Why differing names for the same thing? `response` vs `output`, how about `initial_response` to be extra clear about what it is?

haydenbaker (dismissed):

designs/event-streams.md
JordonPhillips: I changed it to `output`. I want this to be the same in all of them and initial response doesn't make sense for input streams.

haydenbaker (approved):

alextwoods (approved):

PR #451, opened 2025-03-18
Fix issue with double encode of path in canonical request

alextwoods:
*Description of changes:*
This change adds a signing configuration for `uri_encode_path` (defaults to True) and enables double encoding of path when constructing the canonical request.  

The sigv4 test suite for normalize-path/get-space is out of date from the [updated CRT test](https://github.com/awslabs/aws-c-auth/tree/01dd06acd2b8865a4a6bc232380ee69a042af47d/tests/aws-signing-test-suite/v4/get-space-normalized).  Specifically, the new request has the URI path in the request as `GET /example space/` (incldues the space and is NOT uri encoded).  The request we have in our older test suite has `GET /example%20space/` (starts off uri encoded).  We can't easily fix our request path  in the test suite as the `RawRequest` class we are using does not allow this, so without bigger changes to the test suite, we can't fix that.

Note that for S3 you do not normalize URI paths. For example, if you have a bucket with an object named my-object//example//photo.user, use that path. Normalizing the path to my-object/example/photo.user will cause the request to fail. For more information, see [Create a Canonical Request]( http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html#canonical-request) in the S3 API Reference.


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -349,6 +349,7 @@ async def _marshal_request(
                 headers_list.append((fld.name, val))
 
         path = self._render_path(request.destination)
+        print("CRT path: ", path)
```
nateprewitt, line 352:
```suggestion

```

nateprewitt (dismissed):
Minor cleanup, otherwise looks good.

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: Doh - I thought I removed all of these debugs....

nateprewitt (approved):

packages/smithy-http/src/smithy_http/aio/crt.py
haydenbaker: There is a lint rule that we can add to catch print statements: https://docs.astral.sh/ruff/rules/#flake8-print-t20

alextwoods: I did try to add `T20` to the list configured in our top level [pyproject.toml](https://github.com/smithy-lang/smithy-python/blob/develop/pyproject.toml#L54) but it didn't seem to catch usage of print, so I'm not sure what I did wrong.

PR #452, opened 2025-03-18
Implement a configurable credentials resolver chain

alextwoods:
*Description of changes:*
Implements a configurable credentials resolver chain and sets it as the default identity resolver in generated Config.

The credential chain is configured by providing tuples of callables for each credential source, the first returns true if credentials are available from that source, the second is a constructor for that source.  

Individual credential resolvers can be constructed from multiple sources (eg: web identity token from environment or profile) - having the sources be defined using callable's allows us to de-couple the different sources for a given credential resolver from the implementation of the resolver itself.


packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
```diff
@@ -0,0 +1,54 @@
...
+from smithy_core.exceptions import SmithyIdentityException
+from smithy_core.interfaces.identity import IdentityProperties
+
+import os
+
+
+def _env_creds_available() -> bool:
+    return bool(os.getenv("AWS_ACCESS_KEY_ID")) and bool(
+        os.getenv("AWS_SECRET_ACCESS_KEY")
+    )
```
nateprewitt, line 15: nit; this is probably the more pythonic way to do this check.
```suggestion
    return (
        "AWS_ACCESS_KEY_ID" in os.environ
        and "AWS_SECRET_ACCESS_KEY" in os.environ
    )
```

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
```diff
@@ -0,0 +1,54 @@
...
+        self, *, identity_properties: IdentityProperties
+    ) -> AWSCredentialsIdentity:
+        if self._credentials_resolver is not None:
+            return await self._credentials_resolver.get_identity(
+                identity_properties=identity_properties
+            )
+
+        for source in self._sources:
+            if source[0]():
+                self._credentials_resolver = source[1]()
```
nateprewitt, line 47: What are we actually calling here 😅 A state check and then a builder for the resolver?

I think this works, I'm wondering if we want to actually make a class for the abstraction rather than composing lists of functions. This is probably minor at this point, but may be worth spending some time on ergonomics later.

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
```diff
@@ -0,0 +1,54 @@
+from typing import Callable, List
```
JordonPhillips, line 1: missing copyright header

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
```diff
@@ -0,0 +1,54 @@
...
+type CredentialSource = tuple[Callable[[], bool], Callable[[], AWSCredentialsResolver]]
+_DEFAULT_SOURCES: list[CredentialSource] = [(_env_creds_available, _build_env_creds)]
+
+
+class CredentialsResolverChain(
+    IdentityResolver[AWSCredentialsIdentity, IdentityProperties]
+):
+    """Resolves AWS Credentials from system environment variables."""
+
+    def __init__(self, *, sources: List[CredentialSource] | None = None):
```
JordonPhillips, line 31: Sources should be a `Sequence`. But also, why isn't it just a sequence of credential resolvers that take the same identity?

JordonPhillips (changes requested):

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
alextwoods: I was using this check for both set and not-empty.  Its probably not strictly necessary to check not empty though.  The current SEP does not specify either way....

nateprewitt: If the value exists in the environment, we should probably use it. If it doesn't work, I think that's better than ignoring it and potentially using unintended credentials. That provides the least risk and requires the end user to resolve the ambiguity before moving forward.

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
nateprewitt: If we're concerned about perf for instantiation, you can always have a class method on the class reference. That can be used to check availability and then you have the class there to instantiate if we're ready. It's essentially the same abstraction just wrapped into a single input.

alextwoods: The other consideration in this is that construction of a given resolver is based on the source - and its not always a 1:1 mapping and the order of precedence is important.  As an example:  assuming a role from a web identity.  Assuming a role from a web identity token file can be specified either through the environment or through the shared config file and the order of precedence (limited just to static credentials and web identity) is:
1. static credentials from ENV
2. Web identity from ENV
3. static credentials from profile
4. web identity from profile.

alextwoods: See [commit](https://github.com/smithy-lang/smithy-python/pull/452/commits/b325e70b09cb3a950dd42db75d3c2e84e9147cf8) for defining source as a class/Protocol - definitely easier to understand/use I think.

nateprewitt: LGTM!

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/chain.py
```diff
@@ -0,0 +1,57 @@
...
+_DEFAULT_SOURCES: Sequence[CredentialsSource] = (
+    EnvironmentCredentialsSource(),
+    IMDSCredentialsSource(),
+)
+
+
+class CredentialsResolverChain(
+    IdentityResolver[AWSCredentialsIdentity, IdentityProperties]
+):
+    """Resolves AWS Credentials from system environment variables."""
```
nateprewitt, line 28: This isn't just from Environment variables right? I'd expect this to describe something more about resolving an identity from a priority ordered list of credential sources.

nateprewitt (commented):
Minor docstring change, otherwise I think this looks good. @jonathan343 did you have anything else you wanted to see?

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
JordonPhillips: Combining threads here - I'm not convinced that this `Source` concept is worth it. I see this as adding complexity unnecessarily. What I want to see is something dead simple, like:

```python
class ChainedIdentityResolver[I](IdentityResolver[I]):

    def __init__(self, resolvers: Sequence[IdentityResolver[I]]) -> None:
        self._resolvers = resolvers

    async def get_identity(self, *, properties: _TypedProperties) -> I:
        for resolver in self._resolvers:
            try:
                return await resolver.get_identity(properties=properties)
            except SmithyIdentityException as e:
                logger.debug(
                    "Failed to resolve identity from %s: %s", type(resolver), e
                )

        raise SmithyIdentityException("Failed to resolve identity from resolver chain.")
```

I'm currently working on changing the interfaces here, so maybe we should just punt on this for now.

alextwoods: For the resolvers we currently have (env and IMDS) that approach is fine.  I think theres three reasons it doesn't work as simply as that for the full aws credentials resolver chain
1. Availability and resolution are separate concerns - as an example - if a `credential_process` is configured, the process resolver should be considered "available" and should be used by the chain even if executing that processes fails/doesn't return credentials. (ie, we wouldn't move on to the next resolver if a configured credential process fails).
2. Sources don't map 1:1 to resolvers - resolvers like Assume role or web identity can be configured multiple different ways at different levels of precedence - ie, they are different sources. 
3. Separation of concerns - configuration sources vs resolver properties.  How and where we source configuration generally shouldn't be a concern for an individual resolver - that is, I'd argue that something like a web identity resolver shouldn't need to be concerned with where/how the token file is specified (env, shared config, ect) - it should just know how to use that. 

Again though - I think these are longer term concerns, a simple sequence of resolvers is sufficient for what is available now so I'm happy to drop this PR and leave it to future design work to describe how this should work.

packages/smithy-aws-core/tests/unit/credentials_resolvers/test_credentials_resolver_chain.py
jonathan343, line 1: I think we should have a test for fallback behavior in the credential chain. All tests I see only configure one source

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/imds.py
```diff
@@ -235,3 +240,14 @@ async def get_identity(
             account_id=account_id,
         )
         return self._credentials
+
+
+class IMDSCredentialsSource(CredentialsSource):
+    def is_available(self, config: AwsCredentialsConfig) -> bool:
+        # IMDS credentials should always be the last in the chain
+        # We cannot check if they available without actually making a call
```
nateprewitt, line 248:
```suggestion
        # We cannot check if they're available without actually making a call
```

nateprewitt (changes requested):
Just the two wording issues and otherwise code looks good to me!

nateprewitt (approved):

packages/smithy-aws-core/src/smithy_aws_core/credentials_resolvers/credentials_resolver_chain.py
alextwoods: With something like a configured `credential_process`- it was explicitly configured and I'd argue that it should be treated as authoritative if present.  However, looking at existing SDK behavior, this seems to be underspecified and likely inconsistent across SDKs - both the Ruby and Java SDKs will resolve process credentials and fail during a request, surfacing the command failure in that case, but I'm guessing there are other SDKs that might move on in the chain....

We could potentially use different exception types to signal whether credentials are applicable/configured vs invalid, but generally I'd prefer not to use exceptions for control flow.  

On separation of concerns, I see resolvers as consumers of configuration rather than being responsible for doing the resolution of configuration - I think making an identity resolver responsible both for resolving configuration and for then resolving an identity from it increases the complexity and mixes two separate concerns.  

That said again, all of those are longer term concerns I think and with the ENV/IMDS resolvers only, sources are unnecessary and I'm happy to drop this PR for a simple sequence of resolvers.

arandito:
Closing in favor of https://github.com/smithy-lang/smithy-python/pull/749.

PR #454, opened 2025-03-19
Close request when initial response fails and propagate error

alextwoods:
*Description of changes:*
This fixes an issue where event stream initial response failures cause the stream to hang.  There are two issues:
1.  The response body will not be closed until the request body gets closed (because the h2 streams will be left open).  To signal the server to close the connection, we need to also close the request body to end it.  This is accomplished by checking for error status codes and setting the response future to an exception.  If the response is not closed, the deserialize_error methods will hang forever trying to read the body.
2. We were incorrectly not calling the `done` method on the asyncio response_future and so were not setting the exception in it correclty.

There may be better ways to ensure the request body gets closed on an error in initial request.

### DIFF
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -174,15 +174,13 @@ class CRTResponseFactory:
                     kind=FieldPosition.HEADER,
                 )
 
-        response = AWSCRTHTTPResponse(
-            status=status_code,
-            fields=fields,
-            body=self._body,
-        )
-        if status_code != 200 and status_code >= 300:
-            self._response_future.set_exception(CRTErrorResponse(response))
-        else:
-            self._response_future.set_result(response)
+            self._response_future.set_result(
+                AWSCRTHTTPResponse(
+                    status=status_code,
+                    fields=fields,
+                    body=self._body,
+                )
+            )
 
     async def await_response(self) -> AWSCRTHTTPResponse:
         return await asyncio.wrap_future(self._response_future)
@@ -193,18 +191,6 @@ class CRTResponseFactory:
     def _cancel(self, completion_future: ConcurrentFuture[int | Exception]) -> None:
         if not self._response_future.done():
             self._response_future.cancel()
-
-
-class CRTErrorResponse(Exception):
-    def __init__(self, response: AWSCRTHTTPResponse) -> None:
-        self._status = response.status
-        self._response = response
-
-        super().__init__(f"Request failed with status code {self._status}")
-
-    @property
-    def response(self) -> AWSCRTHTTPResponse:
-        return self._response
 
 
 ConnectionPoolKey = tuple[str, str, int | None]
@@ -265,11 +251,12 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
         crt_stream.completion_future.add_done_callback(
             partial(self._close_input_body, body=crt_body)
         )
-        try:
-            return await response_factory.await_response()
-        except CRTErrorResponse as e:
+
+        response = await response_factory.await_response()
+        if response.status != 200 and response.status >= 300:
             await close(crt_body)
-            return e.response
+
+        return response
 
     def _close_input_body(
         self, future: ConcurrentFuture[int], *, body: "BufferableByteStream | BytesIO"
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -174,13 +174,13 @@ class CRTResponseFactory:
                     kind=FieldPosition.HEADER,
                 )
 
-            self._response_future.set_result(
-                AWSCRTHTTPResponse(
-                    status=status_code,
-                    fields=fields,
-                    body=self._body,
-                )
+        self._response_future.set_result(
+            AWSCRTHTTPResponse(
+                status=status_code,
+                fields=fields,
+                body=self._body,
             )
+        )
 
     async def await_response(self) -> AWSCRTHTTPResponse:
         return await asyncio.wrap_future(self._response_future)
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -174,13 +174,15 @@ class CRTResponseFactory:
                     kind=FieldPosition.HEADER,
                 )
 
-        self._response_future.set_result(
-            AWSCRTHTTPResponse(
-                status=status_code,
-                fields=fields,
-                body=self._body,
-            )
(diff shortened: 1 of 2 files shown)


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -174,13 +174,15 @@ def on_response(
...
-                status=status_code,
-                fields=fields,
-                body=self._body,
-            )
+        response = AWSCRTHTTPResponse(
+            status=status_code,
+            fields=fields,
+            body=self._body,
         )
+        if status_code != 200 and status_code >= 300:
```
alextwoods, line 182: Technically the check for status_code != 200 is redundant with >= 300, but this is the logic the deserialize_error methods use.

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -193,6 +195,18 @@ def _cancel(self, completion_future: ConcurrentFuture[int | Exception]) -> None:
...
+class CRTErrorResponse(Exception):
+    def __init__(self, response: AWSCRTHTTPResponse) -> None:
+        self._status = response.status
+        self._response = response
+
+        super().__init__(f"Request failed with status code {self._status}")
+
+    @property
+    def response(self) -> AWSCRTHTTPResponse:
+        return self._response
```
nateprewitt, line 207: What actually gets surfaced to the end user with the stack trace here? Is it only the message being passed to the base Exception constructor? Ideally we have more context if we can.

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -193,6 +195,18 @@ def _cancel(self, completion_future: ConcurrentFuture[int | Exception]) -> None:
             self._response_future.cancel()
 
 
+class CRTErrorResponse(Exception):
```
nateprewitt, line 198: We want all exceptions in the library to inherit from a top level Smithy/AWS exception to allow customers to be able to grab everything.

```suggestion
class CRTErrorResponse(SmithyHTTPException):
```

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -193,6 +195,18 @@ def _cancel(self, completion_future: ConcurrentFuture[int | Exception]) -> None:
...
+    def __init__(self, response: AWSCRTHTTPResponse) -> None:
+        self._status = response.status
+        self._response = response
+
+        super().__init__(f"Request failed with status code {self._status}")
```
JordonPhillips, line 203: Super should be first

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -251,7 +265,11 @@ async def send(
         crt_stream.completion_future.add_done_callback(
             partial(self._close_input_body, body=crt_body)
         )
-        return await response_factory.await_response()
+        try:
+            return await response_factory.await_response()
+        except CRTErrorResponse as e:
+            await close(crt_body)
+            return e.response
```
JordonPhillips, line 272: I'm not a fan of using exceptions for control flow like this. (I know python does it in the standard library but I don't like that either.) Why not just do this:

```suggestion
        response = await response_factory.await_response()
        if response.status != 200 and response.status >= 300:
            await close(crt_body)
        return response
```

JordonPhillips (changes requested):

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: This exception gets caught in the `AWSCRTHTTPClient.send` method, which then closes the request body and returns the response upwards.  The client code then deserializes the response using its deserializers and will create the appropriate service exception.  An example for an expired security token:

```
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/client.py", line 595, in _execute_operation
    return await self._handle_execution(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/client.py", line 751, in _handle_execution
    return await self._finalize_execution(interceptor_chain, output_context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/client.py", line 1003, in _finalize_execution
    raise context.response
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/client.py", line 717, in _handle_execution
    raise output_context.response
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/client.py", line 916, in _handle_attempt
    output = await deserialize(response_context.transport_response, config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexwoo/python/test-lexruntime/lexruntime/deserialize.py", line 403, in _deserialize_start_conversation
    raise e
lexruntime.models.UnknownApiError: ExpiredTokenException: The security token included in the request is expired

```

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: This should never be raised to customers (its internal to the CRT client) - should we still use the Smithy exception?

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: Yeah - good thought - I like checking the response status here better.

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: Edit: With Jordon's suggestion below this exception is no longer used at all.

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -174,13 +174,13 @@ def on_response(
...
-            AWSCRTHTTPResponse(
-                status=status_code,
-                fields=fields,
-                body=self._body,
+            self._response_future.set_result(
+                AWSCRTHTTPResponse(
+                    status=status_code,
+                    fields=fields,
+                    body=self._body,
+                )
```
JordonPhillips, line 182: This seems to still be indented so that it's now in the except block

JordonPhillips (changes requested):

packages/smithy-http/src/smithy_http/aio/crt.py
alextwoods: Good catch - was testing this on top of the event signing PR changes and messed up moving the changes back.

JordonPhillips (approved):

PR #455, opened 2025-03-19
Decouple endpoint resolution from http

JordonPhillips:
This decouples endpoint resolution from http. It does this by using the new `Endpoint` interface that has a typed properties bag. HTTP endpoint providers MAY add a "headers" property that will be merged into  the request's headers.

The input of endpoint resolvers was simplified to be a standard param bundle that should contain anything they could possibly want. This includes the context properties for the invocation, so more stuff can be added there by interceptors. The config object is added as standard to the properties bag, but it's loosely typed since it's a generated type. The endpoint uri is added in a strongly typed way, though I had to play a bit fast and loose with pyright to get it to work with unions.

I also updated the formatting to remove unused imports because that makes the process of changing the default endpoint provider less painful.


codegen/core/src/main/java/software/amazon/smithy/python/codegen/PythonFormatter.java
```diff
@@ -72,6 +72,7 @@ private void format(FileManifest fileManifest) {
             return;
         }
         LOGGER.info("Running code formatter on generated code");
+        CodegenUtils.runCommand("python3 -m ruff check --fix", fileManifest.getBaseDir());
```
haydenbaker, line 75: Hmmm... Instead of having ruff fix it, can we instead fix the codegen itself? In other words, have ruff fail the build if, say, unused imports.

JordonPhillips: The problem is that there's not a good way to have section-conditional imports. That would be a big change, and it'd have to go in smithy itself

haydenbaker (dismissed):

packages/smithy-core/src/smithy_core/endpoints.py
```diff
@@ -0,0 +1,89 @@
...
+    properties: _TypedProperties = field(default_factory=TypedProperties)
+    """Properties required to interact with the endpoint.
+
+    For example, in some AWS use cases this might contain HTTP headers to add to each
+    request.
+    """
+
+
+@dataclass(kw_only=True)
+class EndpointResolverParams[I: SerializeableShape]:
```
alextwoods, line 44: I do like this as a simplification for all endpoint parameters and to some extent it allows more customer extension to clients (since they can interact with any part of the input or context). However, it doesn't quite match to EP2's guidance on generating specific parameters per service.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -488,37 +490,44 @@ await sleep(retry_token.retry_delay)
         writer.popState();
 
         writer.pushState(new ResolveEndpointSection());
+        writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
+        writer.addDependency(SmithyPythonDependency.SMITHY_HTTP);
+        writer.addImport("smithy_core", "URI");
+        writer.addImport("smithy_core.endpoints", "EndpointResolverParams");
+        writer.write("""
+                        # Step 7f: Invoke endpoint_resolver.resolve_endpoint
+                        context.properties["endpoint_uri"] = config.endpoint_uri
```
alextwoods, line 499: How are you binding region to the context properties?  I think this won't work without that.  We could of course add all of the config properties at this point, but ideally we should limit endpoint parameters to whats required.

alextwoods (changes requested):

codegen/aws/core/src/it/java/software/amazon/smithy/python/codegen/test/AwsCodegenTest.java
```diff
@@ -0,0 +1,41 @@
...
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.python.codegen.PythonClientCodegenPlugin;
+
+/**
+ * Simple test that executes the Python client codegen plugin for an AWS-like service.
+ */
+public class AwsCodegenTest {
+
+    @Test
+    public void testCodegen(@TempDir Path tempDir) {
```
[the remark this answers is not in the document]
JordonPhillips: Asserting output on a code generator is just writing the generator twice. We have protocol tests that actually test that the thing works.

packages/smithy-core/src/smithy_core/endpoints.py
JordonPhillips: > it doesn't quite match to EP2's guidance on generating specific parameters per service

I very explicitly do not want to do that. It dramatically increases complexity by foisting the logic of extracting what's important into the request pipeline, which ideally knows nothing about the resolver other than that it exists. There's no way to even do it without codegen that isn't just exactly what the resolver would do when given the world, except there's extra steps because you instead have to ask for the resolver params class, give that the world, and then give that to the resolver.

All for what? A moderately easier to test interface? No thanks.

codegen/aws/core/src/it/java/software/amazon/smithy/python/codegen/test/AwsCodegenTest.java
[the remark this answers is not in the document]
alextwoods: Theres a lot of generation behavior that isn't covered by protocol tests (eg, endpoints and auth) that could be covered by integration testing potentially.  Entire out of scope here though of course, just was thinking about it.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
JordonPhillips: > How are you binding region to the context properties?

It'll need to go into an interceptor. Of course, I'm also just dropping the config in there at the start so maybe it just makes sense to leave it there. Makes typing a bit worse though.

> ideally we should limit endpoint parameters to whats required

This is an ideal not worth pursuing. See my other comment.

alextwoods: I think this is currently broken - it doesn't look like region/arbitrary config is being added to the context properties.

I don't disagree with the other comment - and I'm fine with saying everything needs to be here, but then we do need to provide the entire config through context.

codegen/aws/core/src/it/java/software/amazon/smithy/python/codegen/test/AwsCodegenTest.java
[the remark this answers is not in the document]
haydenbaker: As we move code from being generated to being in the runtime/core libraries, it becomes much more possible to have high coverage/confidence. Integ tests for the generated code (beyond just testing if it generates or not) are better suited outside of java-land though.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
JordonPhillips: Updated to pull from config

packages/smithy-core/src/smithy_core/endpoints.py
JordonPhillips: The problem is that we've created a world where we need a broad interface. We've made the bed, we might as well admit that we're lying in it.

alextwoods: Yeah thats a fair point - should we be taking the same approach for Auth resolver/parameters?

alextwoods (approved):

PR #460, opened 2025-03-20
Copy request for each attempt / don't retry client errors

alextwoods:
*Description of changes:*
In #449 the `copy` on request context when calling `_handle_attempt` was removed and the `copy` method on interceptor context was removed as well.  This breaks endpoint construction during _handle_attempt - without a copy of the request, the destination is modified for each attempt.

Additionally, this PR removes retries for client errors from the SimpleRetryStrategy. 

### DIFF
--- a/packages/smithy-core/src/smithy_core/retries.py
+++ b/packages/smithy-core/src/smithy_core/retries.py
@@ -5,6 +5,8 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from enum import Enum
 
+from smithy_core.interfaces.retries import RetryErrorType
+
 from .exceptions import SmithyRetryException
 from .interfaces import retries as retries_interface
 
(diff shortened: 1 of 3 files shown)


nateprewitt (dismissed):

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -232,13 +234,20 @@ def refresh_retry_token_for_retry(
...
-            raise SmithyRetryException(
-                f"Reached maximum number of allowed attempts: {self.max_attempts}"
-            )
-        retry_delay = self.backoff_strategy.compute_next_backoff_delay(retry_count)
-        return SimpleRetryToken(retry_count=retry_count, retry_delay=retry_delay)
+        if error_info.error_type in [
+            RetryErrorType.TRANSIENT,
+            RetryErrorType.THROTTLING,
+            RetryErrorType.SERVER_ERROR,
+        ]:
```
JordonPhillips, line 241:
```suggestion
        if error_info.error_type is not RetryErrorType.CLIENT_ERROR:
```

JordonPhillips (changes requested):

nateprewitt (approved):

JordonPhillips (approved):

PR #462, opened 2025-03-21
Use base ruff config for all projects

JordonPhillips:
I noticed we weren't sorting imports despite having that configured in our `pyproject.toml`. Turns out ruff wasn't using the top-level config at all. So I fixed it. There were a few minor manual fixes that had to be made too, and a few noqa's had to be added.

### DIFF
--- a/packages/aws-event-stream/src/aws_event_stream/aio/__init__.py
+++ b/packages/aws-event-stream/src/aws_event_stream/aio/__init__.py
@@ -3,21 +3,19 @@
 import asyncio
 import logging
 from collections.abc import Callable
+from typing import Protocol
 
-from smithy_core.aio.interfaces import AsyncByteStream
+from smithy_core.aio.interfaces import AsyncByteStream, AsyncWriter
+from smithy_core.aio.interfaces.eventstream import EventPublisher, EventReceiver
 from smithy_core.codecs import Codec
 from smithy_core.deserializers import DeserializeableShape, ShapeDeserializer
 from smithy_core.exceptions import ExpectationNotMetException
 from smithy_core.serializers import SerializeableShape
-from smithy_core.aio.interfaces.eventstream import EventPublisher, EventReceiver
-from smithy_core.aio.interfaces import AsyncWriter
 
-from .._private.serializers import EventSerializer as _EventSerializer
 from .._private.deserializers import EventDeserializer as _EventDeserializer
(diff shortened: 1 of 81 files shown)


packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
```diff
@@ -20,14 +21,14 @@
 __version__ = importlib.metadata.version("aws-sdk-signers")
 
 __all__ = (
-    "AsyncBytesReader",
-    "AsyncSigV4Signer",
-    "AsyncEventSigner",
+    "URI",
```
nateprewitt, line 24: ??? What is this sort order? Are we using the ascii symbol table to determine ordering?

nateprewitt (dismissed):

packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
jonathan343: > An isort-style sort orders items first according to their casing: SCREAMING_SNAKE_CASE names (conventionally used for global constants) come first, followed by CamelCase names (conventionally used for classes), followed by anything else. Within each category, a [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order) is used to order the elements.

Source: https://docs.astral.sh/ruff/rules/unsorted-dunder-all/

Seems like `URI` got pushed to the top since it's being treated like a constant.

packages/aws-event-stream/src/aws_event_stream/aio/__init__.py
```diff
@@ -50,7 +48,7 @@ def __init__(
 
     async def send(self, event: E) -> None:
         if self._closed:
-            raise IOError("Attempted to write to closed stream.")
```
jonathan343, line 53: For reviewers also wondering, this is related to: https://docs.astral.sh/ruff/rules/os-error-alias/

jonathan343 (dismissed):
LGTM! Thanks for adding this! I knew something felt off with our formatting and linting.

packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
JordonPhillips: This can be fixed

jonathan343 (approved):
🚀

PR #465, opened 2025-03-24
Fix minor generator issues

nateprewitt:
Cleaning up a couple minor generator items that I noticed getting formatted out of the code every time.

### DIFF
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
@@ -28,7 +28,7 @@ public class AwsUserAgentIntegration implements PythonIntegration {
                         ua_suffix=config.user_agent_extra,
                         ua_app_id=config.sdk_ua_app_id,
                         sdk_version=$3T,
-                        service_id='$4L'
+                        service_id=$4S,
                     )
                 )
             """;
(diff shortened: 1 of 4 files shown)


codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
```diff
@@ -28,7 +28,7 @@ def aws_user_agent_plugin(config: $1T):
                         ua_suffix=config.user_agent_extra,
                         ua_app_id=config.sdk_ua_app_id,
                         sdk_version=$3T,
-                        service_id='$4L'
+                        service_id=\"$4L\",
```
jonathan343, line 31: Just curious, why do these need to be escaped? Can't we use just `"` since this is part of a text block/multi-line string?

nateprewitt: Yep, good call. These don't need the escapes, this was just find-and-replace from the other changes.

JordonPhillips, line 31:
```suggestion
                        service_id=$4S,
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/InitGenerator.java
```diff
@@ -43,6 +43,6 @@ public void run() {
         context.writerDelegator()
                 .useFileWriter(
                         "%s/__init__.py".formatted(context.settings().moduleName()),
-                        w -> w.write("__version__: str = '$L'", context.settings().moduleVersion()));
+                        w -> w.write("__version__: str = \"$L\"", context.settings().moduleVersion()));
```
JordonPhillips, line 46:
```suggestion
                        w -> w.write("__version__: str = $S", context.settings().moduleVersion()));
```

JordonPhillips (dismissed):

jonathan343 (approved):

PR #466, opened 2025-03-25
Add error serde to http-binding client protocols

haydenbaker:
*Description of changes:*
- Adds error serde and handling in http binding client protocols
  - Adds amzn header extractor and adds it to the rest json client protocol
- Adds new exceptions, CallException and ModeledException, which are meant to be used when code-generating the error shapes
 
### DIFF
--- a/packages/smithy-core/src/smithy_core/errors.py
+++ b/packages/smithy-core/src/smithy_core/errors.py
@@ -1,0 +1,32 @@
+from dataclasses import dataclass
+from typing import Literal
+
+from smithy_core.deserializers import DeserializeableShape
+
+type Fault = Literal["client", "server", "other"]
+
+
+@dataclass(kw_only=True, frozen=True)
+class CallException(RuntimeError):
+    """The top-level exception that should be used to throw application-level errors
+    from clients and servers.
+
+    This should be used in protocol error deserialization, throwing errors based on
+    protocol-hints, network errors, and shape validation errors. It should not be used
+    for illegal arguments, null argument validation, or other kinds of logic errors
+    sufficiently covered by the Java standard library.
+    """
+
+    fault: Fault = "other"
+    """The party that is at fault for the error, if any."""
+
+    message: str = ""
+    """The error message."""
+
+    # TODO: retry-ability and associated information (throttling, duration, etc.), perhaps 'Retryability' dataclass?
+
+
+@dataclass(kw_only=True, frozen=True)
+class ModeledException(CallException, DeserializeableShape):
+    """The top-level exception that should be used to throw modeled errors from clients
+    and servers."""
--- a/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
+++ b/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
@@ -4,6 +4,7 @@ from typing import Protocol
 
 from smithy_core.aio.interfaces import ClientTransport, Request, Response
 from smithy_core.aio.utils import read_streaming_blob, read_streaming_blob_async
+from smithy_core.shapes import ShapeID
 
 from ...interfaces import (
     Fields,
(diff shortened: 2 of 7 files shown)


packages/smithy-core/src/smithy_core/errors.py
```diff
@@ -0,0 +1,32 @@
...
+    sufficiently covered by the Java standard library.
+    """
+
+    fault: Fault = "other"
+    """The party that is at fault for the error, if any."""
+
+    message: str = ""
+    """The error message."""
+
+    # TODO: retry-ability and associated information (throttling, duration, etc.), perhaps 'Retryability' dataclass?
```
JordonPhillips, line 26: Yeah having a `RetryableException` protocol, mixin, or something would be ideal. Classifying is too complex without embedding the information in the exception.

packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
```diff
@@ -83,3 +84,18 @@ async def send(
...
         """
         ...
+
+
+class ErrorExtractor(Protocol):
+    """Extract error shape IDs from an HTTP response."""
+
+    def get_error(
+        self,
+        response: HTTPResponse,
```
JordonPhillips, line 94: This shouldn't be tied to HTTP

packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
```diff
@@ -83,3 +84,18 @@ async def send(
...
+    ) -> ShapeID | None:
```
JordonPhillips, line 95: This might actually be the place to get the retry info

packages/smithy-http/src/smithy_http/deserializers.py
```diff
@@ -257,3 +263,72 @@ def _consume_payload(self) -> bytes:
...
+        self,
+        operation: APIOperation[Any, Any],
+        error_registry: TypeRegistry,
+        context: TypedProperties,
+    ) -> CallException:
+        body = self._body
+        if isinstance(body, bytearray):
+            body = bytes(body)
+        deserializer = self._payload_codec.create_deserializer(body)
+        document = deserializer.read_document(DOCUMENT)
```
JordonPhillips, line 302: Can errors have streaming payloads? (blob streams that is)

packages/smithy-http/src/smithy_http/deserializers.py
```diff
@@ -257,3 +263,72 @@ def _consume_payload(self) -> bytes:
...
+        if error_id is not None:
+            error_shape = error_registry.get(error_id)
+            # make sure the error shape is derived from modeled exception
+            if not isinstance(error_shape, ModeledException):
+                raise ExpectationNotMetException(
+                    f"Modeled errors must be derived from 'ModeledException', but got {error_shape}"
+                )
+
+            # return the deserialized error
+            return error_shape.deserialize(deserializer)
```
JordonPhillips, line 320: You need to rewind the body since you've already read from it

packages/smithy-http/src/smithy_http/deserializers.py
```diff
@@ -257,3 +263,72 @@ def _consume_payload(self) -> bytes:
...
+            body = bytes(body)
+        deserializer = self._payload_codec.create_deserializer(body)
+        document = deserializer.read_document(DOCUMENT)
+
+        # try to get the error shape-id from the extractor
+        error_id = self._extractor.get_error(self._response)
+
+        # if none, get it from the parsed document (e.g. '__type')
+        if error_id is None:
+            error_id = document.discriminator
```
JordonPhillips, line 309: You're getting this document from the payload codec, but that's not necessarily gonna give you what you want. Different JSON protocols could embed this differently.

JordonPhillips (changes requested):

PR #468, opened 2025-03-26
Decouple auth from http

JordonPhillips:
This decouples auth from HTTP, allowing it to be defined centrally.

A number of changes have been made to various interfaces. Notably identity resolvers are generally expected to have zero-arg constructors, instead getting everything they need from their properties where possible.

Construction of event signers has been moved into AuthScheme.


nateprewitt (commented):
Need to do one more pass tomorrow, but looks good. Just one question on why we're making some of the types private.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAuthIntegration.java
```diff
@@ -69,30 +71,26 @@ public void customize(GenerationContext context) {
...
             writer.write("""
-                    def $1L(auth_params: $2T) -> HTTPAuthOption | None:
-                        return HTTPAuthOption(
-                            scheme_id=$3S,
-                            identity_properties={},
-                            signer_properties={
-                                "service": $4S,
-                                "region": auth_params.region
-                            }
+                    def $1L(auth_params: AuthParams[Any, Any]) -> _AuthOption | None:
```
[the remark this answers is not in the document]
JordonPhillips: It's just so the protocol doesn't conflict with the concrete implementation since they have the same name.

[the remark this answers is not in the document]
nateprewitt: I wonder if it would be better to do something like `AuthOptionType` or some other indication it's an interface if we need to avoid name conflicts. The leading underscore for name disambiguation seems like it might be signaling the wrong thing to the end-user.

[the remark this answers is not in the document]
JordonPhillips: I went with `AuthOptionProtocol`

packages/smithy-core/src/smithy_core/aio/interfaces/auth.py
```diff
@@ -0,0 +1,89 @@
...
+
+        :param request: The request to be signed.
+        :param identity: The identity to use to sign the request.
+        :param properties: Additional properties used to sign the request.
+        """
+        ...
+
+
+class EventSigner[I, SP: Mapping[str, Any]](Protocol):
+    """A class that signs requests before they are sent."""
```
nateprewitt, line 27:
```suggestion
    """A class that signs events before they are sent."""
```

packages/smithy-core/src/smithy_core/auth.py
```diff
@@ -0,0 +1,66 @@
...
+
+@dataclass(kw_only=True)
+class AuthOption:
+    """Auth scheme used for signing and identity resolution."""
+
+    scheme_id: ShapeID
+    """The ID of the auth scheme to use."""
+
+    identity_properties: _TypedProperties = field(default_factory=TypedProperties)
+    """Paramters to pass to the identity resolver method."""
```
nateprewitt, line 37:
```suggestion
    """Parameters to pass to the identity resolver method."""
```

packages/smithy-core/src/smithy_core/auth.py
```diff
@@ -0,0 +1,66 @@
...
+
+    signer_properties: _TypedProperties = field(default_factory=TypedProperties)
+    """Paramters to pass to the signing method."""
```
nateprewitt, line 40:
```suggestion
    """Parameters to pass to the signing method."""
```

packages/smithy-core/src/smithy_core/interfaces/auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class AuthOption(Protocol):
+    """Auth scheme used for signing and identity resolution."""
+
+    scheme_id: ShapeID
+    """The ID of the auth scheme to use."""
+
+    identity_properties: TypedProperties
+    """Paramters to pass to the identity resolver method."""
```
nateprewitt, line 20:
```suggestion
    """Parameters to pass to the identity resolver method."""
```

packages/smithy-core/src/smithy_core/interfaces/auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+    signer_properties: TypedProperties
+    """Paramters to pass to the signing method."""
```
nateprewitt, line 23:
```suggestion
    """Parameters to pass to the signing method."""
```

packages/smithy-core/src/smithy_core/interfaces/identity.py
```diff
@@ -1,35 +1,22 @@
...
 @runtime_checkable
 class Identity(Protocol):
     """An entity available to the client representing who the user is."""
 
-    # The expiration time of the identity. If time zone is provided,
-    # it is updated to UTC. The value must always be in UTC.
     expiration: datetime | None = None
+    """The expiration time of the identity.
+
+    If time zone is provided, it is updated to UTC. The value must always be in UTC.
```
nateprewitt, line 14: We say this in a bunch of places but I'm not sure I've seen an implementation where we actually do this. Is this a note to implementers, or should we have some base implementation that checks timezones and makes adjustments?

nateprewitt (changes requested):

designs/auth.md
```diff
@@ -0,0 +1,178 @@
...
+
+```python
+class Signer[R: Request, I, SP: Mapping[str, Any]](Protocol):
+    async def sign(self, *, request: R, identity: I, properties: SP) -> R:
+        ...
+```
+
+Signers are constructed by the `AuthScheme`'s `signer` method.
+
+Signers MAY modify the given request and return it, or construct a new signed
```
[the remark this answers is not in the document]
JordonPhillips: We already will create a copy of the request before each attempt, so the potential ramifications are limited.

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -804,26 +803,25 @@ class AsyncEventSigner:
...
 
-    async def sign_event(
+    async def sign(
         self,
         *,
-        event_message: "EventMessage",
-        event_encoder_cls: type["EventHeaderEncoder"],
+        event: "EventMessage",
+        identity: _AWSCredentialsIdentity,
+        properties: SigV4SigningProperties,
```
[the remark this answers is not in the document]
JordonPhillips: >  When would we be changing identity/properties between signatures

When temporary credentials are being used. Credentials from AssumeRole last a max of 12 hours, and I've heard from service teams that some streams can last 20 hours or more.

> I think this needs a larger refactoring to account for previous signature

This is being tracked already as `_prior_signature`

> if sign is sequential and stateful

This is true for sigv4 today, but may not be true for other auth schemes. So it's the burden of the sigv4 signer to enforce that signing is sequential and it does that with locks.

packages/smithy-core/src/smithy_core/interfaces/identity.py
JordonPhillips: The identity resolver is expected to do that. I could update this to a dataclass and drop it in post_init?

nateprewitt (approved):

PR #470, opened 2025-03-27
Move aws_event_stream to smithy_aws_event_stream

nateprewitt:
*Description of changes:*
This PR adds an initial changelog and moves `aws-event-streams` to `smithy-aws-event-streams` under the Smithy domain to make it's relationship clearer.


packages/smithy-aws-event-streams/CHANGES.md
```diff
@@ -0,0 +1,4 @@
+## Unreleased
```
alextwoods, line 1: I assume we should add similar changes.md to other packages and move these to 0.0.1 once released?

alextwoods (dismissed):

packages/smithy-aws-event-streams/CHANGES.md
nateprewitt: Yep, was working on that and then started the rename. I'll have 2 more PRs with changes and then finalizing release commits.

packages/smithy-aws-event-streams/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
-name = "aws-event-stream"
+name = "smithy-aws-event-stream"
```
jonathan343, line 2: I see some other smithy packages we have use snake case. Does this get treaded the same as `smithy_aws_event_stream`? We should make them consistent.

jonathan343:
Why is the patent directory plural (`smithy-aws-event-streams`) and the child directory singular (`smithy_aws_event_stream`)? I would expect for them to match

packages/smithy-aws-event-streams/pyproject.toml
nateprewitt: It looks like none of the projects are consistent. I'll do a follow up PR to normalize everything.

jonathan343 (approved):
🚢

PR #472, opened 2025-03-31
Normalize package names to match project hyphenation

nateprewitt:
*Issue #, if available:*

*Description of changes:*
This PR addresses a callout from #470 that our project names in some of the packages is not consistent. This normalizes the project name to being hyphenated while the package directory is snake case as is the current convention.


jonathan343 (approved):
:shipit:

PR #478, opened 2025-04-03
Suffix clients with Client after service name

nateprewitt:
*Description of changes:*
This is a proposal for adding `Client` to the end of our generated service clients. While these all originated from the generated namespace `from service_name.client import ServiceName`, it reads a bit oddly in isolated code snippets.

Prior art from recent SDKs is showing 3 out of the 5 existing Smithy SDKs all suffix with Client to reduce ambiguity. I don't think there's a lot of standing precedent in Python itself for which direction to go. I'm adding this PR as a talking point for if we feel this is necessary.

I don't think it hurts to add this for the moment, but we'll need to decide if we bias one way now, are we're willing to break that based on feedback later?


SamRemis (approved):
Since this PR is opened as a point of discussion, just adding my two cents:

Approving this since I'm very much in favor of it.  The docs seem to generate fine with this change, but I should probably look more into it than just a quick skim.  

Calling this a "client" will make this more familiar for users migrating from botocore and makes it clear that this is not just a general concept, but an object to be used to call the service.  I'd be very surprised if we had strong customer feedback making a compelling argument for a different name.  If we do, we should make the breaking change before this goes GA.

PR #489, opened 2025-04-17
Error rework

JordonPhillips:
This updates exceptions to embed necessary retry info. This will allow any layer that has access to the exception to set or utilize it without having to have additional interfaces and hooks.

I included a design doc to explain things in a durable way

### DIFF
--- a/designs/exceptions.md
+++ b/designs/exceptions.md
@@ -57,7 +57,7 @@ class ErrorRetryInfo(Protocol):
     Retry strategies MAY choose to wait longer.
     """
 
-    is_throttle: bool = False
+    is_throttling_error: bool = False
     """Whether the error is a throttling error."""
 ```
 
@@ -120,5 +120,5 @@ class ServiceException(ModeledException):
 class ThrottlingException(ServcieException):
     fault: Fault = "client"
     is_retry_safe: bool | None = True
-    is_throttle: bool = True
+    is_throttling_error: bool = True
 ```
--- a/designs/exceptions.md
+++ b/designs/exceptions.md
@@ -1,0 +1,124 @@
+# Exceptions
+
+Exceptions are a necessary aspect of any software product (Go notwithstanding),
+and care must be taken in how they're exposed. This document describes how
+smithy-python clients will expose exceptions to customers.
+
+## Goals
+
+* Every exception raised by a Smithy client should be catchable with a single,
+  specific catch statement (that is, not just `except Exception`).
+* Every modeled exception raised by a service should be catchable with a single,
+  specific catch statement.
+* Exceptions should contain information about retryablility where relevant.
+
+## Specification
+
+Every exception raised by a Smithy client MUST inherit from `SmithyException`.
+
+```python
+class SmithyException(Exception):
+    """Base exception type for all exceptions raised by smithy-python."""
+```
+
+If an exception that is not a `SmithyException` is thrown while executing a
+request, that exception MUST be wrapped in a `SmithyException` and the
+`__cause__` MUST be set to the original exception.
+
+Just as in normal Python programming, different exception types SHOULD be made
+for different kinds of exceptions. `SerializationException`, for example, will
+serve as the exception type for any exceptions that occur while serializing a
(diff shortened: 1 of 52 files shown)


packages/smithy-core/src/smithy_core/interfaces/exceptions.py
```diff
@@ -10,4 +12,4 @@ class HasFault(Protocol):
     All modeled errors will have a fault that is either "client" or "server".
     """
 
-    fault: ClassVar[Literal["client", "server"]]
+    fault: Fault | None
```
jonathan343, line 15: is the `| None` needed here if `Fault` is already defined as `Literal["client", "server"] | None`?

jonathan343 (commented):
Dropping some initial comments after reviewing the design doc and python code.

designs/exceptions.md
```diff
@@ -0,0 +1,124 @@
...
+* Every modeled exception raised by a service should be catchable with a single,
+  specific catch statement.
+* Exceptions should contain information about retryablility where relevant.
+
+## Specification
+
+Every exception raised by a Smithy client MUST inherit from `SmithyException`.
+
+```python
+class SmithyException(Exception):
```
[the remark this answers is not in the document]
JordonPhillips: I think that just got missed back in the day. I can change them all now.

designs/retries.md
```diff
@@ -0,0 +1,170 @@
...
+        response = e
+
+    if isinstance(response, Exception):
+        try:
+            retry_token = retry_strategy.refresh_retry_token_for_retry(
+                token_to_renew=retry_token,
+                error=e
+            )
+            continue
+        except RetryError retry_error:
```
jonathan343, line 91:
```suggestion
        except RetryError as retry_error:
```

designs/retries.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+A request using a `RetryStrategy` would look something like the following
+example:
+
+```python
+try:
+    retry_token = retry_strategy.acquire_initial_retry_token()
+except RetryError:
+    transpoort_response = transport_client.send(serialized_request)
```
jonathan343, line 73:
```suggestion
    transport_response = transport_client.send(serialized_request)
```

designs/retries.md
```diff
@@ -0,0 +1,170 @@
...
+
+        :param token: The token used for the previous successful attempt.
+        """
+        ...
+```
+
+A request using a `RetryStrategy` would look something like the following
+example:
+
+```python
```
jonathan343, line 69: Can you break down this example to help me understand it better? I find it really confusing

jonathan343: Update: The content below helped by understand this example better:
> RetryStrategy implementations MUST raise a RetryError if they receive an exception where is_retry_safe is False and SHOULD raise a RetryError if it is None. RetryStrategy implementations SHOULD use a delay that is at least as long as retry_after but MAY choose to wait longer.

packages/smithy-core/src/smithy_core/exceptions.py
```diff
@@ -1,39 +1,96 @@
...
+
+    def __post_init__(self):
+        super().__init__(self.message)
+
+
+@dataclass(kw_only=True)
+class ModeledError(CallError):
+    """Base exception to be used for modeled errors."""
+
+    fault: Fault = "client"
```
jonathan343, line 61: Why default to `"client"` instead of None?

jonathan343 (commented):
Looks good! I have a few more questions and suggestions

packages/smithy-core/src/smithy_core/exceptions.py
JordonPhillips: Modeled errors will always have this defined as the Smithy trait requires it. So this isn't really important either way

designs/retries.md
JordonPhillips: I moved the example to the end to avoid this confusion

jonathan343 (approved):
🚢

PR #492, opened 2025-05-02
Disambiguate generics with unknown parametric assignments

JordonPhillips:
Pyright's strict mode is becoming increasingly strict with regards to generics. You used to be able to effectively narrow them by declaring the type like so:

```python
string_list: list[str] = list()
```

But that's no longer accepted. Now the actual value must itself resolve to something that has resolved parameterization. This unfortunately means that actual casting is sometimes necessary. It's not a huge hit because it's a do-nothing function, but it's still a function call that I'd rather not be doing.


jonathan343 (approved):
LGTM!

PR #493, opened 2025-05-06
Parse errors in http binding protocols

JordonPhillips:
This updates the HTTP binding protocol to handle errors. This supercedes #466 

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
@@ -1,19 +1,44 @@
-from typing import Final
+from typing import Any, Final
 
 from smithy_core.codecs import Codec
+from smithy_core.schemas import APIOperation
 from smithy_core.shapes import ShapeID
+from smithy_http.aio.interfaces import HTTPErrorIdentifier, HTTPResponse
 from smithy_http.aio.protocols import HttpBindingClientProtocol
 from smithy_json import JSONCodec
 
 from ..traits import RestJson1Trait
 
 
+class AWSErrorIdentifier(HTTPErrorIdentifier):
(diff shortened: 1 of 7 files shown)


packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -1,19 +1,44 @@
...
+    def identify(
+        self,
+        *,
+        operation: APIOperation[Any, Any],
+        response: HTTPResponse,
+    ) -> ShapeID | None:
+        if self._HEADER_KEY not in response.fields:
+            return None
+
+        code = response.fields[self._HEADER_KEY].values[0]
```
nateprewitt, line 25: We may want to rework this. You can instantiate a Field with `values=None` which results in an empty list `[]`. That's going to fail if we try to index, we might want to either check values beforehand or do something like:
```suggestion
        error_field = response.fields[self._HEADER_KEY]
        code = error_field.values[0] if len(error_field.values) > 0 else None
```

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -116,3 +137,69 @@ async def deserialize_response[
...
 
         return operation.output.deserialize(deserializer)
+
+    def _is_success(
+        self,
+        operation: APIOperation[Any, Any],
+        context: TypedProperties,
+        response: HTTPResponse,
+    ) -> bool:
+        return 200 <= response.status <= 299
```
nateprewitt, line 147: nit; this is mostly for consistency with other tooling. 299 isn't something you generally think of with http status codes.
```suggestion
        return 200 <= response.status < 300
```

nateprewitt (commented):
Two minor notes, otherwise looks good.

nateprewitt (approved):

PR #495, opened 2025-05-09
Handle AWS error codes in documents

JordonPhillips:
This ensures that error codes in documents (i.e. the bodies of the http response) can be mapped to shape ids.

### DIFF
--- a/packages/smithy-aws-core/tests/unit/test_utils.py
+++ b/packages/smithy-aws-core/tests/unit/test_utils.py
@@ -1,0 +1,78 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from smithy_aws_core.utils import parse_document_discriminator, parse_error_code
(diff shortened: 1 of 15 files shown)


packages/smithy-aws-core/tests/unit/test_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.parametrize(
+    "document, expected",
+    [
+        ({"__type": "FooError"}, "com.test#FooError"),
+        ({"__type": "com.test#FooError"}, "com.test#FooError"),
+        (
+            {
+                "__type": "FooError:http://internal.amazon.com/coral/com.amazon.coral.validate/"
```
nateprewitt, line 17: Is there a better error shape for this? I'm a little surprised we're using the internal coral reference in our tests. I see that's also done in the protocol tests and listed on the smithy website.

nateprewitt: Talked to Kevin, I guess this is just a public detail with Smithy.

nateprewitt (approved):

PR #502, opened 2025-06-03
Fix query binding

JordonPhillips:
This fixes an issue where query values were being double-encoded, and also serialization of maps of lists in the query string

### DIFF
--- a/packages/smithy-http/src/smithy_http/serializers.py
+++ b/packages/smithy-http/src/smithy_http/serializers.py
@@ -570,12 +570,27 @@ class HTTPQueryMapSerializer(MapSerializer):
         :param query_params: The list of query param tuples to append to.
         """
         self._query_params = query_params
-        self._delegate = CapturingSerializer()
 
     def entry(self, key: str, value_writer: Callable[[ShapeSerializer], None]):
-        value_writer(self._delegate)
-        assert self._delegate.result is not None  # noqa: S101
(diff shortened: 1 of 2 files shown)


packages/smithy-http/src/smithy_http/serializers.py
```diff
@@ -486,7 +486,7 @@ def write_big_decimal(self, schema: Schema, value: Decimal) -> None:
 
     def write_string(self, schema: Schema, value: str) -> None:
         key = self._key or schema.expect_trait(HTTPQueryTrait).key
-        self.query_params.append((key, urlquote(value, safe="")))
```
nateprewitt, line 489: Does this have any implications on our current signing setup? The signers can double encode but I'm wondering if we're setting that appropriately if we're removing this.

packages/smithy-http/src/smithy_http/serializers.py
```diff
@@ -328,7 +328,7 @@ def __init__(
             :py:class:`HTTPHeaderTrait` will be checked instead. Required when
             collecting list entries.
         :param headers: An optional list of header tuples to append to. If not
-            set, one will be created.
+            set, one will be created. Values appended will not be escaped.
```
nateprewitt, line 331: What is handling any required escaping now?

JordonPhillips: They're escaped when they're joined

nateprewitt (approved):

PR #509, opened 2025-06-04
Fix media and content type

JordonPhillips:
This updates media type serialization to pick up all the proper cases, including omitting media type if there's no body. It also introduces content-length serialization.

### DIFF
--- a/packages/smithy-http/src/smithy_http/serializers.py
+++ b/packages/smithy-http/src/smithy_http/serializers.py
@@ -9,6 +9,7 @@ from urllib.parse import quote as urlquote
 
 from smithy_core import URI
 from smithy_core.codecs import Codec
+from smithy_core.exceptions import SerializationError
 from smithy_core.schemas import Schema
 from smithy_core.serializers import (
     InterceptingSerializer,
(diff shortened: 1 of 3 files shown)


packages/smithy-http/src/smithy_http/serializers.py
```diff
@@ -154,10 +183,22 @@ def begin_struct(self, schema: Schema) -> Iterator[ShapeSerializer]:
...
+        if (
+            seek := getattr(payload, "seek", None)
+        ) is not None and not iscoroutinefunction(seek):
+            seek(pos, whence)
+
+    def _compute_content_length(self, payload: Any) -> int | None:
+        content_length = self._seek(payload, 0, 2)
+        if content_length is not None:
+            self._seek(payload, 0, 0)
+            return content_length
```
nateprewitt, line 200: I think this is a vector for data corruption right? If you get a partially read stream, we can't seek back to beginning. We'd want to use `tell` to check our current position before reading to the end or take the length and seek in reverse from the end. I think that's potentially going to have issues with length of bytes vs characters though.

nateprewitt (approved):

PR #522, opened 2025-07-08
Move dev dependencies under "[dependency-groups]" in pyproject.toml

jonathan343:
## Summary

In `aws-sdk-signers` and the code generator that is used to build the AWS clients in [aws-sdk-python](https://github.com/awslabs/aws-sdk-python/tree/develop/clients), we're defining optional dependencies in pyproject.toml for test and docs dependencies as shown below:

```
[project.optional-dependencies]
foo = [ .. ]
bar = [ ... ]
```

This PR moves these under `[dependency-groups]` ([ref](https://packaging.python.org/en/latest/specifications/dependency-groups/)):

> This specification defines Dependency Groups, a mechanism for storing package requirements in pyproject.toml files such that they are not included in project metadata when it is built.
>
> Dependency Groups are suitable for internal development use-cases like linting and testing, as well as for projects which are not built for distribution, like collections of related scripts.
>
> Fundamentally, Dependency Groups should be thought of as being a standardized subset of the capabilities of requirements.txt files (which are pip-specific).
>
---

### DIFF
--- a/packages/aws-sdk-signers/pyproject.toml
+++ b/packages/aws-sdk-signers/pyproject.toml
@@ -34,14 +34,6 @@ exclude = [
  "tests",
 ]
 
-[dependency-groups]
-dev = [
-    "freezegun",
-    "pytest",
-    "pytest-asyncio",
-    "mypy",
-    "ruff",
-]
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
--- a/packages/aws-sdk-signers/pyproject.toml
+++ b/packages/aws-sdk-signers/pyproject.toml
@@ -34,8 +34,8 @@ exclude = [
  "tests",
 ]
 
-[project.optional-dependencies]
-test = [
+[dependency-groups]
+dev = [
     "freezegun",
     "pytest",
     "pytest-asyncio",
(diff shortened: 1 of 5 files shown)


packages/aws-sdk-signers/pyproject.toml
```diff
@@ -34,8 +34,8 @@ exclude = [
  "tests",
 ]
 
-[project.optional-dependencies]
-test = [
+[dependency-groups]
+dev = [
```
nateprewitt, line 38: Was there a reason we're using `dev` here? It looks like most examples and our generated code above are using `test` or `tests`. Ideally we want to align a single consistent identifier.

nateprewitt (approved):

packages/aws-sdk-signers/pyproject.toml
jonathan343: I initially thought `dev` was was best since it gets installed by default with `uv` sync. However, we can achieve similar behavior by adding more specific groups to the default groups list like: `default-groups = ["test", "lint", "typing"]`.

As for naming, I'm going to go align with what's used in the (PEP 735 examples](https://peps.python.org/pep-0735/#example-dependency-groups-table).

jonathan343: Updates are reflected in 10b4f2fece1c40d698648f3120b6eeeb7da7d5b3

PR #542, opened 2025-08-14
Add LICENSE and NOTICE files to generated client packages

jonathan343:
## Summary

The `LICENSE` and `NOTICE` files are currently added manually added to the client packages (see [example PR](https://github.com/awslabs/aws-sdk-python/pull/4)).

This PR updates the code generator to generate an Apache-2.0 `LICENSE` file and `NOTICE` file with `Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.` at the top-level of AWS client packages.

## Testing

- Verified `LICENSE` and `NOTICE` files were generated for AWS clients.
- Verified `LICENSE` and `NOTICE` files were NOT generated for non-AWS clients.
---


codegen/core/src/main/java/software/amazon/smithy/python/codegen/sections/NoticeSection.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package software.amazon.smithy.python.codegen.sections;
+
+import software.amazon.smithy.utils.CodeSection;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+/**
+ * This section controls the entire generated NOTICE
+ *
+ * <p>An integration may want to use this if they want to programmatically
```
SamRemis, line 13: Is there a reason we have the <p> tag here?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
```diff
@@ -274,6 +284,38 @@ private static void writeReadme(
         });
     }
 
+    private static void writeLicense(
+            GenerationContext context
+    ) {
+        context.writerDelegator().useFileWriter("LICENSE", writer -> {
```
SamRemis, line 290: How is this code AWS specific.  I haven't tested it, so I could be wrong, but I think this will get generated for all SDKs, even non-AWS ones.  

Does Smithy force a license that a customer must use?  I would expect this to be a generic license that was smithy's default, then have that be overwritten by an AWS specific interceptor.

SamRemis (changes requested):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
jonathan343: This PR now only adds our LICENSE and NOTICE file to AWS clients.

PR #545, opened 2025-08-22
Add standard retry mode

alexgromero:
*Description of changes:*
Adds support for a new retry mode: `standard`.  This behavior is now consistent with other AWS SDKs implementing standard mode. This mode also adds support for retry quotas which control how many unsuccessful retries a client can make.

### DIFF
--- a/packages/smithy-core/src/smithy_core/retries.py
+++ b/packages/smithy-core/src/smithy_core/retries.py
@@ -245,6 +245,59 @@ class SimpleRetryStrategy(retries_interface.RetryStrategy):
         """Not used by this retry strategy."""
 
 
+class StandardRetryQuota:
+    """Retry quota used by :py:class:`StandardRetryStrategy`."""
+
+    INITIAL_RETRY_TOKENS: int = 500
+    RETRY_COST: int = 5
+    NO_RETRY_INCREMENT: int = 1
+    TIMEOUT_RETRY_COST: int = 10
+
+    def __init__(self, initial_capacity: int = INITIAL_RETRY_TOKENS):
+        """Initialize retry quota with configurable capacity.
+
+        :param initial_capacity: The initial and maximum capacity for the retry quota.
+        """
+        self._max_capacity = initial_capacity
+        self._available_capacity = initial_capacity
+        self._lock = threading.Lock()
+
+    def acquire(self, *, error: Exception) -> int:
+        """Attempt to acquire capacity for a retry attempt.
+
+        If there's insufficient capacity available, raise an exception.
+        Otherwise, return the amount of capacity successfully allocated.
+        """
+        capacity_amount = self.RETRY_COST
+
+        with self._lock:
+            if capacity_amount > self._available_capacity:
+                raise RetryError("Retry quota exceeded")
+            self._available_capacity -= capacity_amount
+            return capacity_amount
+
+    def release(self, *, release_amount: int) -> None:
+        """Release capacity back to the retry quota.
+
+        The capacity being released will be truncated if necessary to ensure the max
+        capacity is never exceeded.
+        """
+        increment = self.NO_RETRY_INCREMENT if release_amount == 0 else release_amount
+
+        if self._available_capacity == self._max_capacity:
+            return
+
+        with self._lock:
+            self._available_capacity = min(
+                self._available_capacity + increment, self._max_capacity
+            )
+
+    @property
+    def available_capacity(self) -> int:
+        """Return the amount of capacity available."""
+        return self._available_capacity
+
+
 @dataclass(kw_only=True)
 class StandardRetryToken:
     retry_count: int
@@ -266,6 +319,7 @@ class StandardRetryStrategy(retries_interface.RetryStrategy):
         *,
         backoff_strategy: retries_interface.RetryBackoffStrategy | None = None,
         max_attempts: int = 3,
+        retry_quota: StandardRetryQuota | None = None,
     ):
         """Standard retry strategy using truncated binary exponential backoff with full
         jitter.
@@ -275,6 +329,9 @@ class StandardRetryStrategy(retries_interface.RetryStrategy):
 
         :param max_attempts: Upper limit on total number of attempts made, including
             initial attempt and retries.
+
+        :param retry_quota: The retry quota to use for managing retry capacity. Defaults
+            to a new :py:class:`StandardRetryQuota` instance.
         """
         if max_attempts < 0:
             raise ValueError(
@@ -287,7 +344,7 @@ class StandardRetryStrategy(retries_interface.RetryStrategy):
             jitter_type=ExponentialBackoffJitterType.FULL,
         )
         self.max_attempts = max_attempts
-        self._retry_quota = StandardRetryQuota()
+        self._retry_quota = retry_quota or StandardRetryQuota()
 
     def acquire_initial_retry_token(
         self, *, token_scope: str | None = None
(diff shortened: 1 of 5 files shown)


packages/smithy-aws-core/CHANGES.md
```diff
@@ -12,6 +12,7 @@
 ### Features
 
 * Added a hand-written implmentation for the `restJson1` protocol.
+* Added a new retry mode `standard` and made it the default retry strategy.
```
alexgromero, line 15: Is this considered a feature or breaking change since we're updating a default?

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,158 @@ def refresh_retry_token_for_retry(
...
+class StandardRetryStrategy(retries_interface.RetryStrategy):
+    def __init__(self, *, max_attempts: int = 3):
+        """Standard retry strategy using truncated binary exponential backoff with full
+        jitter.
+
+        :param max_attempts: Upper limit on total number of attempts made, including
+            initial attempt and retries.
+        """
+        self.backoff_strategy = ExponentialRetryBackoffStrategy(
+            backoff_scale_value=1,
```
jonathan343, line 272: nit - Not really relevant to this PR, but we should consider adding some simple validation for these values on initialization (e.g., ensure they're positive integers).

jonathan343 (dismissed):
Thanks Alessandra!
Looks great so far. I have a few suggestions below.

SamRemis (changes requested):

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,158 @@ def refresh_retry_token_for_retry(
...
+
+        :param token_scope: This argument is ignored by this retry strategy.
+        """
+        retry_delay = self.backoff_strategy.compute_next_backoff_delay(0)
+        return StandardRetryToken(retry_count=0, retry_delay=retry_delay)
+
+    async def refresh_retry_token_for_retry(
+        self,
+        *,
+        token_to_renew: StandardRetryToken,
```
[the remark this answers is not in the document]
alexgromero: Good catch. Thanks for the suggestions! I went with the second option and the type checks are passing now.

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,178 @@ def refresh_retry_token_for_retry(
...
+        """Standard retry strategy using truncated binary exponential backoff with full
+        jitter.
+
+        :param max_attempts: Upper limit on total number of attempts made, including
+            initial attempt and retries.
+        """
+        if max_attempts < 1:
+            raise ValueError(
+                f"max_attempts must be a positive integer, got {max_attempts}"
+            )
```
nateprewitt: If it's set to `0` that just means we won't retry, right? This error message seems misleading if that's the case.

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,178 @@ def refresh_retry_token_for_retry(
...
+                retry_count=retry_count,
+                retry_delay=retry_delay,
+                quota_consumed=total_quota,
+                last_quota_acquired=quota_acquired,
+            )
+        else:
+            raise RetryError(f"Error is not retryable: {error}") from error
+
+    async def record_success(self, *, token: retries_interface.RetryToken) -> None:
+        """Return token after successful completion of an operation.
```
nateprewitt: nit; "Return" may not be the right word to use here. We're not returning from this method, we may want to say "Release". It seems like this sentence and the next one may be somewhat redundant?

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,178 @@ def refresh_retry_token_for_retry(
...
+        self._lock = asyncio.Lock()
+
+    async def acquire(self, *, error: Exception) -> int:
+        """Attempt to acquire a certain amount of capacity.
+
+        If there's no sufficient amount of capacity available, raise an exception.
+        Otherwise, we return the amount of capacity successfully allocated.
+        """
+        # TODO: update `is_timeout` when `is_timeout_error` is implemented
+        is_timeout = False
```
nateprewitt: If we have `is_timeout_error` done, should we merge that before we add this? We should try to avoid adding non-functioning code.

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,178 @@ def refresh_retry_token_for_retry(
...
+
+
+class RetryStrategyMode(Enum):
+    """Enumeration of available retry strategies."""
+
+    SIMPLE = "simple"
+    STANDARD = "standard"
+
+
+RETRY_MODE_MAP = {
```
nateprewitt: How are we envisioning this being used and what's prompting adding it now? It doesn't look like it's actually implemented in this PR unless I'm missing it?

packages/smithy-core/tests/unit/test_retries.py
```diff
@@ -55,48 +60,228 @@ def test_exponential_backoff_strategy(
...
+    )
+    assert token.quota_consumed == first_consumed + StandardRetryQuota.RETRY_COST
+
+
+async def test_standard_retry_invalid_max_attempts() -> None:
+    with pytest.raises(ValueError):
+        StandardRetryStrategy(max_attempts=0)
+
+    with pytest.raises(ValueError):
+        StandardRetryStrategy(max_attempts=-1)
```
nateprewitt: Similar note here, parameterizing this gives us a distinct failure for each test case. Then we know what all is broken at once. If the first statement fails and we fix it, we won't know the second is broken until we run the tests again.

nateprewitt (commented):
Looks like we're off to a good start. I left some high-level feedback, but let me know if I can clarify anything!

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,158 @@ def refresh_retry_token_for_retry(
...
+
+class StandardRetryStrategy(retries_interface.RetryStrategy):
+    def __init__(self, *, max_attempts: int = 3):
+        """Standard retry strategy using truncated binary exponential backoff with full
+        jitter.
+
+        :param max_attempts: Upper limit on total number of attempts made, including
+            initial attempt and retries.
+        """
+        self.backoff_strategy = ExponentialRetryBackoffStrategy(
```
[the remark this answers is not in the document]
nateprewitt: This is probably a higher level decision we should make. Requirements from the spec are giving us what the default behavior needs to be.

Nothing is stopping from someone implementing their own RetryStrategy that's copying this and adjusting it. At that point, should we surface the primitives so they can be updated as needed? It seems like a lot of overhead to reimplement an entire RetryStrategy just to extend your backoff time.

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -240,5 +241,178 @@ def refresh_retry_token_for_retry(
...
+            backoff_scale_value=1,
+            max_backoff=20,
+            jitter_type=ExponentialBackoffJitterType.FULL,
+        )
+        self.max_attempts = max_attempts
+        self._retry_quota = StandardRetryQuota()
+
+    async def acquire_initial_retry_token(
+        self, *, token_scope: str | None = None
```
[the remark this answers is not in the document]
nateprewitt: We can leave it as is for the moment, but let's talk to Jordon about long term intent. I'll need to go look at the reference architecture again to see if there's info there. It seems like we should have our reasoning in the [design](https://github.com/smithy-lang/smithy-python/blob/develop/designs/retries.md) but it's not there right now. That'll help future implementers both on our team and externally.

packages/smithy-core/src/smithy_core/retries.py
alexgromero: I chatted with Sam about this and he mentioned he would update this TODO in the timeout error PR https://github.com/smithy-lang/smithy-python/pull/590 since we're merging to a feature branch. Should I remove the timeout logic entirely for now and let him add it, or leave the TODO as-is since it'll be addressed soon?

packages/smithy-core/src/smithy_core/retries.py
alexgromero: I added it based on the design doc which specifies that users can pass `RetryStrategyMode` enum to have a strategy constructed for them like this:
```
smithy_core.retries import RetryStrategyMode
retry_mode = RetryStrategyMode.Adaptive
bedrock_client = BedrockRuntimeClient(
    config=Config(
        retry_strategy = retry_mode
    )
)
```
Should I remove it for now and add it when we need it?

packages/smithy-core/tests/unit/test_retries.py
```diff
@@ -55,48 +60,228 @@ def test_exponential_backoff_strategy(
...
+    with pytest.raises(RetryError):
+        await strategy.refresh_retry_token_for_retry(token_to_renew=token, error=error)
+
+
+async def test_standard_retry_does_not_retry_unsafe() -> None:
+    strategy = StandardRetryStrategy()
+    error = CallError(fault="client", is_retry_safe=False)
+    token = await strategy.acquire_initial_retry_token()
+    with pytest.raises(RetryError):
+        await strategy.refresh_retry_token_for_retry(token_to_renew=token, error=error)
```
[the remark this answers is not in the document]
alexgromero: That makes sense. I can refactor these. I was following the test patterns from `SimpleRetryStrategy` tests, but I agree parameterization would reduce redundancy.

packages/smithy-core/tests/unit/test_retries.py
```diff
@@ -55,48 +60,228 @@ def test_exponential_backoff_strategy(
...
+    initial_capacity = strategy._retry_quota.available_capacity  # pyright: ignore[reportPrivateUsage]
+
+    await strategy.record_success(token=token)
+
+    # Should increment by NO_RETRY_INCREMENT
+    expected = min(
+        initial_capacity + StandardRetryQuota.NO_RETRY_INCREMENT,
+        StandardRetryQuota.INITIAL_RETRY_TOKENS,
+    )
+    assert strategy._retry_quota.available_capacity == expected  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
alexgromero: Got it. I'm not sure what the best way to verify the quota behavior is. Should I make it a public property, or test it in a different way? For the functional tests, I added a helper function [`get_retry_quota`](https://github.com/smithy-lang/smithy-python/pull/601/files#diff-e4eb2fd89418e9e2ecb92e8320b3a7a62686e599861843a32a6bef8e67d6b41bR9-R10) but I'm guessing that's not an ideal approach either?

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -242,3 +243,174 @@ def refresh_retry_token_for_retry(
...
+
+@dataclass(kw_only=True)
+class StandardRetryToken:
+    retry_count: int
+    """Retry count is the total number of attempts minus the initial attempt."""
+
+    retry_delay: float
+    """Delay in seconds to wait before the retry attempt."""
+
+    quota_consumed: int = 0
```
nateprewitt: What are we doing with this? It looks like we're tracking it on the token but I don't actually see it used elsewhere?

nateprewitt (dismissed):

SamRemis (approved):

PR #550, opened 2025-08-28
Update to Gradle 9

JordonPhillips:
This updates the gradle wrapper to version 9. We weren't using anything that had breaking changes, thankfully.

I also integrated the version bump from #549 as that failed due to the embedded kotlin version being too old. The embedded version in Gradle 9 is what it wants.


jonathan343 (approved):
LGTM!

PR #552, opened 2025-09-04
Add Simple Changelog Management

jonathan343:
### Summary

Changelog entries are currently being managed manually for each smithy package. This PR includes the following changes to help automate this process and improve consistency:
- Rename changelog files from `CHANGES.md` to `CHANGELOG.md`
- Adds a `new-entry.py` script that creates a new change entry under a package's `.changes/next-release` directory.
- Adds a `new-release.py` script that consolidates all next-release changes into a permanent `.changes/x.y.z.json` file.
- Adds a `render.py` script that re-renders a package's CHANGELOG.md file using all the `.changes/<version>.json` entries.
- Backfills all existing changelog entries for each package and re-renders each changelog file.

---

### Workflow Eample

1. Generate a new changelog entry:
```
./scripts/changelog/new-entry.py -p <package_name> -t <type> -d <description>
```

This generates a file like`<package_name>/.changes/next-release/<unique_id>.json` with the following content:
```
{
  "type": "<type>",
  "description": "<description>"
}
```

2. Generate a new release
```
./scripts/changelog/new-release.py -p <package_name> -v <version>
```

This generates a new file like `<package_name>/.changes/<version>.json` with all changes to be released:
```
{
  "changes": [
    {
      "type": "<type>",
      "description": "<description>"
    },
    {
      "type": "<type>",
      "description": "<description>"
    }
  ]
}
```

3. Render the new changelog:
```
./scripts/changelog/render.py -p <package_name> -o <path_to_changelog>
```

This command will re-render the changelog and replace the content at `<path_to_changelog>`
 
---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #553, opened 2025-09-05
fix: HTTPAPIKeyAuthTrait initialization

nated0g:
*Issue #, if available:*

*Description of changes:*

The initializer of the `HTTPAPIKeyAuthTrait` as currently implemented is incorrect. As it is currently written, initialization will always fail. The `httpApiKeyAuth` trait in the generated schema code ends up looking like this:

```py
Trait.new(
    id=ShapeID("smithy.api#httpApiKeyAuth"),
    value=MappingProxyType({"name": "Authorization", "in": "header"}),
),
```

When building:
```sh
  File "/app/smithy/build/conversations/python-client-codegen/src/conversations/_private/schemas.py", line 1036, in <module>
    Trait.new(
    ~~~~~~~~~^
        id=ShapeID("smithy.api#httpApiKeyAuth"),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        value=MappingProxyType({"name": "Authorization", "in": "header"}),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ),
    ^
  File "/app/.venv/lib/python3.13/site-packages/smithy_core/traits.py", line 91, in new
    return cls(value)
  File "/app/.venv/lib/python3.13/site-packages/smithy_core/traits.py", line 338, in __init__
    object.__setattr__(self, "location", APIKeyLocation(value))
                                         ~~~~~~~~~~~~~~^^^^^^^
  File "/usr/local/lib/python3.13/enum.py", line 726, in __call__
    return cls.__new__(cls, value)
           ~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/enum.py", line 1203, in __new__
    raise ve_exc
ValueError: mappingproxy({'name': 'Authorization', 'in': 'header'}) is not a valid APIKeyLocation
```

The current implementation tries to construct an `APIKeyLocation` from this entire object, instead of from the required `in` property (https://smithy.io/2.0/spec/authentication-traits.html#httpapikeyauth-trait), as it should.

This change constructs the APIKeyLocation correctly from the `in` property

---

### DIFF
--- a/packages/smithy-core/src/smithy_core/traits.py
+++ b/packages/smithy-core/src/smithy_core/traits.py
@@ -335,10 +335,13 @@ class HTTPAPIKeyAuthTrait(Trait, id=ShapeID("smithy.api#httpApiKeyAuth")):
 
     def __init__(self, value: "DocumentValue | DynamicTrait" = None):
         super().__init__(value)
-        object.__setattr__(self, "location", APIKeyLocation(value))
         assert isinstance(self.document_value, Mapping)
+        assert isinstance(self.document_value["in"], str)
         assert isinstance(self.document_value["name"], str)
         assert isinstance(self.document_value.get("scheme"), str | None)
+
+        location = self.document_value["in"]
+        object.__setattr__(self, "location", APIKeyLocation(location))
 
     @property
     def name(self) -> str:


jonathan343 (changes requested):
Hey @nated0g,

Thanks for opening this pull request!
I think this looks good but the type checker is probably gonna complain since this is being added before the `assert isinstance(self.document_value, Mapping)` line.

I'll run our CI so you can see what I'm taking about.
All you'll need to do is move these lines below the one I mentioned above.

nated0g:
> Hey @nated0g,
> 
> Thanks for opening this pull request! I think this looks good but the type checker is probably gonna complain since this is being added before the `assert isinstance(self.document_value, Mapping)` line.
> 
> I'll run our CI so you can see what I'm taking about. All you'll need to do is move these lines below the one I mentioned above.

Good call, done.

jonathan343 (approved):
LGTM!

jonathan343:
Thanks again @nated0g!
I'll add a bugfix changelog entry for this in https://github.com/smithy-lang/smithy-python/pull/552

PR #554, opened 2025-09-08
fix: api key identity resolver config codegen

nated0g:
*Issue #, if available:*

*Description of changes:*

`self.api_key_identity_resolver` was never being set on the generated `Config`, only a local `api_key_identity_resolver` was being set (and then never being used).

This was leading to an error when trying to make requests to a service with api key auth.

`'Config' object has no attribute 'api_key_identity_resolver'`

---


jonathan343 (approved):
LGTM!

PR #578, opened 2025-10-17
Fix host header port mismatch in signing

SamRemis:
This PR includes a bugfix which breaks signing any requests that are sent to a non-default port.  

### Problem:
There's a mismatch between the host header used for signing vs. the actual request header when connecting to non-default ports.

• **During signing**: The host header includes port (from URI netloc) and gets added to signed headers.  The header does not get added to the final request.
• **During request creation**: The host header excludes port (from URI host) if not already present.
• **Result**: The request is signed with `example.com:8080` but sent with the host header having a value of `example.com`.  This creates a signature validation issue server-side.

### Solution:
This PR updates the CRT client to use netloc, similar to the signer.  These values should always match if a given HTTP client is compliant with the HTTP RFCs.  

### DIFF
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -272,7 +272,7 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
         :py:class:`smithy_http.aio.HTTPRequest`"""
         headers_list: list[tuple[str, str]] = []
         if "host" not in request.fields:
-            host_header = request.destination.hostname
+            host_header = request.destination.host
             if port := request.destination.port:
                 host_header += f":{port}"
             request.fields.set_field(Field(name="host", values=[host_header]))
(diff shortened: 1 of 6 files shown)


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -342,7 +342,7 @@ async def _marshal_request(
         headers_list = []
         if "host" not in request.fields:
             request.fields.set_field(
-                Field(name="host", values=[request.destination.host])
+                Field(name="host", values=[request.destination.netloc])
```
jonathan343: We should look into if `netloc` is the right choice. It would follow the format of `{username}:{password}@{host}:{port}`.

I see ruby does `{host}:{port}` ([src](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L636-L643))

jonathan343 (approved):

PR #582, opened 2025-10-21
Fix characters interpreted as Smithy format specifiers in doc gen

arandito:
#### Description

Fixed an issue in doc string generation where special characters (specifically $) in the Smithy model's documentation traits were being incorrectly interpreted as Smithy format specifiers ([docs](https://smithy.io/2.0/guides/building-codegen/generating-code.html#formatters)). 

The fix adds the `$L` formatter to `write()` and `writeInline()` calls in `MarkdownToRstDocConverter.java`, ensuring doc strings are treated as string literals rather than as format expressions. This prevents the unintended interpolation of doc strings that contain characters like $.

#### Testing

Successfully rebuilt `aws-sdk-bedrock-runtime` client and ensured no changes were reflected in docstrings.

### DIFF
--- a/codegen/aws/core/src/test/java/software/amazon/smithy/python/aws/codegen/MarkdownToRstDocConverterTest.java
+++ b/codegen/aws/core/src/test/java/software/amazon/smithy/python/aws/codegen/MarkdownToRstDocConverterTest.java
@@ -106,4 +106,13 @@ public class MarkdownToRstDocConverterTest {
         String result = markdownToRstDocConverter.convertCommonmarkToRst(html);
         assertEquals(expected, result.trim());
     }
+
+    @Test
+    public void testConvertCommonmarkToRstWithFormatSpecifierCharacters() {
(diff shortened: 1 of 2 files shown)


jonathan343 (commented):
Thanks for this PR!

I think this looks good. We should consider adding some test coverage to this file: https://github.com/smithy-lang/smithy-python/blob/3c3bacffdcfeda6dfede98a780e107a60bb8ac4f/codegen/aws/core/src/test/java/software/amazon/smithy/python/aws/codegen/MarkdownToRstDocConverterTest.java. Maybe something simple like being able to handle the `$` character and any other cases you're aware of

ubaskota (dismissed):

arandito:
Added a unit test to cover case where HTML string has the character $. This will also cover any future format specifiers as we are treating the entire string as a literal.

Ran `make build-java` to ensure test fails with the current state on `develop` and that it passes with the PR fix.

jonathan343 (approved):
LGTM, thanks Antonio!

PR #586, opened 2025-10-23
Resolve signing issues from mismatched list header formatting

jonathan343:
## Overview

This PR fixes `InvalidSignatureException` errors caused by inconsistent formatting of list-valued headers during request signing. Previously, headers like were sent as `value1,value2` but signed as `value1, value2`, leading to signature mismatches. The SDK now ensures both the request and signature use identical header formatting.

> [!NOTE]
> This is a bit weird since the CRT will handle multi-valued headers by adding them individually on the request. The server seems to be handling this by joining them by the `,` for signature verification.

## Testing

Tested this change using the script below with both the `AWSCRTHTTPClient` and `AIOHTTPClient` clients:

```py
import asyncio
import json

from aws_sdk_bedrock_runtime.models import InvokeModelWithResponseStreamInput

MODEL_ID = "us.amazon.nova-lite-v1:0"


async def test_async():
    c = BedrockRuntimeClient(
        config=Config(
            endpoint_uri="https://bedrock-runtime.us-east-1.amazonaws.com",
            region="us-east-1",
            transport=AIOHTTPClient() # Comment to use the default CRT client
        )
    )
    # Define your system prompt(s).
    system_list = [
        {
            "text": "Act as a creative writing assistant. When the user provides you with a topic, write a short story about that topic."
        }
    ]

    # Define one or more messages using the "user" and "assistant" roles.
    message_list = [{"role": "user", "content": [{"text": "A camping trip"}]}]

    # Configure the inference parameters.
    inf_params = {"maxTokens": 500, "topP": 0.9, "topK": 20, "temperature": 0.7}

    request_body = {
        "schemaVersion": "messages-v1",
        "messages": message_list,
        "system": system_list,
        "inferenceConfig": inf_params,
    }
    response = c.invoke_model_with_response_stream(
        InvokeModelWithResponseStreamInput(
            model_id=MODEL_ID,
            body=json.dumps(request_body).encode("utf-8"),
            content_type="application/json"
        )
    )

    async with await response as stream:
        output = ""
        async for r in stream.output_stream:
            try:
                text = json.loads(r.value.bytes_.decode())["contentBlockDelta"][
                    "delta"
                ]["text"]
                output += text
            except Exception:
                continue
        print(f"Response:\n------------\n{output}")


if __name__ == "__main__":
    asyncio.run(test_async())
```

Without this PR I was receiving the following error:
```
smithy_core.exceptions.CallError: Unknown error for operation com.amazonaws.bedrockruntime#InvokeModelWithResponseStream - status: 403 - id: com.amazonaws.bedrockruntime#InvalidSignatureException
``` 

With this change I received successful responses.

---

### DIFF
--- a/packages/smithy-http/.changes/next-release/smithy-http-bugfix-20251023131033.json
+++ b/packages/smithy-http/.changes/next-release/smithy-http-bugfix-20251023131033.json
@@ -1,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Fix `InvalidSignatureException` caused by mismatched list header formatting during signing (e.g., `MyHeader: value1, value2` vs `MyHeader: value1,value2`)."
+}
--- a/packages/aws-sdk-signers/src/aws_sdk_signers/interfaces/http.py
+++ b/packages/aws-sdk-signers/src/aws_sdk_signers/interfaces/http.py
@@ -57,7 +57,7 @@ class Field(Protocol):
         """Remove all matching entries from list."""
         ...
 
-    def as_string(self, delimiter: str = ", ") -> str:
+    def as_string(self, delimiter: str = ",") -> str:
         """Serialize the ``Field``'s values into a single line string."""
         ...
 
--- a/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
+++ b/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
@@ -705,7 +705,7 @@ class AsyncSigV4Signer:
 
     async def _normalize_signing_fields(self, *, request: AWSRequest) -> dict[str, str]:
         normalized_fields = {
-            field.name.lower(): field.as_string()
+            field.name.lower(): field.as_string(delimiter=",")
             for field in request.fields
             if self._is_signable_header(field.name.lower())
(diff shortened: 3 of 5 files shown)


jonathan343:
After further investigation, I've learned that it doesn't matter what delimiter is used to send multi-valued list headers. What really matters is that we sign with the `,` delimiter. I think it makes more sense to be consistent across the ecosystem though.

SamRemis (approved):

packages/smithy-http/.changes/next-release/smithy-http-bugfix-20251023131033.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Fix `InvalidSignatureException` caused by mismatched list header formatting during signing (e.g., `MyHeader: value1, value2` vs `MyHeader: value1,value2`)."
```
nateprewitt:
```suggestion
  "description": "Standardized header concatenation delimiter to "," to avoid signature mismatches (e.g., `MyHeader: value1, value2` vs `MyHeader: value1,value2`)."
```

PR #589, opened 2025-10-26
Fix missing deepcopy import for event stream operations

SamRemis:
The ClientGenerator was missing the deepcopy import in generateEventStreamOperation(). Services with only event stream operations fail when smithy-build runs `ruff check` on the generated code because the generated code calls deepcopy(self._config) without importing it. Non-event-stream operations [already had the import](https://github.com/SamRemis/smithy-python/blob/8000e36a2bbf82fffea075a6567221a674a257e3/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java#L134).

This adds the missing `writer.addStdlibImport("copy", "deepcopy")` line.


jonathan343 (approved):
:shipit:

PR #590, opened 2025-10-28
Add timeout error classification

SamRemis:
This pull request centralizes and clarifies timeout handling across the async client and transport layers by adding a transport-level error classification API and surfacing transport-detected timeouts as a dedicated ClientTimeoutError. Transports now return an ErrorInfo that indicates whether an exception represents a timeout and whether the fault is client- or server-side, and the core async client consults that information and raises ClientTimeoutError when appropriate so callers see a single, consistent exception for client-side timeouts.

ClientTransport implementations must now implement get_error_info(exception, **kwargs) and return an ErrorInfo indicating whether the exception is a timeout and whether the fault is client- or server-side. The break was required so the core async client can reliably classify transport errors and raise a single ClientTimeoutError for client-side timeouts.  This information is necessary for handling errors as a part of retries.


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -133,6 +136,22 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
     _HTTP_PORT = 80
     _HTTPS_PORT = 443
 
+    def get_error_info(self, exception: Exception, **kwargs: Any) -> ErrorInfo:
+        """Get information about CRT errors."""
+
+        timeout_indicators = (
+            "AWS_IO_SOCKET_TIMEOUT",
+            "AWS_IO_CHANNEL_ERROR_SOCKET_TIMEOUT",
+            "AWS_ERROR_HTTP_REQUEST_TIMEOUT",
```
jonathan343: Where are these coming from? Is this an exhaustive list for CRT errors?

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -133,6 +136,22 @@ class AWSCRTHTTPClient(http_aio_interfaces.HTTPClient):
...
+        )
+        if isinstance(exception, TimeoutError):
+            return ErrorInfo(is_timeout_error=True, fault="client")
+
+        if isinstance(exception, AwsCrtError) and exception.name in timeout_indicators:
+            return ErrorInfo(is_timeout_error=True, fault="client")
+
+        return ErrorInfo(is_timeout_error=False)
```
jonathan343: why set `fault="client"` above and not here?

jonathan343 (commented):
I need to take a deeper look, but I don't think we're catching the new exception somewhere. I hacked around the CRT client to set the `connect_timeout_ms` to `10`. The result is the SDK will keep trying to send the request indefinitely. When debugging I see the response is:

```
Client timeout occurred: AWS_IO_SOCKET_TIMEOUT: socket operation timed out.
```

**Update:** This might be specific to the simple retry mode not being able to handle this. I need a bit more investigation time to get a more clean answer though

packages/smithy-core/src/smithy_core/aio/interfaces/__init__.py
```diff
@@ -86,7 +96,23 @@ async def resolve_endpoint(self, params: EndpointResolverParams[Any]) -> Endpoin
...
 
 class ClientTransport[I: Request, O: Response](Protocol):
-    """Protocol-agnostic representation of a client tranport (e.g. an HTTP client)."""
+    """Protocol-agnostic representation of a client transport (e.g. an HTTP client).
+
+    Transport implementations must define the get_error_info method to determine which
+    exceptions represent timeout conditions for that transport.
+    """
+
+    def get_error_info(self, exception: Exception, **kwargs: Any) -> ClientErrorInfo:
```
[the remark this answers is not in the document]
nateprewitt: > The rationale behind the current implementation is that each client has its own unique set of exceptions that will be raised in the case of a timeout. These need to be caught and flagged as timeout exceptions so that our retry handlers can adjust the retry behavior appropriately.

I think we can accomplish this a few ways without needing to add this new concept. I think the original proposal for this was adding `TIMEOUT_ERRORS` and potentially other `FOO_ERRORS` constants onto each transport and making those part of the Protocol. Then we have a static location to look for this information across all transports and can write a single generalized function for error handling. Did we encounter issues with that approach?

Another option is to have the client's `send` method do that classification for us with its own error handling that ensures the subset of errors that should be classified as Timeouts are raised with an error that subclasses a general TimeoutError.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/HttpProtocolTestGenerator.java
```diff
@@ -634,6 +634,8 @@ class $3L:
                     def __init__(self, *, client_config: HTTPClientConfiguration | None = None):
                         self._client_config = client_config
 
+                    TIMEOUT_EXCEPTIONS = ()
+
```
nateprewitt: Conventionally, class constants should be at the top of the class, above the `__init__`.

packages/smithy-core/tests/unit/test_retries.py
```diff
@@ -2,6 +2,8 @@
 #  SPDX-License-Identifier: Apache-2.0
 
 import pytest
+
+from smithy_core import SmithyError
```
nateprewitt:
```suggestion
```

packages/smithy-http/src/smithy_http/aio/aiohttp.py
```diff
@@ -52,6 +52,8 @@ def __post_init__(self) -> None:
 class AIOHTTPClient(HTTPClient):
     """Implementation of :py:class:`.interfaces.HTTPClient` using aiohttp."""
 
+    TIMEOUT_EXCEPTIONS = (TimeoutError,)
```
nateprewitt: Where is this coming from?

packages/smithy-http/tests/unit/aio/test_protocols.py
```diff
@@ -11,14 +12,15 @@
...
-from smithy_http.aio import HTTPRequest
+from smithy_http.aio import HTTPRequest, HTTPResponse
 from smithy_http.aio.interfaces import HTTPRequest as HTTPRequestInterface
 from smithy_http.aio.interfaces import HTTPResponse as HTTPResponseInterface
-from smithy_http.aio.protocols import HttpClientProtocol
+from smithy_http.aio.protocols import HttpBindingClientProtocol, HttpClientProtocol
 
 
-class TestProtocol(HttpClientProtocol):
+class MockProtocol(HttpClientProtocol):
```
nateprewitt: Why did we change the name of this?

packages/smithy-http/tests/unit/aio/test_protocols.py
```diff
@@ -135,3 +137,26 @@ def test_http_protocol_joins_uris(
...
+            destination=URI(host="example.com"), method="POST", fields=Fields()
+        ),
+        response=response,
+        response_body=b"",
+        error_registry=TypeRegistry({}),
+        context=ConcreteTypedProperties(),
+    )
+
+    assert error.is_timeout_error is True
+    assert error.fault == "client"
```
nateprewitt: This seems like it may make more sense as a functional test than in the unit tests. We shouldn't be invoking `_create_error()` directly in the tests.

nateprewitt (changes requested):

nateprewitt (approved):

PR #592, opened 2025-10-30
Switch to Google-style docstrings with Markdown and remove Sphinx/RST generation

arandito:
### Description

This PR updates the generated docstring format to use the [Google style](https://google.github.io/styleguide/pyguide.html#s3.8-comments-and-docstrings) with Markdown descriptions for ***all services*** (AWS and non-AWS).

This PR also removes the Sphinx-based reStructuredText (RST) documentation generation system introduced in #418. All documentation stubs will now be generated at build time in the `awslabs/aws-sdk-python` repo. This will prevent static doc stubs from bloating the repo when 400+ services are supported.

#### Key changes:
  - Updates all code generators (`Client`, `Structure`, `Config`, `Enum`, `Union`) to generate Google style docstrings with Markdown descriptions using the new `MarkdownConverter` class
 - Adds [pandoc](https://pandoc.org/) CLI tool as a required build dependency to convert Smithy model documentation strings to Markdown
  - Updates README with new pandoc dependency
  - Removes the `AwsRstDocFileGenerator` plugin that generates `.rst` files for Sphinx doc gen 
  - Remove docs dependencies and Sphinx configuration files from generated clients
  - Removes RST-to-Markdown conversion logic

> [!NOTE]
> The new docstring format in AWS clients enables us to generate documentation using [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/getting-started/). We will generate MkDocs stubs in [awslabs/aws-sdk-python](https://github.com/awslabs/aws-sdk-python) that work with the [`mkdocstrings`](https://mkdocstrings.github.io/) tool. This will automatically create documentation from docstrings for all clients.

### Testing
- Added unit tests
- Regenerated Bedrock Runtime and Transcribe Streaming clients and confirmed all docstrings were updated
  - PR for updated clients: awslabs/aws-sdk-python#24

> [!IMPORTANT]
> For local testing, please install [pandoc v3.8.2](https://pandoc.org/installing.html) before running code generator. 


codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsMkDocsFileGenerator.java
```diff
@@ -0,0 +1,525 @@
...
+            String symbolName = symbol.getName();
+            String docsFileName = String.format("docs/structures/%s.md", symbolName);
+            // Don't generate separate docs for input/output structures (they're included with operations)
+            if (!shape.hasTrait(InputTrait.class) && !shape.hasTrait(OutputTrait.class)) {
+                structures.add(symbolName);
+                context.writerDelegator().useFileWriter(docsFileName, "", writer -> {
+                    writer.write("::: " + symbol.toString());
+                    writer.write("""
+                                options:
+                                    heading_level: 1
```
jonathan343: Can we change the default heading_level to 1 so we don't need to specify this case? Then if we need to go lower, we can be explicit

Related Docs: https://mkdocstrings.github.io/python/usage/configuration/headings/#heading_level

jonathan343 (commented):
Thanks Antonio, this is a great start!

This PR can really be broken down into two main components:
1. Formatting docstrings to comply with the requirements of MkDocs
2. Generating the MkDocs specific files and static `.md`

The first 100% needs to happen at codegen time. However, the more I look at this, the more I feel like doing 2 during codegen time and committing all the static files will really bloat our SDK repo when we're working with 400+ clients. This [open PR](https://github.com/awslabs/aws-sdk-python/pull/24/files) is adding 4000+ lines for one client. The more scalable approach here might be to do all this generation on the fly when needed like we do with botocore.

I don't see anything in the generated files in https://github.com/awslabs/aws-sdk-python/pull/24/files that would make it difficult to do the approach mentioned above. The generated clients themselves have all the information we need.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/MarkdownConverter.java
```diff
@@ -0,0 +1,169 @@
...
+     * @throws InterruptedException if process is interrupted
+     */
+    private static String convertWithPandoc(String input, String fromFormat, String toFormat)
+            throws IOException, InterruptedException {
+        ProcessBuilder processBuilder = new ProcessBuilder(
+                "pandoc",
+                "--from=" + fromFormat,
+                "--to=" + toFormat,
+                "--wrap=auto",
+                "--columns=72");
```
jonathan343: Why this number?

jonathan343 (commented):
This looks good so far, thanks Antonio! Just had a few comments/questions.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsMkDocsFileGenerator.java
jonathan343: This comment isn't relevant anymore since the .md files aren't going to be generated statically during codegen.

jonathan343 (approved):
LGTM

PR #600, opened 2025-11-07
Implement Retry Strategy Resolver Pattern with Per-Client Caching

SamRemis:
### Summary
Introduces a resolver pattern for retry strategies that enables per-client caching and state management. This allows multiple operations from the same client to share a single retry strategy instance, which is important for strategies that maintain state across retries such as token buckets and rate limiters.

### Motivation
Previously, each operation created its own retry strategy instance, preventing state sharing across operations from the same client. This change implements a resolver-based approach.

### Key Changes
- Added a new `RetryStrategyResolver` protocol similar to `AuthSchemeResolver`.  This will be used for objects which will supply clients with retry strategies.
- Added a `CachingRetryStrategyResolver` which implements the protocol and returns a unique instance of a requested retry strategy object per unique combination of max attempts setting and retry mode setting.  This will get called by the client when no RetryStrategy is specified.  
- Added `RetryStrategyOptions` for users to configure their retry strategy mode and max attempts per client call.

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
@@ -55,17 +55,20 @@ public final class ConfigGenerator implements Runnable {
             ConfigProperty.builder()
                     .name("retry_strategy")
                     .type(Symbol.builder()
-                            .name("RetryStrategy")
-                            .namespace("smithy_core.interfaces.retries", ".")
-                            .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                            .name("RetryStrategy | RetryStrategyOptions")
+                            .addReference(Symbol.builder()
+                                    .name("RetryStrategy")
+                                    .namespace("smithy_core.interfaces.retries", ".")
+                                    .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                                    .build())
+                            .addReference(Symbol.builder()
+                                    .name("RetryStrategyOptions")
+                                    .namespace("smithy_core.retries", ".")
+                                    .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                                    .build())
                             .build())
-                    .documentation("The retry strategy for issuing retry tokens and computing retry delays.")
-                    .nullable(false)
-                    .initialize(writer -> {
-                        writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
-                        writer.addImport("smithy_core.retries", "SimpleRetryStrategy");
-                        writer.write("self.retry_strategy = retry_strategy or SimpleRetryStrategy()");
-                    })
+                    .documentation(
+                            "The retry strategy or options for configuring retry behavior. Can be either a configured RetryStrategy or RetryStrategyOptions to create one.")
                     .build(),
             ConfigProperty.builder()
                     .name("endpoint_uri")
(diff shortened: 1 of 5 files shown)


packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -1,12 +1,60 @@
...
+
+class CachingRetryStrategyResolver(RetryStrategyResolver[RetryStrategy]):
+    """Retry strategy resolver that caches retry strategies based on configuration options.
+
+    This resolver caches retry strategy instances based on their configuration to reuse existing
+    instances of RetryStrategy with the same settings.
+    """
+
+    def __init__(self) -> None:
+        self._cache: dict[RetryStrategyOptions, RetryStrategy] = {}
```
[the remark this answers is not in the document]
SamRemis: Honestly, I just didn't think to check if lru_cache was thread-safe before implementing this.   Thanks for catching that.  I've already updated it.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -55,17 +55,20 @@ public final class ConfigGenerator implements Runnable {
...
                             .build())
-                    .documentation("The retry strategy for issuing retry tokens and computing retry delays.")
-                    .nullable(false)
-                    .initialize(writer -> {
-                        writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
-                        writer.addImport("smithy_core.retries", "SimpleRetryStrategy");
-                        writer.write("self.retry_strategy = retry_strategy or SimpleRetryStrategy()");
-                    })
+                    .documentation(
+                            "The retry strategy or options for configuring retry behavior. Can be either a fully configured RetryStrategy or RetryStrategyOptions to create one.")
```
nateprewitt: nit; you don't need to configure the RetryStrategy, you can just pass it the default.
```suggestion
                            "The retry strategy or options for configuring retry behavior. Can be either a configured RetryStrategy or RetryStrategyOptions to create one.")
```

nateprewitt (dismissed):
Not sure what happened with the commits for the last change but once that's fixed this should be set.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
SamRemis: I misread this nit and thought you were asking for me to make it clearer that the retry strategy or options were optional as input parameters to this method.  After making a commit and reverting it, GitHub no longer allowed me to merge this commit, so I just had to make an identical one myself.

Thanks for pointing this out :)

nateprewitt (dismissed):

packages/smithy-core/src/smithy_core/retries.py
```diff
@@ -5,9 +5,60 @@
...
+            case "simple":
+                if max_attempts is None:
+                    return SimpleRetryStrategy()
+                else:
+                    return SimpleRetryStrategy(max_attempts=max_attempts)
+            case "standard":
+                if max_attempts is None:
+                    return StandardRetryStrategy()
+                else:
+                    return StandardRetryStrategy(max_attempts=max_attempts)
```
nateprewitt: This seems a bit awkward and easy to miss when adding new ones. Can we create a single kwargs list and pass it in to avoid the conditional swap for every method?

nateprewitt (approved):

PR #601, opened 2025-11-11
Add functional tests for standard retries

alexgromero:
*Issue #, if available:*

*Description of changes:*
Added functional tests for standard retries

### DIFF
--- a/packages/smithy-core/tests/functional/test_retries.py
+++ b/packages/smithy-core/tests/functional/test_retries.py
@@ -1,155 +1,133 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from asyncio import gather, sleep
+
 import pytest
 from smithy_core.exceptions import CallError, RetryError
-from smithy_core.retries import StandardRetryQuota, StandardRetryStrategy
+from smithy_core.interfaces import retries as retries_interface
+from smithy_core.retries import (
+    ExponentialBackoffJitterType,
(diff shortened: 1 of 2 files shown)


packages/smithy-core/tests/functional/test_retries.py
```diff
@@ -0,0 +1,133 @@
...
+from smithy_core.interfaces import retries as retries_interface
+from smithy_core.retries import (
+    ExponentialBackoffJitterType,
+    ExponentialRetryBackoffStrategy,
+    StandardRetryQuota,
+    StandardRetryStrategy,
+)
+
+
+async def retry_operation(
```
nateprewitt: We should probably add a TODO to refactor this to use a generated client from `smithy-testing` once we have that finalized. This is a great first step to make sure we have coverage in the mean time though.

packages/smithy-core/tests/functional/test_retries.py
```diff
@@ -0,0 +1,133 @@
...
+    strategy: retries_interface.RetryStrategy,
+    status_codes: list[int],
```
nateprewitt: Something I'm not sure if we're covering in this or the smithy-http client is what happens if we get an actual network issue like a timeout exception. Do we have code that's exercising that part of our retry process?

nateprewitt (dismissed):
Just a small question and an optional TODO to make sure we update these later. Otherwise, this looks great. Thanks @alexgromero!

nateprewitt (approved):
Thanks @alexgromero! ⛵

PR #604, opened 2025-11-21
Prepare smithy-core 0.2.0 release

SamRemis:
Prepare release of standard retry mode and add minor version bumps for it.

This release bumps all required packages for a minor release and bumps codegen to 0.1.0.  

### DIFF
--- a/packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
+++ b/packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
@@ -16,7 +16,7 @@ from .signers import (
 )
 
 __license__ = "Apache-2.0"
-__version__ = "0.2.0"
+__version__ = "0.1.0"
 
 __all__ = (
     "URI",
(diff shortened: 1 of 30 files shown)


packages/aws-sdk-signers/src/aws_sdk_signers/__init__.py
```diff
@@ -16,7 +16,7 @@
 )
 
 __license__ = "Apache-2.0"
-__version__ = "0.1.0"
+__version__ = "0.2.0"
```
jonathan343: Why does signers need a minor version bump? This is a leaf note in the dependency tree of the SDKs and `smithy-aws-core`. There are also no new changes being added to this package. Personally, I don't think we should be bumping this.

jonathan343 (changes requested):
Thanks Sam! Left a few comments. You're missing a few steps for each package in this PR:
- We need to run `./scripts/new-release -p <package_name> -v <package_version>`
- Then run `./scripts/render -p <package_name> -o packages/<package_name>/CHANGELOG.md`
- Was not including the bumps in [SmithyPythonDependency.java](https://github.com/smithy-lang/smithy-python/blob/develop/codegen/core/src/main/java/software/amazon/smithy/python/codegen/SmithyPythonDependency.java) intentional? We need that to have new SDKs get generated with the new minor versions

jonathan343 (dismissed):
LGTM! 🚢

jonathan343 (approved):

PR #606, opened 2025-11-24
Consolidate retry strategy resolver logic

SamRemis:
*Description of changes:*
Consolidate retry strategy resolver logic from being generated once per operation to being shared in a centralized function for each operation.

A few files are attached to show the diff and testing on the generated clients: 
- [The generated Bedrock Runtime client](https://github.com/user-attachments/files/24059096/generated-bedrock-client.py)
- [A diff of the generated bedrock runtime client compared to the client in the aws-sdk-python](https://github.com/user-attachments/files/24059192/generated-client-diff.patch)
- [Testing script to confirm functionality](https://github.com/user-attachments/files/23726493/retry_strategy_test_script.py)

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -101,6 +101,27 @@ final class ClientGenerator implements Runnable {
                         self._retry_strategy_resolver = RetryStrategyResolver()
                     """, configSymbol, pluginSymbol, writer.consumer(w -> writeDefaultPlugins(w, defaultPlugins)));
 
+            writer.addImport("smithy_core.retries", "RetryStrategyOptions");
+            writer.addImport("smithy_core.interfaces.retries", "RetryStrategy");
+            writer.write("""
+                    async def _resolve_retry_strategy(self, retry_strategy: RetryStrategy | RetryStrategyOptions | None) -> RetryStrategy:
+                        if isinstance(retry_strategy, RetryStrategy):
+                            return retry_strategy
+                        elif isinstance(retry_strategy, RetryStrategyOptions):
+                            return await self._retry_strategy_resolver.resolve_retry_strategy(
+                                options=retry_strategy
+                            )
+                        elif retry_strategy is None:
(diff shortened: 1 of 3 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -101,6 +101,27 @@ def __init__(self, config: $1T | None = None, plugins: list[$2T] | None = None):
...
+                            )
+                        elif retry_strategy is None:
+                            return await self._retry_strategy_resolver.resolve_retry_strategy(
+                                options=RetryStrategyOptions()
+                            )
+                        else:
+                            raise TypeError(
+                                f"retry_strategy must be RetryStrategy, RetryStrategyOptions, or None, "
+                                f"got {type(retry_strategy).__name__}"
+                            )
```
nateprewitt: Looking at this in the generated clients now, I don't think we need this when we already have `resolve_retry_strategy()`. We have this extra level of abstraction that's creating the clutter. Pushing this logic down to the resolver itself doesn't change behavior and better encapsulates resolution. This isn't a function of the client, it should be deferring that.

Something like this [here](https://github.com/smithy-lang/smithy-python/blob/2fe8d3ccebdc7e54dd1158a3b73e19fb17c96d40/packages/smithy-core/src/smithy_core/retries.py#L36C1-L44C1):

```python
async def resolve_retry_strategy(
        self, *, retry_strategy: RetryStrategy | RetryStrategyOptions | None
    ) -> RetryStrategy:
        """Resolve a retry strategy from the provided options, using cache when possible.

        :param retry_strategy: An explicitly configured retry strategy or options for creating one.
        """
        if isinstance(retry_strategy, RetryStrategy):
            return retry_strategy
        elif retry_strategy is None:
            retry_strategy = RetryStrategyOptions()
        
        if not isinstance(retry_strategy, RetryStrategyOptions):
            raise TypeError(
                f"retry_strategy must be RetryStrategy, RetryStrategyOptions, or None, "
                f"got {type(retry_strategy).__name__}"
            )   
        
        return self._create_retry_strategy(options.retry_mode, options.max_attempts)
```

nateprewitt (approved):

PR #607, opened 2025-11-24
Update error messages for timeout errors

SamRemis:
## Description 
This PR improves timeout error messages to provide more actionable information for debugging. Currently, timeout errors sometimes display messages like "Client timeout occurred: " with no additional context.  This occurs when the error's string representation is an empty string.

For the CRT client, we updated _CRTTimeoutError to include the underlying AWS error name and message, changing errors from `Client timeout occurred:`  to `CRT AWS_IO_SOCKET_TIMEOUT: socket operation timed out.`

For all clients, we now default to using the exception's class name when the string representation is empty.  This can occur in aiohttp when the [`total` attribute](https://github.com/aio-libs/aiohttp/blob/72fadb8f1a56e8bff0fef23ccf02e2067cd87e41/aiohttp/client.py#L192) is set on the `ClientTimeout` class, which will [raise a TimeoutError](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L727) with no additional info.  This updates the same blank error message to `A timeout error occurred.`.

Test file:
[timeout_exceptions_test.py](https://github.com/user-attachments/files/23918715/timeout_exceptions_test.py)

### DIFF
--- a/packages/smithy-core/src/smithy_core/aio/client.py
+++ b/packages/smithy-core/src/smithy_core/aio/client.py
@@ -469,8 +469,7 @@ class RequestPipeline[TRequest: Request, TResponse: Response]:
                     )
             except Exception as e:
                 if isinstance(e, self.transport.TIMEOUT_EXCEPTIONS):
-                    error_msg = str(e) or "A timeout error occurred."
-                    raise ClientTimeoutError(message=error_msg) from e
+                    raise ClientTimeoutError(message="A timeout error occurred.") from e
                 raise
 
             _LOGGER.debug("Received response: %s", transport_response)
--- a/packages/smithy-core/src/smithy_core/aio/client.py
+++ b/packages/smithy-core/src/smithy_core/aio/client.py
@@ -469,10 +469,8 @@ class RequestPipeline[TRequest: Request, TResponse: Response]:
                     )
             except Exception as e:
                 if isinstance(e, self.transport.TIMEOUT_EXCEPTIONS):
-                    error_msg = str(e) or type(e).__name__
-                    raise ClientTimeoutError(
-                        message=f"Client timeout occurred: {error_msg}"
-                    ) from e
+                    error_msg = str(e) or "A timeout error occurred."
+                    raise ClientTimeoutError(message=error_msg) from e
                 raise
 
             _LOGGER.debug("Received response: %s", transport_response)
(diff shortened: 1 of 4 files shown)


SamRemis:
Setting back to draft mode for now; I wanted to get this out there to confirm everyone is good with the two-pronged approach of both updating our CRT errors and having a default.  

I think the exception message should ultimately be something along the lines of `""Client timeout occurred: {error_msg}, please see underlying exception for more details."`, but wanted to get team input on final messaging.

packages/smithy-core/src/smithy_core/aio/client.py
```diff
@@ -469,8 +469,9 @@ async def _handle_attempt[I: SerializeableShape, O: DeserializeableShape](
                     )
             except Exception as e:
                 if isinstance(e, self.transport.TIMEOUT_EXCEPTIONS):
+                    error_msg = str(e) or type(e).__name__
```
nateprewitt: This seems like we're leaking a detail somewhere. Is this something coming from the CRT that's requiring us to pull the `__name__` off the error?

packages/smithy-http/.changes/next-release/smithy-http-enhancement-20251122132342.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "enhancement",
+  "description": "Improved error messaging for CRT-based timeout errors"
```
nateprewitt:
```suggestion
  "description": "Improved error messaging for CRT-based timeout errors."
```

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -187,7 +187,7 @@ async def send(
             return await self._await_response(crt_stream)
         except AwsCrtError as e:
             if e.name in self._TIMEOUT_ERROR_NAMES:
-                raise _CRTTimeoutError() from e
+                raise _CRTTimeoutError(f"CRT {e.name}: {e.message}")
```
nateprewitt: Ideally, we should be keeping the context with the `from` but we can include the message to make the immediate information clearer.

packages/smithy-core/src/smithy_core/aio/client.py
nateprewitt: `from` should give us the context (and the originating error). If they don't include anything, that's fine. By creating the chain, we can look to see what happened. Including `{e}` in the message to begin with is likely not necessary. That's an artifact of a time before we had better tracebacks.

nateprewitt: `A timeout error occurred.` is probably sufficient in this case.

nateprewitt (approved):

PR #621, opened 2025-12-24
Use UUID instead of timestamp for changelog entry filenames

jonathan343:
## Summary

Replace timestamp-based changelog entry filenames with UUID to prevent overwrites when generating multiple entries for the same package/type in automated workflows.

### Before

```
smithy-core-feature-20251224113018.json
```

### After

```
smithy-core-feature-c72e95f3e3df456c87d973f4b7bcd23e.json
```

---


nateprewitt (approved):
Thanks @jonathan343 :shipit:

PR #622, opened 2025-12-30
Smithy Release - December 30, 2025

nateprewitt:
This is a set of changes to release the 0.3.0 versions of `smithy-core` and `smithy-aws-core`, along with their dependencies. This will also bring the code generator up to date so that the next set of clients we release will pull in the minor version updates.

Changes for `smithy-aws-event-stream`, `smithy-http`, and `smithy-json` are being done as patch versions which can be picked up immediately as they're bugfixes or minor stylistic enhancements.

# Changelogs

### Breaking Changes
* Refactored `resolve_retry_strategy` to avoid code duplication per operation.

### Enhancements
* Improved default error message for instances of ClientTimeoutError.

### Dependencies
* Bump `smithy-core` from `~=0.2.0` to `~=0.3.0`.

### Dependencies
* Removed strict pinning on `smithy-core` in favor of client managed versions.

### Bug fixes
* Fixed empty error messaging for CRT-based timeout errors.

### Dependencies
* Removed strict pinning on `smithy-core` in favor of client managed versions.

### Dependencies
* Removed strict pinning on `smithy-core` in favor of client managed versions.

### DIFF
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsPythonDependency.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsPythonDependency.java
@@ -22,7 +22,7 @@ public class AwsPythonDependency {
      */
     public static final PythonDependency SMITHY_AWS_CORE = new PythonDependency(
             "smithy_aws_core",
-            "~=0.2.0",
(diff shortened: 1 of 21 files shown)


jonathan343 (approved):
:shipit:

PR #640, opened 2026-02-18
Add config source interface and environment variable source for config resolution

ubaskota:
*Issue #, if available:*

*Description of changes:*
This change adds the `ConfigSource` protocol to s`mithy-core` interfaces and implements `EnvironmentSource` in `smithy-aws-core` for resolving configuration values from environment variables. Includes unit tests covering protocol compliance, key-to-env-var mapping, edge cases, and no-caching behavior.

* Tests * 
Added unit tests to validate the expected behavior. Verified that tests pass.


SamRemis (approved):

packages/smithy-aws-core/src/smithy_aws_core/config/sources.py
```diff
@@ -0,0 +1,35 @@
...
+
+        :param key: The standard configuration key (e.g., 'region', 'retry_mode').
+
+        :returns: The value from the corresponding environment variable, or None if not set or empty.
+        """
+        env_var = f"{self._prefix}{key.upper()}"
+        config_value = os.environ.get(env_var)
+        if config_value is None:
+            return None
+        return config_value.strip()
```
arandito: I think we should call `strip()` before checking if the `config_value` is an empty string.

For example, if a customer provides `region="      "`, it will return `""` instead  of None. 

If we strip the config value prior to checking emptiness, it would correctly return None.

packages/smithy-aws-core/tests/unit/config/test_sources.py
```diff
@@ -0,0 +1,87 @@
...
+from smithy_aws_core.config.sources import EnvironmentSource
+
+
+class TestEnvironmentSource:
+    def test_source_name(self):
+        source = EnvironmentSource()
+        assert source.name == "environment"
+
+    def test_get_region_from_aws_region(self):
+        with patch.dict(os.environ, {"AWS_REGION": "us-west-2"}, clear=False):
```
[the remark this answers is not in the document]
arandito: We should still clear the environment vars in each test to keep them isolated. I think we only need to do this for the first `patch.dict` call in each test to clear the previous values before adding new ones in the test.

packages/smithy-aws-core/src/smithy_aws_core/config/sources.py
ubaskota: We want to use whatever value is set for that environment variable. If a value is set to an empty string, it'll return an empty string. The docstring will be updated.

jonathan343 (approved):
Looks good to me. There are one or two nits left you might want to consider

arandito (approved):
:shipit:

PR #641, opened 2026-02-20
Add ConfigResolver with necessary tests

ubaskota:
*Issue #, if available:*

*Description of changes:*

Implements `ConfigResolver` to resolve configuration values from multiple sources in precedence order. The resolver iterates through sources and returns the first non-None value found, along with the source name for provenance tracking. Returns `(None, None)` when no source provides a value.

*Tests* 
Added unit tests covering:
- Single and multiple source resolution
- Source precedence ordering
- Unresolved value handling
- Tuple return format validation


packages/smithy-aws-core/src/smithy_aws_core/config/resolver.py
```diff
@@ -0,0 +1,30 @@
...
+    This resolver is shared across all config objects and AWS service clients to avoid
+    redundant reads from environment variables, config files, etc.
+
+    :returns: The shared ConfigResolver instance
+    """
+    sources: list[ConfigSource] = [EnvironmentSource()]
+    return ConfigResolver(sources=sources)
+
+
+def reset_shared_resolver() -> None:
```
jonathan343: I don't think we should be exposing this as a public interface. If it's only for testing, can we move this into the test file?

packages/smithy-core/src/smithy_core/config/resolver.py
```diff
@@ -0,0 +1,36 @@
...
+    The resolver iterates through sources in precedence order, returning
+    the first non-None value found for a given configuration key.
+    """
+
+    def __init__(self, sources: list[ConfigSource]) -> None:
+        """Initialize the resolver with sources in precedence order.
+
+        :param sources: List of configuration sources in precedence order. The first
+            source in the list has the highest priority. The list is copied to
+            prevent external modification.
```
jonathan343: > The list is copied to prevent external modification.

We should add a test for this

packages/smithy-core/src/smithy_core/config/resolver.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ConfigResolver:
+    """Resolves configuration values from multiple sources.
+
+    The resolver iterates through sources in precedence order, returning
+    the first non-None value found for a given configuration key.
+    """
+
+    def __init__(self, sources: list[ConfigSource]) -> None:
```
jonathan343: nit - Maybe expand this to `Sequence` instead of `list` so we can also accept tuples like below:

```py
config_resolver = ConfigResolver(sources=(SourceA, SourceB))
```

packages/smithy-aws-core/src/smithy_aws_core/config/resolver.py
```diff
@@ -0,0 +1,30 @@
...
+from functools import lru_cache
+
+from smithy_core.config.resolver import ConfigResolver
+from smithy_core.interfaces.config import ConfigSource
+
+from .sources import EnvironmentSource
+
+
+@lru_cache(maxsize=1)
+def get_shared_resolver() -> ConfigResolver:
```
jonathan343: This implementation doesn't follow any of the approaches we discussed during the design phase.

jonathan343 (commented):
Thanks Ujjwal.

I left some minor comments and suggestions. However, my biggest concern with this PR is the approach for config resolver sharing.
1. It doesn't follow any of the approaches discussed in previous designs
2. In cases where multiple threads hit the function before the first result is cached, it will result in each building their own resolver which is unnecessary/duplicate work

SamRemis (approved):

jonathan343 (approved):

arandito (approved):

PR #642, opened 2026-02-24
Add a config property descriptor along with a custom resolver and validators

ubaskota:
*Issue #, if available:*

This PR adds a `ConfigProperty` descriptor that serves as the central instrument for configuration resolution. The descriptor is responsible for resolving config values from sources, applying custom resolvers for complex resolutions, validating values, caching them, and facilitating value updates. 

**The `ConfigProperty` descriptor implements the Python descriptor protocol to provide:**
  - Lazy resolution
  - Validation
  - Caching
  - Custom resolver
  - Source tracking

**Initial config variables:**
- `region` - demonstrates simple resolution
- `retry_strategy` - demonstrates complex resolution (aggregates retry_mode and max_attempts)

These two variables ensure that the design handles both simple and complex resolution.

**Tests**
- Verified that the unit tests for config property, validators and custom revolvers pass successfully.

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
@@ -5,7 +5,7 @@ from typing import cast
 from smithy_core.config.resolver import ConfigResolver
 from smithy_core.retries import RetryStrategyOptions, RetryStrategyType
 
-from smithy_aws_core.config.validators import validate_retry_mode
+from smithy_aws_core.config.validators import validate_max_attempts, validate_retry_mode
 
 
 def resolve_retry_strategy(
@@ -20,10 +20,10 @@ def resolve_retry_strategy(
 
     :param resolver: The config resolver to use for resolution
 
-    :returns: Tuple of (RetryStrategyOptions or None, source_name or None).
-        Returns (None, None) if neither retry_mode nor max_attempts is set.
+    :returns: Tuple of (RetryStrategyOptions, source_name) if both retry_mode and max_attempts
+        are resolved. Returns (None, None) if either value is missing.
 
-        For mixed sources, the source string includes both component sources:
+        For mixed sources, the source name includes both component sources:
         "retry_mode=environment, max_attempts=config_file"
     """
     # Get retry_mode
@@ -32,21 +32,20 @@ def resolve_retry_strategy(
     # Get max_attempts
     max_attempts, attempts_source = resolver.get("max_attempts")
 
-    # If neither is set, return None
-    if retry_mode is None and max_attempts is None:
-        return (None, None)
+    if retry_mode is None or max_attempts is None:
+        return None, None
 
-    if retry_mode is not None:
-        retry_mode = validate_retry_mode(retry_mode, mode_source)
-        retry_mode = cast(RetryStrategyType, retry_mode)
+    retry_mode = validate_retry_mode(retry_mode, mode_source)
+    retry_mode = cast(RetryStrategyType, retry_mode)
 
-    # Construct options with defaults
+    max_attempts = validate_max_attempts(max_attempts, attempts_source)
+
     options = RetryStrategyOptions(
-        retry_mode=retry_mode or "standard",
-        max_attempts=int(max_attempts) if max_attempts else None,
+        retry_mode=retry_mode,
+        max_attempts=max_attempts,
     )
 
     # Construct mixed source string showing where each component came from
-    source = f"retry_mode={mode_source or 'unresolved'}, max_attempts={attempts_source or 'unresolved'}"
+    source = f"retry_mode={mode_source}, max_attempts={attempts_source}"
 
     return (options, source)
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
@@ -1,0 +1,52 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from typing import cast
+
+from smithy_core.config.resolver import ConfigResolver
+from smithy_core.retries import RetryStrategyOptions, RetryStrategyType
+
+from smithy_aws_core.config.validators import validate_retry_mode
+
+
+def resolve_retry_strategy(
+    resolver: ConfigResolver,
+) -> tuple[RetryStrategyOptions | None, str | None]:
+    """Resolve retry strategy from multiple config keys.
+
+    Resolves both retry_mode and max_attempts from sources and constructs
+    a RetryStrategyOptions object. This allows the retry strategy to be
+    configured from multiple sources. Example: retry_mode from config file and
+    max_attempts from environment variables.
+
+    :param resolver: The config resolver to use for resolution
+
+    :returns: Tuple of (RetryStrategyOptions or None, source_name or None).
+        Returns (None, None) if neither retry_mode nor max_attempts is set.
+
+        For mixed sources, the source string includes both component sources:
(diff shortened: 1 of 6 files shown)


packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns (None, None) if neither retry_mode nor max_attempts is set.
+
+        For mixed sources, the source string includes both component sources:
+        "retry_mode=environment, max_attempts=config_file"
+    """
+    # Get retry_mode
+    retry_mode, mode_source = resolver.get("retry_mode")
+
+    # Get max_attempts
+    max_attempts, attempts_source = resolver.get("max_attempts")
```
SamRemis: What does the code path look like when a customer sets RetryStrategyOptions directly on the config object's init method? 

Do we properly store the source then?

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    if not isinstance(region_name, str):
+        raise ConfigValidationError(
+            "region",
+            region_name,
+            f"Region must be a string, got {type(region_name).__name__}",
+            source,
+        )
+
+    pattern = r"^(?![0-9]+$)(?!-)[a-zA-Z0-9-]{,63}(?<!-)$"
```
SamRemis: The regex definition and the match aren't specific to region- they're specific to any configurations that can go in the host.  Maybe we rename this validator is_valid_host or similar and make it more reusable?

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,116 @@
...
+
+    :param value: The retry strategy value to validate (None is allowed and returns None)
+    :param source: The source that provided this value (for error messages)
+
+    :returns: The validated retry strategy (RetryStrategy or RetryStrategyOptions)
+
+    :raises: ConfigValidationError: If the value is not a valid retry strategy type
+    """
+    # Allow RetryStrategy instances
+    if isinstance(value, RetryStrategy):
```
SamRemis: You can combine this and the below check with `    if isinstance(value, RetryStrategy | RetryStrategyOptions):`

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,116 @@
...
+            "retry_mode",
+            retry_mode,
+            f"Retry mode must be one of {RetryStrategyType}, got {retry_mode}",
+            source,
+        )
+
+    return retry_mode
+
+
+def validate_retry_strategy(value: Any, source: str | None = None) -> Any:
```
SamRemis: Any is a code smell in general; this should return RetryStrategy | RetryStrategyOptions, right?

packages/smithy-core/src/smithy_core/config/property.py
```diff
@@ -0,0 +1,99 @@
...
+        :param validator: Optional validation function that takes (value, source)
+                      and returns validated value or raises an exception
+        :param resolver_func: Optional custom resolver function for complex resolution.
+                  Takes a ConfigResolver and returns (value, source) tuple.
+        """
+        self.key = key
+        self.validator = validator
+        self.resolver_func = resolver_func
+        # Cache attribute name in instance __dict__ (e.g., "_cache_region")
+        self.cache_attr = f"_cache_{key}"
```
SamRemis: Consider adding a default here so the default is coupled with the config property instead of the DefaultSource

packages/smithy-core/src/smithy_core/config/property.py
```diff
@@ -0,0 +1,99 @@
...
+            ]  # Return value from tuple (value, source) if already cached
+
+        # If not cached, use a resolver to go through the sources to get (value, source)
+        # For complex config resolutions, use a custom resolver function to resolve values
+        if self.resolver_func:
+            value, source = self.resolver_func(obj._resolver)
+        else:
+            value, source = obj._resolver.get(self.key)
+
+        if self.validator and value is not None:
```
SamRemis: do we want to allow None values for all configurations?  This will break for something required like region

packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
```diff
@@ -0,0 +1,52 @@
...
+        return (None, None)
+
+    if retry_mode is not None:
+        retry_mode = validate_retry_mode(retry_mode, mode_source)
+        retry_mode = cast(RetryStrategyType, retry_mode)
+
+    # Construct options with defaults
+    options = RetryStrategyOptions(
+        retry_mode=retry_mode or "standard",
+        max_attempts=int(max_attempts) if max_attempts else None,
```
alexgromero: Should we add validation for `max_attempts` before converting to int? If it comes from a config source as a string, calling `int(max_attempts)` directly could raise an unhandled `ValueError` for invalid values like "abc".

packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
```diff
@@ -0,0 +1,51 @@
...
+        For mixed sources, the source name includes both component sources:
+        "retry_mode=environment, max_attempts=config_file"
+    """
+    # Get retry_mode
+    retry_mode, mode_source = resolver.get("retry_mode")
+
+    # Get max_attempts
+    max_attempts, attempts_source = resolver.get("max_attempts")
+
+    if retry_mode is None or max_attempts is None:
```
SamRemis: This is ignoring one of these values if both aren't set.  It's valid for a customer to only set `retry_mode` and not `max_attempts` and expect that to be respected.

packages/smithy-core/src/smithy_core/config/property.py
```diff
@@ -0,0 +1,106 @@
...
+        :param objtype: The Config class
+
+        :returns: The resolved and validated config value
+        """
+        # If accessed on class instead of instance, return descriptor itself
+        if obj is None:
+            return self
+
+        cached = getattr(obj, self.cache_attr, None)
+        if cached is not None:
```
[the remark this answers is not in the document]
SamRemis: Right- that's my mistake.  I was thinking getattr() would return `None` when the object resolved to None, but in reality it will resolve `(None, "defaulted")` or similar

packages/smithy-aws-core/src/smithy_aws_core/config/custom_resolvers.py
```diff
@@ -0,0 +1,50 @@
...
+        return None, None
+
+    if retry_mode is not None:
+        retry_mode = validate_retry_mode(retry_mode, mode_source)
+
+    if max_attempts is not None:
+        max_attempts = validate_max_attempts(max_attempts, attempts_source)
+
+    options = RetryStrategyOptions(
+        retry_mode=retry_mode or "standard",  # type: ignore
```
SamRemis:
```suggestion
        retry_mode=retry_mode
```

packages/smithy-core/src/smithy_core/config/property.py
```diff
@@ -0,0 +1,106 @@
...
+            return cached[
+                0
+            ]  # Return value from tuple (value, source) if already cached
+
+        # If not cached, use a resolver to go through the sources to get (value, source)
+        # For complex config resolutions, use a custom resolver function to resolve values
+        if self.resolver_func:
+            value, source = self.resolver_func(obj._resolver)
+        else:
+            value, source = obj._resolver.get(self.key)
```
SamRemis: As another follow up, we should expose each source from the resolver so they can be called directly, for example for use in environment only configs like the bedrock token

SamRemis (approved):

PR #645, opened 2026-02-26
Add protocol support for AWS JSON-RPC

jonathan343:
## Overview

This PR adds client-side `awsJson1_0` and `awsJson1_1` support, including protocol generators, test projections, runtime request/response handling, modeled error resolution, and retry-after propagation. It also adds JSON serde support for `NaN`, `Infinity`, and `-Infinity`. Event streams, endpoint host prefixes, request compression, and remaining default-value cases are deferred.

> [!NOTE]
> This PR does not implement event streaming support for JSON-RPC. Of the 149 AWS services that use this protocol, only 2 of them use event streaming. This will be implemented as a follow up along with proper event streaming protocol tests.

## Testing

I was able to generate a client for Amazon SQS from its [smithy service model](https://github.com/aws/api-models-aws/blob/main/models/sqs/service/2012-11-05/sqs-2012-11-05.json) and make successfully calls:

### SQS Test Script
```python
import asyncio

from aws_sdk_sqs.client import AsyncSQSClient, ListQueuesInput
from aws_sdk_sqs.config import Config

async def main():
    client = AsyncSQSClient(
        config=Config(
            endpoint_uri="https://sqs.us-west-2.amazonaws.com",
            region="us-west-2",
        )
    )
    response = await client.list_queues(
        ListQueuesInput()
    )
    print(response)

asyncio.run(main())
```

**Output:**

```
(smithy-python) $ python test.py
/Users/gytndd/dev/GitHub/smithy-python/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py:794: AWSSDKWarning: Payload signing is enabled. This may result in decreased performance for large request bodies.
  warnings.warn(
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x10b598110>
ListQueuesOutput(queue_urls=['https://sqs.us-west-2.amazonaws.com/<REDACTED>.fifo', 'https://sqs.us-west-2.amazonaws.com/<REDACTED>.fifo'], next_token=None)
```

---


JordonPhillips (changes requested):

JordonPhillips (dismissed):

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -115,71 +222,315 @@ def payload_codec(self) -> Codec:
...
+
+    def _is_success(
+        self,
+        operation: APIOperation[Any, Any],
+        context: TypedProperties,
+        response: HTTPResponse,
+    ) -> bool:
+        return 200 <= response.status < 300
+
+    def _resolve_host_prefix(
```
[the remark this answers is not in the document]
jonathan343: Removed. I'll defer to when we have endpoints actually working so the integration is clean.

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -115,71 +220,277 @@ def payload_codec(self) -> Codec:
...
+        # normalize empty bodies to {} for structured errors.
+        error_id = self.error_identifier.identify(
+            operation=operation, response=response
         )
+
+        if error_id is None and self._matches_content_type(response):
+            if isinstance(response_body, bytearray):
+                response_body = bytes(response_body)
+            deserializer = self.payload_codec.create_deserializer(source=response_body)
+            document = deserializer.read_document(schema=DOCUMENT)
```
SamRemis: [Genuine question]

What happens if they sent us a content type but invalid json, for instance in the form of an incomplete payload?  Is this going to safely pass and fall back to the next step, or will it raise a clear error?

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -97,3 +114,319 @@ def test_aws_json_document_discriminator(
...
+
+    assert request.method == "POST"
+    assert request.destination.path == "/"
+    assert request.fields["content-type"].as_string() == "application/x-amz-json-1.1"
+    assert request.fields["x-amz-target"].as_string() == "JsonService.EmptyOperation"
+    assert request.fields["content-length"].as_string() == "2"
+    assert await request.consume_body_async() == b"{}"
+
+
+@pytest.mark.asyncio
```
SamRemis: I got these tests to pass locally without this annotation, likely due to https://github.com/smithy-lang/smithy-python/blob/2e3bb3376826ff1ee9c2f3f1a54c9b14db99a32c/pyproject.toml#L48

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -115,71 +220,277 @@ def payload_codec(self) -> Codec:
...
+        response: HTTPResponse,
+        response_body: SyncStreamingBlob,
+        error_registry: TypeRegistry,
+        context: TypedProperties,
+    ) -> CallError:
+        # Keeps HttpBindingClientProtocol._create_error's error ID resolution,
+        # but awsJson operations do not use per-operation HTTP bindings. Once
+        # the error shape is resolved, deserialize it directly from JSON and
+        # normalize empty bodies to {} for structured errors.
+        error_id = self.error_identifier.identify(
```
SamRemis: [Non-blocking suggestion]

Bit of an odd logical flow here:

- (line 416) error identifier is an instance of `AWSErrorIdentifier` and only looks in the header
- (lines 420-434) If the first identifier fails, then we parse the body as JSON and try to grab it from the expected locations there; if it's found, we set error_id to that 
- (line 436-441) If we _have_ found an error id, we try to resolve it again... with the already resolved value from 428 if we entered that block

Can we update this code path to remove that double resolve?  So it would first try to identify the header, then the body discriminator if it wasn't in the header; in another block later, we would call _resolve_error_id instead of twice.

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -87,11 +111,92 @@ def discriminator(self) -> ShapeID:
         return parsed
 
 
-class RestJsonClientProtocol(HttpBindingClientProtocol):
+class _EventStreamClientProtocolMixin:
```
SamRemis: [Genuine question]

What happens when you try to generate an AwsJson client with an event stream?  Does codegen fail?  Or does it pass and then fail when you try to call the operation with the NotImplementedError?

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -97,3 +114,319 @@ def test_aws_json_document_discriminator(
...
+    protocol = AwsJson11ClientProtocol(
+        Schema(id=ShapeID("com.test#JsonService"), shape_type=ShapeType.SERVICE)
+    )
+    response = HTTPResponse(
+        status=500,
+        fields=tuples_to_fields(
+            [("content-type", "application/x-amz-json-1.1; charset=utf-8")]
+        ),
+    )
+    assert getattr(protocol, "_matches_content_type")(response)
```
SamRemis: Should we assert the negative here as well?  Something doesn't match the expected content type?

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -97,3 +114,319 @@ def test_aws_json_document_discriminator(
...
+    )
+    operation = _mock_operation(_operation_schema("FailingOperation"))
+    response = HTTPResponse(
+        status=400,
+        reason="Bad Request",
+        fields=tuples_to_fields([("x-amzn-errortype", "com.other#OtherNsError")]),
+        body=b"",
+    )
+
+    error = await getattr(protocol, "_create_error")(
```
SamRemis: We are doing a lot of these tests where we are testing the private _create_error attribute with calling getattr; can we avoid that and just pass in an error response and make sure it gets handled properly?

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -241,6 +246,199 @@ def create_event_receiver[
...
+            # make sure the error shape is derived from modeled exception
+            if not issubclass(error_shape, ModeledError):
+                raise ExpectationNotMetError(
+                    f"Modeled errors must be derived from 'ModeledError', "
+                    f"but got {error_shape}"
+                )
+
+            body = response_body if len(response_body) > 0 else b"{}"
+            deserializer = self.payload_codec.create_deserializer(body)
+            return error_shape.deserialize(deserializer)
```
Alan4506:
```suggestion
            modeled_error = error_shape.deserialize(deserializer)
            if retry_after is not None:
                modeled_error.retry_after = retry_after
            return modeled_error
```

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -241,6 +246,199 @@ def create_event_receiver[
...
+        response: HTTPResponse,
+        response_body: bytes,
+        error_registry: TypeRegistry,
+        context: TypedProperties,
+    ) -> CallError:
+        error_id = self.error_identifier.identify(
+            operation=operation, response=response
+        )
+        if error_id is not None and error_id not in error_registry:
+            error_id = self._resolve_error_id(operation=operation, error_id=error_id)
```
Alan4506, line 363:
```suggestion
            error_id = self._resolve_error_id(operation=operation, error_id=error_id)
            
        retry_after = parse_retry_after(response)
```

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -241,6 +246,199 @@ def create_event_receiver[
...
+        is_timeout = response.status == 408
+        is_throttle = response.status == 429
+        fault = "client" if response.status < 500 else "server"
+
+        return CallError(
+            message=message,
+            fault=fault,
+            is_throttling_error=is_throttle,
+            is_timeout_error=is_timeout,
+            is_retry_safe=is_throttle or is_timeout or None,
```
Alan4506, line 418:
```suggestion
            is_retry_safe=is_throttle or is_timeout or None,
            retry_after=retry_after,
```

Alan4506 (changes requested):
Thanks @jonathan343! The PR status looks good to me overall, but it conflicts with the `develop` branch after we merged https://github.com/smithy-lang/smithy-python/pull/733 last Friday. Please make a new commit to resolve the conflicts - just some import conflicts in 2 files and we should keep changes from both branches. 

In the new commit, you may also need to update `packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py` - in https://github.com/smithy-lang/smithy-python/pull/733, we started honoring the `x-amz-retry-after` header when building errors from a response. Each protocol parses the header and attaches it to both the modeled error and the generic CallError fallback, so the retry strategy can use the server-provided backoff. Both `restJson` and `awsQuery` have implemented this on the develop branch. Now we need to apply the same thing for `awsJson`. I've suggested changes on that file and you can see comments below.

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
jonathan343: Addressed in 65f695c1e759f75a9a48728340bdc61161825906

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
jonathan343: Addressed in 65f695c1e759f75a9a48728340bdc61161825906

packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
jonathan343: Addressed in 65f695c1e759f75a9a48728340bdc61161825906

Alan4506 (approved):
LGTM!

jonathan343:
Merging since these changes are functional and there's more urgency on them. Any improvements are a 2-way door and can be made in the future.

PR #646, opened 2026-02-26
Update the codegen files that generate config.py to support config resolution.

ubaskota:
*Issue #, if available:*

*Description of changes:*
This PR updates the Python client code generator to support descriptor-based configuration resolution with source tracking. The changes enable generated `config.py` files to use the descriptor pattern for resolution of configuration properties from multiple sources with proper precedence and caching.

*Tests:*
Verified that the newly generated code in config.py has all the expected changes as shown below:

- Verified that it has all the required import statements:
- Verified that the changes for `region` and `retry_strategy` are generated:
```
    region = ConfigProperty(
        "region", validator=validate_host
        )
    retry_strategy = ConfigProperty(
        "retry_strategy",
        validator=validate_retry_strategy,
        resolver_func=resolve_retry_strategy,
        default_value=RetryStrategyOptions(retry_mode="standard", max_attempts=3),
    )
```

- Verified that the changes to handle instance values are generated.

```
       descriptor_keys = ["region", "retry_strategy"]
        for key in descriptor_keys:
            value = locals().get(key)
            if value is not None:
                setattr(self, key, value)
```
- Verified that the function to get the source of a config value is generated.
```
    def get_source(self, key: str) -> str | None:
       .......
        cached = self.__dict__.get(f"_cache_{key}")
        return cached[1] if cached else None
```

All tests passed successfully, confirming that the descriptor-based resolution works as expected . Validated configuration resolution using the generated code with the following scenarios:
```
**Region resolution**
- region uses the value from the instance 
- region uses the value set in environment variable 
- Will throw an error if the value for region is not found (It won't have a default)

**Retry Strategy resolution**
- retry_strategy uses the default value when not available in the config sources 
- retry_strategy uses the value set in instance 
- retry_strategy uses the value set in environment 

**Other**
- Source name for complex resolution is as expected
- Config value set in instance overrides the values set in environment 
```


codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -358,21 +404,65 @@ def __init__(
...
                     ? "$L: $T | None"
                     : "$L: $T";
             writer.write(formatString, property.name(), property.type());
             writer.writeDocs(property.documentation(), context);
             writer.write("");
         }
+
+        // Add _resolver declaration
+        writer.write("_resolver: ConfigResolver");
+        writer.writeDocs("Shared resolver for all Config instances.", context);
```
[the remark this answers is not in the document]
ubaskota: Yeah, its not shared. I realized it after publishing this PR and decided to wait until the first round of review is completed.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -358,21 +373,67 @@ def __init__(
...
                 serviceId,
+                writer.consumer(w -> writeDescriptorDeclarations(w, finalProperties)),
                 writer.consumer(w -> writePropertyDeclarations(w, finalProperties)),
                 writer.consumer(w -> writeInitParams(w, finalProperties)),
                 writer.consumer(w -> initializeProperties(w, finalProperties)));
         writer.popState();
     }
 
+    // Write descriptor declarations for properties using ConfigProperty descriptor
+    private void writeDescriptorDeclarations(PythonWriter writer, Collection<ConfigProperty> properties) {
```
SamRemis: If we include the type annotations like this:
```
region: str = ConfigProperty(
    "region", validator=validate_region, default_value="us-east-1"
)
```
Then we will give the type hint for IDEs.  We could also use a pyi file for this.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -335,6 +319,35 @@ private void generateConfig(GenerationContext context, PythonWriter writer) {
         }
 
         var finalProperties = List.copyOf(properties);
+
+        writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
+        writer.addImport("smithy_core.config.property", "ConfigProperty");
+        writer.addImport("smithy_core.config.resolver", "ConfigResolver");
```
SamRemis: This is adding an AWS resolver to generic clients that never gets used, please make sure this only goes in AWS clients

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsConfiguration.java
```diff
@@ -20,5 +20,45 @@ private AwsConfiguration() {}
...
             .documentation(" The AWS region to connect to. The configured region is used to "
                     + "determine the service endpoint.")
+            .nullable(false)
+            .useDescriptor(true)
+            .validator(Symbol.builder()
+                    .name("validate_region")
+                    .namespace("smithy_aws_core.config.validators", ".")
+                    .addDependency(AwsPythonDependency.SMITHY_AWS_CORE)
+                    .build())
+            .defaultValue("'us-east-1'")
```
jonathan343: Where is this requirement coming from? I think setting a default region on behalf of customers isn't a good idea.

jonathan343 (commented):
I'll continue reviewing but I added two comments for things that immediately stand out:
- We lost docstrings for the `retry_strategy` and `region` config options.
- I don't think we should be setting `us-east-1` as the default region. We can continue discussion in the thread for that comment.

We should also make sure the test failures are addressed in a subsequent commit.

jonathan343 (commented):
Thanks for addressing my initial concerns. Left a few more. I'm also seeing weird behavior when no region is set:

```python
>>> config = Config()
>>> config.region
  File "<stdin>", line 1, in <module>
  File "/Users/gytndd/dev/GitHub/temp/smithy-python/packages/smithy-core/src/smithy_core/config/property.py", line 84, in __get__
    value = self.validator(value, source)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gytndd/dev/GitHub/temp/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/validators.py", line 38, in validate_region
    if not re.match(pattern, region):
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gytndd/.local/share/uv/python/cpython-3.12.12-macos-aarch64-none/lib/python3.12/re/__init__.py", line 167, in match
    return _compile(pattern, flags).match(string)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```

This is something we need addressed in this PR. We should be throwing a config error when this gets resolved

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -359,31 +397,134 @@ def __init__(
...
+                .filter(ConfigProperty::useDescriptor)
+                .toList();
+
+        if (!descriptorProperties.isEmpty()) {
+            writer.write("# Set instance values for descriptor properties");
+            writer.write("self._resolver = ConfigResolver(sources=[EnvironmentSource()])");
+            writer.write("");
+
+            writer.write("# Only set if provided (not None) to allow resolution from sources");
+            writer.write("for key in self.__class__._descriptors.keys():");
```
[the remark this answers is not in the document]
ubaskota: We'd need to create a new dictionary if we were to use the approach you recommended. `locals()` automatically stays in sync with the value set for descriptor keys.

jonathan343 (approved):

PR #648, opened 2026-02-28
Add tests for config resolution mechanism

ubaskota:
*Issue #, if available:*

*Description of changes:*
- Add unit tests for invalid sources.

After discussing with the team, we've decided to hold off on adding functional tests until we have a generated test client. The functional tests referenced below have been removed and will be reintroduced once the test client is ready.

```
Add functional tests for the config resolution system to verify the flow. The added tests cover the following:

- Default, environment, and instance value resolution
- Custom resolvers and validation
```


packages/smithy-aws-core/tests/functional/test_config_resolution.py
```diff
@@ -0,0 +1,239 @@
...
+
+    return type("TestConfig", (BaseTestConfig,), class_dict)
+
+
+class TestConfigResolution:
+    """Functional tests for complete config resolution flow."""
+
+    def test_environment_var_resolution(self, monkeypatch: MonkeyPatch) -> None:
+        TestConfig = make_config_class(
+            {"region": ConfigProperty("region", default_value="us-east-1")}
```
arandito: From #646, it looks like we are not setting a default value for region. We should not be including that here since it doesn't match our code generator's behavior.

Looks like we are setting the default in line 69, 108, and 128.

packages/smithy-aws-core/tests/functional/test_config_resolution.py
```diff
@@ -0,0 +1,239 @@
...
+                ),
+                "standard",
+            ),
+        ],
+    )
+    def test_uses_default_when_no_sources(
+        self, property_name: str, property_config: ConfigProperty, expected_value: str
+    ) -> None:
+        TestConfig = make_config_class({property_name: property_config})
+        config = TestConfig(sources=[])
```
arandito: Is `sources` supposed to be part of the Config constructor?

It is defined as a property of `ConfigResolver` not `Config`. This mixes both classes and is confusing. I'm also not sure we ever generate a Config object where sources is empty.

packages/smithy-aws-core/tests/functional/test_config_resolution.py
```diff
@@ -0,0 +1,239 @@
...
+        config = TestConfig()
+
+        retry_strategy = config.retry_strategy
+        assert isinstance(retry_strategy, RetryStrategyOptions)
+        assert retry_strategy.retry_mode == "standard"
+        # None for max_attempts means the RetryStrategy will use its
+        # own default max_attempts value for the set retry_mode
+        assert retry_strategy.max_attempts is None
+        source = config.get_source("retry_strategy")
+        assert source == "default"
```
arandito: This behavior is a bit weird to me. I would expect `retry_strategy.max_attempts` to return the default value instead of None. We now have hidden behavior that doesn't make it clear that `retry_strategy.max_attempts=None` means that we are using the default value. I would also expect `source` to be "retry_mode=default max_attempts=default".

I may be missing a decision that was made that required this behvaior, so feel free to explain it in case I'm missing something.

arandito (commented):
Thanks Ujjwal! I left a few comments. My main concern is that we are creating test Config objects that are inconsistent with what we are code generating. But I am curious with what other reviewers think as I'm not too familiar with our functional testing approach for our runtime packages. Let me know if you have any questions!

packages/smithy-aws-core/tests/functional/test_config_resolution.py
ubaskota: When neither `retry_mode` nor `max_attempts` are found in any source, `resolve_retry_strategy` returns `(None, None)` and the descriptor falls back to the default value object: `RetryStrategyOptions(retry_mode="standard")`. Since `max_attempts` isn't set on this default object, it's None at the config layer. The actual default value (max_attempts=3) gets applied later when [create_retry_strategy](https://github.com/smithy-lang/smithy-python/blob/5c439887f1151f3b3044bd7c056ed919558b78c8/packages/smithy-core/src/smithy_core/retries.py#L57) is called. That is why the comment. 

And similarly, the source is only `default` because it used the whole default value object instead of getting individual components from different sources.

PR #653, opened 2026-03-11
Add support for sdk_ua_app_id.

ubaskota:
*Issue #, if available:*

*Description of changes:*
Add codegen support for a new config property: `sdk_ua_app_id`, along with a validator. For AWS SDKs, these are generated as descriptor-based properties with config resolver and default values. 

*Testing*
- Added a unit tests to verify default value behavior: sdk_ua_app_id 
- All existing tests pass.
- Manually tested the generated config from the client side and verified that all tests pass.

### DIFF
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsConfiguration.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsConfiguration.java
@@ -60,4 +60,30 @@ public final class AwsConfiguration {
                     .build())
             .defaultValue("RetryStrategyOptions(retry_mode=\"standard\")")
             .build();
+
+    public static final ConfigProperty USER_AGENT_EXTRA = ConfigProperty.builder()
+            .name("user_agent_extra")
+            .type(Symbol.builder().name("str").build())
+            .documentation("Additional suffix to be appended to the User-Agent header.")
+            .nullable(false)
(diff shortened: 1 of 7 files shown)


codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsConfiguration.java
```diff
@@ -60,4 +60,30 @@ private AwsConfiguration() {}
                     .build())
             .defaultValue("RetryStrategyOptions(retry_mode=\"standard\")")
             .build();
+
+    public static final ConfigProperty USER_AGENT_EXTRA = ConfigProperty.builder()
+            .name("user_agent_extra")
```
jonathan343: Why was this config option added? This config option isn't standardized across all SDKs ([source](https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html)). By adding it here we're now reading `AWS_USER_AGENT_EXTRA` from the environment which we shouldn't be doing.

jonathan343 (commented):
Had a few comments. I also noticed there are no validators for any of these options. I know at least the user agent app ID will need it:

> Valid values: String with maximum length of 50. Letters, numbers and the following special characters are allowed: !,#,$,%,&,',*,+,-,.,^,_,`,|,~.

SamRemis (approved):

PR #655, opened 2026-03-11
Add smithy-xml package

arandito:
### Description
This PR adds an initial implementation of the new `smithy-xml` runtime package, which provides generic XML serialization and deserialization support for Smithy clients and servers.

This PR primarily implements the XML deserialization support needed by the `awsQuery` protocol. The serializer is intentionally omitted for now as `awsQuery` does not use XML request serialization. It will be introduced later alongside the `restXml` protocol, where XML request serialization is required and can be validated more thoroughly.

The implementation follows the XML binding rules defined in the Smithy [awsQuery](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html) and [restXml](https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html) protocol specifications.

> [!IMPORTANT]
> This implementation follows the Smithy Python [Protocol Serialization and Deserialization design](https://github.com/smithy-lang/smithy-python/blob/develop/designs/serialization.md). I strongly recommend reviewing this document first to better understand the implementation.

### Testing
Added round trip serde test cases that follow the same testing strategy as `smithy-json`.


packages/smithy-xml/src/smithy_xml/settings.py
```diff
@@ -0,0 +1,16 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from dataclasses import dataclass
+
+from smithy_core.types import TimestampFormat
+
+
+@dataclass(frozen=True)
+class XMLSettings:
```
jonathan343: The smithy docs say we should be supporting the `[timestampFormat](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait)` trait. I see that the `JSONSettings` class has this. Is there a reason why you didn't add this. Or is that just a miss?

jonathan343 (commented):
I'll need to go over this a second time, but dropping some quick comments now since I might not be able to get back to this soon:

jonathan343 (dismissed):

JordonPhillips (commented):
Good work! I bet there will be a few issues uncovered when we plug in the protocol tests, but that's always gonna be true.

jonathan343 (approved):
🚢

PR #659, opened 2026-03-16
Refactor and move config resolution codegen logic from generic layer to AWS layer

ubaskota:
*Issue #, if available:*

*Description of changes:*
Moved AWS-specific config resolution codegen (descriptors, validators, custom resolvers, default values) out of the generic ConfigProperty and ConfigGenerator into the AWS codegen layer. Introduced `AwsConfigResolutionIntegration` as a CodeInterceptor that transforms the generated Config class to inject descriptor support, resolver initialization, and source tracking. Each integration now owns its own property metadata (e.g., `SDK_UA_APP_ID_METADATA` lives in `AwsUserAgentIntegration`). The generic codegen layer (`codegen/core/`) has zero diff from develop.

*Testing*
- Verified that the config file is generated with descriptor, resolver and validators.
- Verified that the config file for the generic sdk is unchanged.
- Ran tests using the the generated config file and verified that all tests pass. 


SamRemis (approved):
Thanks Ujjwal :) looks good

PR #670, opened 2026-03-26
Remove enum usage for header/trailer in smithy-http and aws-sdk-signers

ubaskota:
*Issue #, if available:*

*Description of changes:*
`FieldPosition` was an enum, which meant any library implementing the Field protocol had to import `smithy-http` just to reference the `enum` values. Switch to `Literal["header", "trailer"]` so that implementations can pass equality checks using plain strings without depending on `smithy-http`. Also change identity checks (is) to equality (==) to ensure correctness with string literals.

*Testing*
- Updated the tests to reflect the changes and verified that all tests pass.

### DIFF
--- a/packages/smithy-http/src/smithy_http/__init__.py
+++ b/packages/smithy-http/src/smithy_http/__init__.py
@@ -24,7 +24,7 @@ class Field(interfaces.Field):
         *,
         name: str,
         values: Iterable[str] | None = None,
-        kind: FieldPosition = FieldPosition.HEADER,
+        kind: FieldPosition = "header",
     ):
         self.name = name
         self.values: list[str] = list(values) if values is not None else []
@@ -79,7 +79,7 @@ class Field(interfaces.Field):
             return False
         return (
             self.name == other.name
-            and self.kind is other.kind
+            and self.kind == other.kind
             and self.values == other.values
         )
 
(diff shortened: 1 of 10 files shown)


packages/smithy-http/src/smithy_http/__init__.py
```diff
@@ -24,7 +24,7 @@ def __init__(
         *,
         name: str,
         values: Iterable[str] | None = None,
-        kind: FieldPosition = FieldPosition.HEADER,
+        kind: FieldPosition = "header",
```
jonathan343: Same suggestion here as with aws-sdk-signers: We should add some runtime validation.

packages/smithy-http/tests/unit/test_fields.py
jonathan343: Same here about the `test_field_invalid_kind` test

jonathan343 (changes requested):
Thanks Ujjwal!

I left a few comments. Let me know if you have any questions!

With more adoption, I'd normally want to have a transition period where we continue to support enum usage, but I searched through GitHub and don't see anyone using `Field` or `FieldPosition` classes directly. This approach seems fine to me!

jonathan343 (approved):

PR #671, opened 2026-03-27
Refactor RetryStrategy to be async

JordonPhillips:
This refactors RetryStrategy to be async. This is necessary because a strategy may need to use a synchronization primitive or may need to internally wait.

For example, AWS retry strategies may need to wait internally even if a retry would not be permitted if the operation is a long-poll operation.


packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -0,0 +1,227 @@
...
+                f"retry_strategy must be RetryStrategy, RetryStrategyOptions, or None, "
+                f"got {type(retry_strategy).__name__}"
+            )
+        return self._create_retry_strategy(
+            retry_strategy.retry_mode, retry_strategy.max_attempts
+        )
+
+    @lru_cache
+    def _create_retry_strategy(
+        self, retry_mode: RetryStrategyType, max_attempts: int | None
```
ubaskota, line 50: `_create_retry_strategy` is synchronous with `@lru_cache`. Is this intentional? It's just creating and returning retry strategy today, but do you see a case where this would need to be async?

packages/smithy-core/src/smithy_core/retries.py
ubaskota, line 218: Is leaving `threading.Lock` intentional? If so, do you think it's worth adding a short comment explaining the choice? And do you see a scenario where we may need to make this async?

ubaskota (commented):
Thanks for the PR. It looks good overall. I have some clarifying questions inline where I could add them, plus two more below that didn't fit in any specific part:  

- The PR description mentions long-poll operations as motivation, and notes that a strategy "may need to wait internally." The current implementations only do computation (backoff math, quota checks) and no I/O operations. What does "waiting internally" mean concretely here? 

- Should `designs/retries.md` file be updated with the change? Adding it here since GitHub doesn't let me comment on unchanged files.

JordonPhillips:
> The PR description mentions long-poll operations as motivation, and notes that a strategy "may need to wait internally." The current implementations only do computation (backoff math, quota checks) and no I/O operations. What does "waiting internally" mean concretely here?

The proposed changes to the AWS standard retry mode require a delay even if the token bucket is empty. That is, you delay and then return the empty bucket error.

> Should designs/retries.md file be updated with the change? Adding it here since GitHub doesn't let me comment on unchanged files.

good point

packages/smithy-core/src/smithy_core/aio/retries.py
JordonPhillips: Anything is possible I suppose. This is an internal method, so it could be changed if needed. The calling method is already async.

packages/smithy-core/src/smithy_core/retries.py
JordonPhillips: This is existing code. At one point I had a thing where I completely rewrote all of this but it got too unwieldy and was missing the forest for the trees.

This is a thread lock to make sure it's thread safe. Since this is the kind of thing you might want to share across threads, that makes sense, even in an async context. You do stop the world, but in this case you aren't doing so for long.

An async lock actually would not make any sense in a token bucket class like this. In a threading world, you have no idea who's touching your memory and when. In an async world, you know exactly when you're potentially passing priority to another coroutine. None of the methods in this would actually pass priority if converted to async.

That could change if you add a wait-for-capacity mode, however. For example:


```python
class AsyncTokenBucket:
    def __init__(self, capacity: int = 500) -> None:
        self._capacity = capacity
        self._max_capacity = capacity
        self._capacity_condition = asyncio.Condition()

    async def drain(self, amount: int, wait: bool = False) -> bool:
        assert amount > 0

        # Immediately check the capacity without passing priority. If there's enough,
        # go ahead and drain it.
        if self._capacity >= amount:
            self._capacity -= amount
            return True

        # If there's not enough capacity, wait for there to be capacity if waiting is
        # enabled.
        if wait:
            async with self._capacity_condition:
                # Wait until there's enough capacity to drain the requested amount.
                await self._capacity_condition.wait_for(
                    lambda: self._capacity >= amount
                )
                self._capacity -= amount
                return True

        return False

    async def fill(self, amount: int) -> None:
        assert amount > 0
        self._capacity = max(self._max_capacity, self._capacity + amount)

        # Notify any waiting coroutines that there's more capacity.
        async with self._capacity_condition:
            self._capacity_condition.notify_all()
```

This class IS NOT thread safe. But it does showcase an instance where an async lock makes sense. When there's not enough capacity in the bucket, priority is passed to the next coroutine. The waiting coroutine won't become active again until `notify_all` is checked. When it becomes active, it'll check to see if the condition is met. If it is, it'll drain its capacity and return. If it isn't, it'll become inactive again as it waits for another notify.

Note the start of the `drain` method in particular. It immediately checks and drains the capacity if there is enough. This is safe because it's async and we know that no other coroutine is running. In an equivalent threading method, the capacity could have changed between the time you check it and the time you drain it, thus requiring a lock.

Now if you remove the waiting capability, then the condition can go away and you no longer need async methods. It still won't be thread safe unless thread locks are added.

designs/retries.md
```diff
@@ -144,7 +144,7 @@ example:
 
 ```python
 try:
-    retry_token = retry_strategy.acquire_initial_retry_token()
+    retry_token = await retry_strategy.acquire_initial_retry_token()
 except RetryError:
     transport_response = transport_client.send(serialized_request)
```
jonathan343, line 149: nit - Looks like there are at least two additional lines missing `await`. The other `transport_client.send` below needs this too. Not sure if deserialize also needs one. Our [real code](https://github.com/smithy-lang/smithy-python/blob/ef6efcd1f9a3d093a06843759009d9edb828442e/packages/smithy-core/src/smithy_core/aio/client.py#L499) in `smithy_core` awaits deserialization. 

```suggestion
    transport_response = await transport_client.send(serialized_request)
```

packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -0,0 +1,227 @@
...
+    ExponentialBackoffJitterType,
+    ExponentialRetryBackoffStrategy,
+    RetryStrategyOptions,
+    SimpleRetryToken,
+    StandardRetryQuota,
+    StandardRetryToken,
+)
+from .interfaces.retries import RetryStrategy
+
+RetryStrategyType = Literal["simple", "standard"]
```
jonathan343, line 18: We should import this via `from ..retries import RetryStrategyType` instead of redefining

jonathan343 (approved):
Thanks Jordon! This PR looks good to me as well. I have some very minor nits. Approving since I'd be okay without them, but they're small enough that it should be quick to get them in.

jonathan343:
I'm going to merge and follow up with a minor clean up PR.
Thanks for opening this @JordonPhillips, and sorry for the slow turnaround on reviewing.

PR #673, opened 2026-03-30
Merge feat/add-query-protocol branch into develop

arandito:
## Description

This PR merges the `feat/add-query-protocol` feature branch into `develop`, adding AWS Query protocol support and generic XML deserialization to smithy-python.

**Related PRs in `feat/add-query-protocol`:**
  - #655
  - #663

In addition to the above, this PR includes:
  - `ca5b8c92` — Updates the smithy-xml initial version to `0.0.0` to prepare for release

## Testing
  - Ran smithy protocol tests and confirmed they pass for the newly added Query protocol support.
  - Generated an STS client from its Smithy service model and successfully made live service calls.

#### STS test script
```python
import asyncio

from aws_sdk_sts.client import STSClient
from aws_sdk_sts.config import Config
from aws_sdk_sts.models import GetCallerIdentityInput


async def main():
    client = STSClient(
        config=Config(
            endpoint_uri="https://sts.us-west-2.amazonaws.com",
            region="us-west-2",
        )
    )
    response = await client.get_caller_identity(GetCallerIdentityInput())

    print(response)

asyncio.run(main())
```
#### Output
```bash
> python test.py
/Users/aearanda/repos/github/smithy-python/packages/aws_sdk_signers/signers.py:794: AWSSDKWarning: Payload signing is enabled. This may result in decreased performance for large request bodies.
  warnings.warn(
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x106caa240>
GetCallerIdentityOutput(user_id='<REDACTED>', account='<REDACTED>', arn='<REDACTED>')
```

### How to Run Protocol Tests
```bash
make install
source .venv/bin/activate
make test-protocols
```


jonathan343 (approved):
Great work @arandito! These changes seem mergable in this state. I think we might end up refactoring some of the work on the `smithy-aws-core` side once we define a more common hierarchy for non-binding aws protocols. But I don't think we should block this on that. 

I was able to run an additional sanity check by code generating the SNS client and making successful calls. The code generated clients currently default to an `aiohttp` based client (since there are no event streaming operations) which works, however, when I swap in the CRT based client, I'm seeing the following error:

```
smithy_core.exceptions.SmithyError: 2056 (AWS_ERROR_HTTP_MISSING_BODY_STREAM): Given the provided headers (ex: Content-Length), a body is expected.
```

This is something for us to look into as a follow up

PR #674, opened 2026-03-30
Fix SigV4 signature for URIs with literal query parameters

Alan4506:
## Problem

When generating a client for Amazon Q Business using smithy-python, the `ChatSync` operation fails with `InvalidSignatureException`. The same request succeeds with boto3 using identical credentials and endpoint.

The root cause is that `ChatSync` has a literal query parameter in its Smithy HTTP trait:
```json
"smithy.api#http": {
    "uri": "/applications/{applicationId}/conversations?sync",
    "method": "POST"
}
```

Link to the model file: https://github.com/aws/api-models-aws/blob/9eafbc359b1c6a7187ac921009b9eef85de71d91/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L2244-L2250

The `?sync` query parameter (no `=` sign, no value) is dropped during SigV4 canonical query string computation. `parse_qsl("sync")` returns `[]` because it ignores keys without values by default. The server includes `sync=` in its canonical query string, so the signatures don't match.

## Fix

Add `keep_blank_values=True` to both sync and async `_format_canonical_query` methods. This makes `parse_qsl("sync", keep_blank_values=True)` return `[('sync', '')]`, which correctly produces `sync=` in the canonical query string.

## Testing

1. Generate a Q Business client using smithy-python
2. Create a Q Business anonymous application and pass the application id to an environment variable:
```
export QBUSINESS_APPLICATION_ID=<application-id>
```
3. Run a test like the following:
```python
async def test_chat_sync():
    client = QBusinessClient(config=Config(
        endpoint_uri="https://qbusiness.us-east-1.api.aws",
        region="us-east-1",
    ))
    response = await client.chat_sync(ChatSyncInput(
        application_id=APPLICATION_ID,
        user_message="Hello",
        client_token=str(uuid.uuid4()),
    ))
```
4. Without this fix: `InvalidSignatureException`
5. With this fix: `200 OK` with valid response

### DIFF
--- a/packages/aws-sdk-signers/.changes/next-release/aws-sdk-signers-bugfix-23a4eca5165f48efb1524b95f8498299.json
+++ b/packages/aws-sdk-signers/.changes/next-release/aws-sdk-signers-bugfix-23a4eca5165f48efb1524b95f8498299.json
@@ -1,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Fixed SigV4 signature computation for URIs with literal query parameters (e.g., ?sync). parse_qsl was silently dropping query keys without values, causing InvalidSignatureException."
+}
--- a/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
+++ b/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
@@ -315,7 +315,7 @@ class SigV4Signer:
         if query is None:
             return ""
 
-        query_params = parse_qsl(qs=query)
+        query_params = parse_qsl(qs=query, keep_blank_values=True)
         query_parts = (
             (quote(string=key, safe=""), quote(string=value, safe=""))
             for key, value in query_params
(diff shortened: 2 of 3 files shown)


jonathan343 (changes requested):
Looks good, thanks @Alan4506!

Can you add some unit tests in `tests/unit/test_signers.py` for this case to prevent any regressions? It would look something like below:

> Note: Both `TestSigV4Signer` and `TestAsyncSigV4Signer` already exist.

```python
class TestSigV4Signer:
    SIGV4_SYNC_SIGNER = SigV4Signer()

    def test_format_canonical_query_keeps_blank_values(self) -> None:
        canonical_query = self.SIGV4_SYNC_SIGNER._format_canonical_query(
            query="foo=bar&baz="
    )

    assert canonical_query == "foo=bar&baz="

class TestAsyncSigV4Signer:
    SIGV4_ASYNC_SIGNER = AsyncSigV4Signer()

    async def test_format_canonical_query_keeps_blank_values(self) -> None:
        canonical_query = await self.SIGV4_ASYNC_SIGNER._format_canonical_query(
            query="foo=bar&baz="
    )

    assert canonical_query == "foo=bar&baz="
```

jonathan343 (approved):
LGTM. Thanks @Alan4506!

PR #680, opened 2026-04-07
Handle unknown event types gracefully instead of crashing

Alan4506:
### Background

The Smithy spec ([streaming.html#client-behavior](https://smithy.io/2.0/spec/streaming.html#client-behavior)) states:

> "Adding new events to an event stream union is considered a backward compatible change; clients SHOULD NOT fail when an unknown event is received."

Currently, in `smithy-aws-event-stream` [`deserializers.py`](https://github.com/smithy-lang/smithy-python/blob/3e54be044faa0d38e75aa37972db7e236fb80d2e/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py#L53), `EventDeserializer.read_struct()` does a hard dictionary lookup (`schema.members[member_name]`) which raises `KeyError` when the service sends an event type not present in the generated model. So we have to add some logic to handle unknown event types.

Additionally, although smithy-python automatically creates unknown variants for streams, these variants only have a `tag` field unlike normal events that have a `value` field. When `AWSEventReceiver.receive()` calls `getattr(result, "value")` without a default, it raises `AttributeError` for unknown variants.

These were discovered during Q Business Chat streaming integration testing. The service sends `intermediateGroupEvent` and `intermediateMessageEvent` events not in any published SDK model, causing the `async for event in output_stream` loop to crash.

**Reference implementations in other Smithy SDKs:**

Both smithy-java and smithy-rs (Rust) already handle unknown event types gracefully:

- **smithy-java** ([`TestEventStream.java`](https://github.com/smithy-lang/smithy-java/blob/main/aws/aws-event-streams/src/test/java/software/amazon/smithy/java/aws/events/model/TestEventStream.java)): Generates a `$Unknown(String memberName)` record for each event stream union. The builder has an `$unknownMember(String memberName)` method that the deserializer calls when it encounters an unrecognized event type. The unknown variant preserves the event type name.

- **smithy-rs** ([`EventStreamUnmarshallerGenerator.kt`](https://github.com/smithy-lang/smithy-rs/blob/main/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt)): The generated unmarshaller has an `_unknown_variant` match arm that returns `Ok(UnmarshalledMessage::Event(Output::Unknown))` for client targets. Unlike Java, the Rust unknown variant does not preserve the event type name.

This PR follows the smithy-java approach of preserving the event type name in the unknown variant's `tag` field, so users can identify which unmodeled event was received (e.g., `ChatOutputStreamUnknown(tag="intermediateGroupEvent")`).

### Description of Changes

1. **smithy-aws-event-stream** ([`deserializers.py`](https://github.com/smithy-lang/smithy-python/blob/3e54be044faa0d38e75aa37972db7e236fb80d2e/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py#L53)): When `member_name` is not in `schema.members`, construct a synthetic `Schema` with `member_index=-1` and call the consumer, which triggers the generated `case _:` branch in [`UnionGenerator.java`](https://github.com/smithy-lang/smithy-python/blob/3e54be044faa0d38e75aa37972db7e236fb80d2e/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java#L169-L170).
2. **smithy-python codegen** ([`UnionGenerator.java`](https://github.com/smithy-lang/smithy-python/blob/3e54be044faa0d38e75aa37972db7e236fb80d2e/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java#L169-L170)): The `case _:` branch (previously just logging) now also constructs the per-union unknown variant with the event type name as `tag`.
3. **smithy-aws-event-stream** ([`aio/__init__.py`](https://github.com/smithy-lang/smithy-python/blob/3e54be044faa0d38e75aa37972db7e236fb80d2e/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py#L135)): Change `getattr(result, "value")` to `getattr(result, "value", None)` to prevent `AttributeError` on unknown variants.

### Testing
Regenerated all existing clients locally and all their integration tests passed. Also, verified with Q Business Chat streaming — unknown events (`intermediateGroupEvent`, `intermediateMessageEvent`) are now handled gracefully instead of crashing. Users can identify unknown events via `isinstance` and access the event type name:

```python
from aws_sdk_qbusiness.models import ChatOutputStreamUnknown

async for event in output_stream:
    if isinstance(event, ChatOutputStreamUnknown):
        print(f"Unknown event: {event.tag}")
```

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
@@ -166,7 +166,6 @@ public final class UnionGenerator implements Runnable {
                         match schema.expect_member_index():
                             ${4C|}
                             case _:
-                                logger.debug("Unexpected member schema: %s", schema)
                                 self._set_result($5L(tag=schema.expect_member_name()))
 
                     def _set_result(self, value: $2T) -> None:
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
@@ -144,8 +144,6 @@ public final class UnionGenerator implements Runnable {
         writer.addImport("smithy_core.deserializers", "ShapeDeserializer");
         writer.addImport("smithy_core.exceptions", "SerializationError");
 
-        // TODO: add in unknown handling
-
         var symbol = symbolProvider.toSymbol(shape);
         var deserializerSymbol = symbol.expectProperty(SymbolProperties.DESERIALIZER);
         var schemaSymbol = symbol.expectProperty(SymbolProperties.SCHEMA);
@@ -169,7 +167,7 @@ public final class UnionGenerator implements Runnable {
                             ${4C|}
                             case _:
                                 logger.debug("Unexpected member schema: %s", schema)
-                                self._set_result($5L(tag=schema.member_name or ""))
+                                self._set_result($5L(tag=schema.expect_member_name()))
 
                     def _set_result(self, value: $2T) -> None:
                         if self._result is not None:
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
@@ -149,6 +149,7 @@ public final class UnionGenerator implements Runnable {
         var symbol = symbolProvider.toSymbol(shape);
         var deserializerSymbol = symbol.expectProperty(SymbolProperties.DESERIALIZER);
         var schemaSymbol = symbol.expectProperty(SymbolProperties.SCHEMA);
+        var unknownSymbol = symbol.expectProperty(SymbolProperties.UNION_UNKNOWN);
         writer.putContext("schema", schemaSymbol);
         writer.write("""
                 class $1L:
@@ -168,6 +169,7 @@ public final class UnionGenerator implements Runnable {
                             ${4C|}
                             case _:
                                 logger.debug("Unexpected member schema: %s", schema)
+                                self._set_result($5L(tag=schema.member_name or ""))
 
                     def _set_result(self, value: $2T) -> None:
                         if self._result is not None:
@@ -177,7 +179,8 @@ public final class UnionGenerator implements Runnable {
                 deserializerSymbol.getName(),
                 symbol,
                 schemaSymbol,
-                writer.consumer(w -> deserializeMembers()));
+                writer.consumer(w -> deserializeMembers()),
+                unknownSymbol.getName());
     }
 
     private void deserializeMembers() {
--- a/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
+++ b/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
@@ -10,7 +10,7 @@ from smithy_core.deserializers import (
     SpecificShapeDeserializer,
 )
 from smithy_core.schemas import Schema
-from smithy_core.shapes import ShapeType
+from smithy_core.shapes import ShapeID, ShapeType
 from smithy_core.traits import EventHeaderTrait
 from smithy_core.utils import expect_type
 
@@ -57,19 +57,14 @@ class EventDeserializer(SpecificShapeDeserializer):
                         # member_name and a member_index of -1 so the
                         # generated default branch constructs the unknown
                         # variant with the correct tag.
-                        logger.debug(
-                            "Unknown event type: %s", member_name
-                        )
-                        from smithy_core.shapes import ShapeID
+                        logger.debug("Unknown event type: %s", member_name)
 
                         _UNKNOWN_TARGET = Schema(
                             id=ShapeID("smithy.unknown#Unknown"),
                             shape_type=ShapeType.STRUCTURE,
                         )
                         unknown_schema = Schema(
-                            id=ShapeID(
-                                f"smithy.unknown#Unknown${member_name}"
-                            ),
+                            id=ShapeID(f"smithy.unknown#Unknown${member_name}"),
                             shape_type=ShapeType.STRUCTURE,
                             member_target=_UNKNOWN_TARGET,
                             member_index=-1,
--- a/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
+++ b/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
@@ -50,11 +50,39 @@ class EventDeserializer(SpecificShapeDeserializer):
                     message_deserializer = self._create_deserializer(schema, headers)
                     message_deserializer.read_struct(schema, consumer)
                 else:
-                    member_schema = schema.members[member_name]
-                    message_deserializer = self._create_deserializer(
-                        member_schema, headers
-                    )
-                    consumer(member_schema, message_deserializer)
+                    member_schema = schema.members.get(member_name)
+                    if member_schema is None:
+                        # Unknown event type. Call the consumer with a
+                        # schema that carries the event type name as
+                        # member_name and a member_index of -1 so the
+                        # generated default branch constructs the unknown
+                        # variant with the correct tag.
+                        logger.debug(
+                            "Unknown event type: %s", member_name
+                        )
+                        from smithy_core.shapes import ShapeID
+
+                        _UNKNOWN_TARGET = Schema(
+                            id=ShapeID("smithy.unknown#Unknown"),
+                            shape_type=ShapeType.STRUCTURE,
+                        )
+                        unknown_schema = Schema(
+                            id=ShapeID(
+                                f"smithy.unknown#Unknown${member_name}"
+                            ),
+                            shape_type=ShapeType.STRUCTURE,
+                            member_target=_UNKNOWN_TARGET,
+                            member_index=-1,
+                        )
+                        consumer(
+                            unknown_schema,
+                            self._payload_codec.create_deserializer(b"{}"),
+                        )
+                    else:
+                        message_deserializer = self._create_deserializer(
+                            member_schema, headers
+                        )
+                        consumer(member_schema, message_deserializer)
             case "exception":
                 member_name = expect_type(str, headers[":exception-type"])
                 member_schema = schema.members[member_name]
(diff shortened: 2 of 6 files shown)


packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
```diff
@@ -126,3 +127,20 @@ async def test_read_closed_receiver_source() -> None:
     with pytest.raises(IOError):
         await receiver.receive()
     assert receiver.closed
+
+
+def test_deserialize_unknown_event_type():
+    message = EventMessage(
+        headers={
+            ":message-type": "event",
+            ":event-type": "intermediateGroupEvent",
```
jonathan343:
```suggestion
            ":event-type": "unmodeledEvent",
```

packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
```diff
@@ -126,3 +127,20 @@ async def test_read_closed_receiver_source() -> None:
...
+            ":content-type": "application/json",
+        },
+        payload=b"{}",
+    )
+    source = Event.decode(BytesIO(message.encode()))
+    assert source is not None
+    deserializer = EventDeserializer(event=source, payload_codec=JSONCodec())
+    result = EventStreamDeserializer().deserialize(deserializer)
+    assert isinstance(result, EventStreamUnknown)
+    assert result.tag == "intermediateGroupEvent"
```
jonathan343:
```suggestion
    assert result.tag == "unmodeledEvent"
```

packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
```diff
@@ -20,6 +20,7 @@
     EventStreamDeserializer,
     EventStreamErrorEvent,
     EventStreamOperationInputOutput,
+    EventStreamUnknown,
```
jonathan343, line 23:
```suggestion
    EventStreamUnknownEvent,
```

packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
```diff
@@ -126,3 +127,20 @@ async def test_read_closed_receiver_source() -> None:
...
+            ":event-type": "intermediateGroupEvent",
+            ":content-type": "application/json",
+        },
+        payload=b"{}",
+    )
+    source = Event.decode(BytesIO(message.encode()))
+    assert source is not None
+    deserializer = EventDeserializer(event=source, payload_codec=JSONCodec())
+    result = EventStreamDeserializer().deserialize(deserializer)
+    assert isinstance(result, EventStreamUnknown)
```
jonathan343, line 145:
```suggestion
    assert isinstance(result, EventStreamUnknownEvent)
```

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -635,6 +635,19 @@ def _consumer(schema: Schema, de: ShapeDeserializer) -> None:
 ]
 
 
+UNKNOWN_EVENT_CASE = (
+    EventStreamUnknown(tag="intermediateGroupEvent"),
```
jonathan343: nit - this event type is specific to QBusiness, we should use something generic:

```suggestion
    EventStreamUnknownEvent(tag="unmodeledEvent"),
```

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -635,6 +635,19 @@ def _consumer(schema: Schema, de: ShapeDeserializer) -> None:
...
+    EventMessage(
+        headers={
+            ":message-type": "event",
+            ":event-type": "intermediateGroupEvent",
```
jonathan343:
```suggestion
            ":event-type": "unmodeledEvent",
```

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -429,7 +429,7 @@ def _consumer(self, schema: Schema, de: ShapeDeserializer) -> None:
                 self._set_result(EventStreamErrorEvent(ErrorEvent.deserialize(de)))
 
             case _:
-                raise SmithyError(f"Unexpected member schema: {schema}")
+                self._set_result(EventStreamUnknown(tag=schema.member_name or ""))
```
jonathan343:
```suggestion
                self._set_result(EventStreamUnknownEvent(tag=schema.member_name or ""))
```

Also, why `tag=schema.member_name or ""` instead of something like `tag=schema.expect_member_name()`

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -381,7 +381,7 @@ def serialize_members(self, serializer: ShapeSerializer):
 
 
 @dataclass
-class EventStreamUnknownEvent:
+class EventStreamUnknown:
```
jonathan343, line 384: Why the name change here?

```suggestion
class EventStreamUnknownEvent:
```

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
```diff
@@ -396,7 +396,7 @@ def serialize_members(self, serializer: ShapeSerializer):
     | EventStreamPayloadEvent
     | EventStreamBlobPayloadEvent
     | EventStreamErrorEvent
-    | EventStreamUnknownEvent
+    | EventStreamUnknown
```
jonathan343, line 399:
```suggestion
    | EventStreamUnknownEvent
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
```diff
@@ -168,6 +169,7 @@ def _consumer(self, schema: Schema, de: ShapeDeserializer) -> None:
                             ${4C|}
                             case _:
                                 logger.debug("Unexpected member schema: %s", schema)
+                                self._set_result($5L(tag=schema.member_name or ""))
```
jonathan343: Why use `tag=schema.member_name or ""` here instead of something like `tag=schema.expect_member_name()`?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
jonathan343: We should remove this TODO

packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
```diff
@@ -50,11 +50,34 @@ def read_struct(
                     message_deserializer = self._create_deserializer(schema, headers)
                     message_deserializer.read_struct(schema, consumer)
                 else:
-                    member_schema = schema.members[member_name]
-                    message_deserializer = self._create_deserializer(
-                        member_schema, headers
-                    )
-                    consumer(member_schema, message_deserializer)
+                    member_schema = schema.members.get(member_name)
```
jonathan343: I'm thinking we can consolidate this logic into a `member_schema = self._resolve_member_schema(schema, member_name)` method to make this a bit cleaner.

packages/smithy-aws-event-stream/src/smithy_aws_event_stream/_private/deserializers.py
```diff
@@ -50,11 +50,34 @@ def read_struct(
...
+                    if member_schema is None:
+                        # Unknown event type. Call the consumer with a
+                        # schema that carries the event type name as
+                        # member_name and a member_index of -1 so the
+                        # generated default branch constructs the unknown
+                        # variant with the correct tag.
+                        logger.debug("Unknown event type: %s", member_name)
```
jonathan343: This logic seems odd, why can't we do something like:

```python
    def _resolve_member_schema(self, schema: Schema, member_name: str) ->
Schema:
        if member_schema := schema.members.get(member_name):
            return member_schema

        logger.debug(
            "Received unmodeled event stream member %s for union %s",
            member_name,
            schema.id,
        )
        return Schema.member(
            id=schema.id.with_member(member_name),
            target=schema,
            index=-1,
        )
```

This keeps the real event payload/headers intact and lets the generated union fallback produce `*Unknown(tag=...)` in one place, instead of special-casing unknown events in the runtime with a fake {} document.

jonathan343 (changes requested):

packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
Alan4506: I did not add "Event" at the end because the generated code does not have "Event" suffix after "Unknown".

packages/smithy-aws-event-stream/tests/unit/_private/test_deserializers.py
Alan4506: I did not add "Event" at the end because the generated code does not have "Event" suffix after "Unknown".

packages/smithy-aws-event-stream/tests/unit/_private/__init__.py
Alan4506: I modified this because the generated code does not have "Event" suffix after "Unknown".

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
```diff
@@ -168,6 +167,7 @@ def _consumer(self, schema: Schema, de: ShapeDeserializer) -> None:
                             ${4C|}
                             case _:
                                 logger.debug("Unexpected member schema: %s", schema)
```
jonathan343: Logging the schema here is very noisy and bloats logs. I recommend trying it out so you see what I'm talking about. I think we should remove this or update it to not log the schema

Alan4506: It makes sense to me. Logging removed.

jonathan343 (commented):
LGTM. One last request, can you add an `enhancement` changelog entry for the `smithy-aws-event-stream` package? This will ensure these changes get shipped in our next release and generate a related changelog entry.

jonathan343 (approved):
LGTM! Thanks @Alan4506

PR #682, opened 2026-04-14
Fix name collisions between service-defined shapes and framework types

Alan4506:
### Problem

When generating a Python client for services like Q Business, the codegen produces invalid Python code due to two categories of name collisions:

**1. schemas.py — import collision**

[The Q Business model](https://github.com/aws/api-models-aws/blob/75d1a543dd4a4d2a54d3529ffe77da4f544a6046/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L1629-L1631) defines `com.amazonaws.qbusiness#Blob`. The codegen generates a local `BLOB = Schema(...)` in `schemas.py`. [A member elsewhere](https://github.com/aws/api-models-aws/blob/75d1a543dd4a4d2a54d3529ffe77da4f544a6046/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L6395-L6409) targets `smithy.api#Blob`, causing `from smithy_core.prelude import BLOB` to be added in `schemas.py`. The local definition shadows the import.

**2. models.py — import collision**

[The Q Business model](https://github.com/aws/api-models-aws/blob/75d1a543dd4a4d2a54d3529ffe77da4f544a6046/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L5941) defines `com.amazonaws.qbusiness#Document`. The codegen generates `class Document:` in `models.py`. [The model](https://github.com/aws/api-models-aws/blob/75d1a543dd4a4d2a54d3529ffe77da4f544a6046/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L274-L276) also defines `document` type shapes (e.g., `ActionPayloadFieldValue`), causing `from smithy_core.documents import Document` to be added in `models.py`. The local class definition shadows the import.

**3. models.py — union member variant name collision**

[The Q Business model](https://github.com/aws/api-models-aws/blob/75d1a543dd4a4d2a54d3529ffe77da4f544a6046/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json#L13086-L13094) defines:
```json
"com.amazonaws.qbusiness#Principal": {
    "type": "union",
    "members": {
        "user": {
            "target": "com.amazonaws.qbusiness#PrincipalUser"
        }
    }
}
```
The codegen synthesizes variant class name `Principal` + `capitalize("user")` = `PrincipalUser`, which collides with the target structure `PrincipalUser` that is also generated as a separate class in the same file.

Additionally, the framework defines `AuthOption` in both `smithy_core.auth` and `smithy_core.interfaces.auth`. When both are imported into the same file (e.g., `auth.py`), one shadows the other. This was previously handled with manual aliases (e.g., `addImport("smithy_core.interfaces.auth", "AuthOption", "AuthOptionProtocol")`), along with similar manual aliases for `EndpointResolver` and `HTTPResponse`. These worked but were ad-hoc — each case required a hardcoded alias in the generator code. Our change replaces all manual aliases with a unified automatic collision detection system.

### Description of Changes

**Import collisions — deferred collision detection in PythonWriter**

Inspired by smithy-java's `DeferredSymbolWriter` pattern:

- Created `RuntimeTypes.java` and `AwsRuntimeTypes.java` with pre-defined `Symbol` constants for all framework types used across generators. This replaces hardcoded `writer.addImport("smithy_core.X", "Name")` + string literal `"Name"` patterns with `$T` using RuntimeTypes symbols, making all framework type references visible to the collision detection system.

- Modified `PythonWriter` to track a **symbol table** (all symbols referenced via `$T`, keyed by simple name) and a **locally-defined names set** (class names and top-level variables registered by generators via `addLocallyDefinedSymbol()`). The `$T` formatter now returns a placeholder token (`«namespace.Name»`) instead of the final name. At `toString()` time, `resolveCollisions()` compares the two sets and:
  - If an imported name collides with a locally-defined name → aliases the import as `_Name`
  - If two imports share the same simple name → aliases each with a module-based name (e.g., `_smithy_core_auth_AuthOption` and `_smithy_core_interfaces_auth_AuthOption`)
  - Import statements are generated *after* collision resolution so aliases take effect

- Added `aliasImport()` to `ImportDeclarations` to retroactively rewrite an import's alias at `toString()` time.

- All generators that define classes or top-level names now call `addLocallyDefinedSymbol()`.

**Union member variant collisions — name check in PythonSymbolProvider**

- Added `allShapeNames` set to `PythonSymbolProvider`, populated at construction time with the PascalCase names of all shapes that generate classes (structure, union, enum, intEnum).

- In `memberShape()`, after synthesizing the variant name (`UnionName + capitalize(MemberName)`), checks if it collides with any shape name. If so, uses `UnionName_MemberName` (e.g., `Principal_User`) instead. Smithy shape names cannot contain underscores, so this format is guaranteed collision-free.

- Same check in `unionShape()` for the `Unknown` variant (`UnionName + "Unknown"` → `UnionName_Unknown` if collision detected).

### Testing

We successfully generated Q Business code with no ruff errors and no pyright errors. We inspected the generated code:

Import collisions resolved — `models.py`:
```python
from smithy_core.documents import Document as _Document, TypeRegistry
# ...
class Document:  # service-defined class keeps its name
    """A document in an Amazon Q Business application."""
# ...
value: _Document  # framework Document uses alias
```

Import collisions resolved — `_private/schemas.py`:
```python
from smithy_core.prelude import BLOB as _BLOB, BOOLEAN, UNIT
# ...
BLOB = Schema(id=ShapeID("com.amazonaws.qbusiness#Blob"), ...)  # local definition keeps its name
members={"blob": {"target": _BLOB}, ...}  # prelude BLOB uses alias
```

Import-vs-import collision resolved — `auth.py`:
```python
from smithy_core.auth import AuthOption as _smithy_core_auth_AuthOption, AuthParams
from smithy_core.interfaces.auth import AuthOption as _smithy_core_interfaces_auth_AuthOption
```

Import-vs-import collision resolved — `tests/test_protocol.py`:
```python
from smithy_http.aio import HTTPResponse as _smithy_http_aio_HTTPResponse
from smithy_http.aio.interfaces import HTTPRequest, HTTPResponse as _smithy_http_aio_interfaces_HTTPResponse
```

Union member variant collision resolved — `models.py`:
```python
class PrincipalUser:    # service-defined structure, unchanged
class PrincipalGroup:   # service-defined structure, unchanged
class Principal_User:   # union variant, disambiguated with "_"
class Principal_Group:  # union variant, disambiguated with "_"
```

We also locally tested that all modules import successfully, and integration tests pass.

We also regenerated all three existing clients (Bedrock Runtime, Sagemaker Runtime, Transcribe Streaming). All build successfully and tests pass.


codegen/core/src/main/java/software/amazon/smithy/python/codegen/RuntimeTypes.java
```diff
@@ -0,0 +1,175 @@
...
+            "aio.protocols",
+            "RestJsonClientProtocol",
+            SmithyPythonDependency.SMITHY_AWS_CORE);
+
+    // smithy_aws_core.identity 
+    public static final Symbol STATIC_CREDENTIALS_RESOLVER = createSymbol(
+            "identity",
+            "StaticCredentialsResolver",
+            SmithyPythonDependency.SMITHY_AWS_CORE);
+
```
ubaskota, line 166: I noticed `REST_JSON_CLIENT_PROTOCOL` and `STATIC_CREDENTIALS_RESOLVER` are defined in both `RuntimeTypes.java` and `AwsRuntimeTypes.java`. Was the duplication intentional? Do we need them in both files?

Alan4506: Good catch! The duplication was unintentional. These two symbols are only used by generators in the core module (`RestJsonProtocolGenerator` and `HttpProtocolTestGenerator`), which should not depend on the AWS module. I've removed them from `AwsRuntimeTypes.java` and kept them only in `RuntimeTypes.java`.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/ImportDeclarations.java
```diff
@@ -104,6 +104,36 @@ private ImportDeclarations addImportToMap(
...
+     * is detected between an imported name and a locally-defined name. The import statement
+     * is rewritten from {@code from namespace import name} to {@code from namespace import name as alias}.
+     *
+     * @param namespace The module namespace of the import.
+     * @param name The original imported name.
+     * @param alias The alias to use instead.
+     */
+    void aliasImport(String namespace, String name, String alias) {
+        aliasImportInMap(namespace, name, alias, externalImports);
+        aliasImportInMap(namespace, name, alias, localImports);
```
jonathan343: This collision handling also needs to update stdlib imports when it rewrites symbol references.

Example: if a Smithy model defines a structure named Decimal and another shape uses bigDecimal, the generated Python can end up like this:

```python
from decimal import Decimal

class Decimal:
    ...

@dataclass
class Foo:
    value: _Decimal
```

The body was rewritten to use `_Decimal`, but the import still exposes `Decimal`, so importing the module fails with `NameError: name '_Decimal' is not defined.`

Could we either include `stdlibImports` in the alias rewrite path or avoid aliasing stdlib symbols here? If you considered this and there is an issue with it, let me know

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
```diff
@@ -347,6 +475,28 @@ public String apply(Object type, String indent) {
...
+         * <p>Only symbols that are imported from another module (have a namespace but no
+         * definition file) get a placeholder. Symbols without a namespace (builtins like
+         * int, str, bool) and symbols with a definition file (locally-defined types) are
+         * returned as-is since they don't need collision detection.
+         */
+        private String resolvePlaceholder(Symbol symbol) {
+            if (symbol.getNamespace().isEmpty() || !symbol.getDefinitionFile().isEmpty()) {
+                // No namespace means builtin. Has definition file means locally-defined.
+                // Neither case needs collision detection.
+                return symbol.getName();
```
jonathan343, line 505: This early return seems too broad. A symbol with a definition file may still be imported into a different generated file, so skipping it here can hide real name collisions from the import resolver.

Example: `config.py` may need both a runtime resolver and a generated model with the same name:

```python
from smithy_aws_core.endpoints import StandardRegionalEndpointsResolver
from .models import StandardRegionalEndpointsResolver

class Config:
    def __init__(self):
        self.endpoint_resolver = StandardRegionalEndpointsResolver(
            endpoint_prefix="..."
        )
```

The `.models` import shadows the runtime resolver, so this calls the generated model class with resolver arguments. The expected output would alias one of them, for example:

```python
from smithy_aws_core.endpoints import StandardRegionalEndpointsResolver as _RegionalResolver
from .models import StandardRegionalEndpointsResolver
```

Could we only skip symbols that are actually defined in the current writer, while still registering generated symbols that are imported from other generated files?

jonathan343 (commented):
Thanks @Alan4506!

I left some comments on potential gaps we should look into. Can you look into these and address if appropriate? It would also be ideal to add some related tests to prevent regressions

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/ImportDeclarations.java
Alan4506: Included `stdlibImports` in the alias rewrite path and added tests in https://github.com/smithy-lang/smithy-python/pull/682/changes/a7b667670285bdd00981f3cf67231140b59ac99e. Thank you!

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
Alan4506: Addressed and tested in https://github.com/smithy-lang/smithy-python/pull/682/changes/a7b667670285bdd00981f3cf67231140b59ac99e. Thank you!

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
```diff
@@ -316,18 +338,124 @@ public PythonWriter maybeWrite(boolean shouldWrite, Object content, Object... ar
...
+
+            if (collidesWithLocal && !hasImportCollision) {
+                // Single import collides with a locally-defined name — alias it as _Name.
+                for (Symbol symbol : symbols) {
+                    String placeholderKey = symbol.getNamespace() + "." + symbol.getName();
+                    String aliasedName = "_" + simpleName;
+                    getImportContainer().aliasImport(symbol.getNamespace(), simpleName, aliasedName);
+                    resolvedNames.put(placeholderKey, aliasedName);
+                }
+            } else {
```
jonathan343, line 417: When two types share a simple name (e.g. `AuthOption` from `smithy_core.auth` and `smithy_core.interfaces.auth`), `resolveCollisions()` aliases both with module-derived names. IMO this is way less readable and also changes existing imports when a new one is added that introduced a collision. One the new one needs aliasing.

Can we keep the original (if it is pre-existing) or first one (if both are new) with the plain name and alias only the subsequent ones that collide?

Example diff from the re-generated `aws-sdk-bedrock-runtime`:

```diff
-from smithy_core.auth import AuthOption, AuthParams
-from smithy_core.interfaces.auth import AuthOption as AuthOptionProtocol
+from smithy_core.auth import AuthOption as _smithy_core_auth_AuthOption, A
uthParams
+from smithy_core.interfaces.auth import (
+    AuthOption as _smithy_core_interfaces_auth_AuthOption,
+)
 from smithy_core.shapes import ShapeID


 class HTTPAuthSchemeResolver:
     def resolve_auth_scheme(
         self, auth_parameters: AuthParams[Any, Any]
-    ) -> list[AuthOptionProtocol]:
-        auth_options: list[AuthOptionProtocol] = []
+    ) -> list[_smithy_core_interfaces_auth_AuthOption]:
+        auth_options: list[_smithy_core_interfaces_auth_AuthOption] = []
```

Additionally, can we keep custom aliases? Can we still define our own preferred alias when we don't want to rely on the less readable one? 

Does java also use a module path prefix?

jonathan343 (dismissed):

codegen/core/src/main/java/software/amazon/smithy/python/codegen/writer/PythonWriter.java
Alan4506: Good catch on all points. I agree with the direction.

>Does java also use a module path prefix?

No. smithy-java doesn't alias on collision at all. It drops the import and writes the fully-qualified name inline at every use site. See its own [naming-conflict test fixture](https://github.com/smithy-lang/smithy-java/blob/cafffaf6f352e1b39313109853a33b67285533d5/codegen/codegen-plugin/src/test/resources/software/amazon/smithy/java/codegen/types/test-cases/naming-conflict/expected/Schema.java), where a generated Schema collides with the framework `software.amazon.smithy.java.core.schema.Schema` and the latter is written fully-qualified everywhere, such as:
```
public software.amazon.smithy.java.core.schema.Schema schema() {
```

Python can't write inline FQNs in annotations, so we do import + alias instead. Given that, your suggestion (keep the first/pre-existing one bare and alias only the subsequent collisions, and probably allow a custom alias) seems like a reasonable way to improve readability. However, I'd lean towards opening an issue as a follow-up rather than in this PR. Let me know if you think differently.

**Heads up:** I pushed a new [commit](https://github.com/smithy-lang/smithy-python/pull/682/changes/405fad28a8b5689af3fd913ede49e03eb118ba94) to rebase onto the latest develop. Could you take another look when you get a chance? Thanks!

jonathan343 (approved):

PR #691, opened 2026-05-04
Fix missing signed empty end frame on event stream close

alexgromero:
*Issue #, if available:*
https://github.com/awslabs/aws-sdk-python/issues/46

*Description of changes:*
Fix `AWSEventPublisher.close()` to send a signed empty end frame to signal stream completion.

SigV4 signed event streams require a final signed empty message before the HTTP body stream closes. This fix:
1. Adds `sign_empty()` method to the `EventSigner` protocol in `smithy-core`.
2. Updates `AWSEventPublisher.close()` to call `sign_empty()` when sending the termination frame in `smithy-aws-event-stream`.

This eliminates the following errors:
- Transcribe Streaming: `BadRequestException: A complete signal was sent without the preceding empty frame.`
- Bedrock Runtime: `ValidationException: Invalid input request, please fix your input and try again.`

Related PR: https://github.com/smithy-lang/smithy-python/pull/692

### DIFF
--- a/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py
+++ b/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py
@@ -17,7 +17,7 @@ from smithy_core.serializers import SerializeableShape
 
 from .._private.deserializers import EventDeserializer as _EventDeserializer
 from .._private.serializers import EventSerializer as _EventSerializer
-from ..events import Event
+from ..events import Event, EventMessage
 from ..exceptions import EventError
 
 logger = logging.getLogger(__name__)
@@ -84,6 +84,20 @@ class AWSEventPublisher[E: SerializeableShape](EventPublisher[E]):
         if self._closed:
             return
         self._closed = True
+
+        # Send a signed empty frame to signal stream completion.
+        if self._signing_config is not None:
+            end_frame = EventMessage()
+            identity = await self._signing_config.identity_resolver.get_identity(
+                properties=self._signing_config.identity_properties
+            )
+            end_frame = await self._signing_config.signer.sign(
+                event=end_frame,
+                identity=identity,
+                properties=self._signing_config.signing_properties,
+            )
+            logger.debug("Sending signed empty message to terminate the event stream")
+            await self._writer.write(end_frame.encode())
 
         if (close := getattr(self._writer, "close", None)) is not None:
             if asyncio.iscoroutine(result := close()):
--- a/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py
+++ b/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py
@@ -91,12 +91,12 @@ class AWSEventPublisher[E: SerializeableShape](EventPublisher[E]):
             identity = await self._signing_config.identity_resolver.get_identity(
                 properties=self._signing_config.identity_properties
             )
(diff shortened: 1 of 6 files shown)


Alan4506:
Just curious: with this change applied, are we able to modify some existing integration tests to remove `await asyncio.sleep` statements? For example, [Bedrock Runtime](https://github.com/awslabs/aws-sdk-python/blob/44dc092d89706412124bf04d33f9378afd81d76f/clients/aws-sdk-bedrock-runtime/tests/integration/test_bidirectional_streaming.py#L194) and [Transcribe Streaming](https://github.com/awslabs/aws-sdk-python/blob/44dc092d89706412124bf04d33f9378afd81d76f/clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py#L56)? When I wrote tests for new clients (e.g., Lex Runtime V2 and Q business), I also had similar concerns - tests would fail without `await asyncio.sleep` statements.

Alan4506 (dismissed):

Alan4506:
Approved. However, there are some other issues when I test this with the [Transcribe Streaming integration tests](https://github.com/awslabs/aws-sdk-python/blob/44dc092d89706412124bf04d33f9378afd81d76f/clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py#L56) and the [simple_file.py example](https://github.com/awslabs/aws-sdk-python/blob/44dc092d89706412124bf04d33f9378afd81d76f/clients/aws-sdk-transcribe-streaming/examples/simple_file.py#L118). The end frame fix resolves the "complete signal was sent without the preceding empty frame" error, but the `asyncio.sleep` calls before `close()` seem still needed. Without them, it returns `aws_sdk_transcribe_streaming.models.BadRequestException: Could not decode the audio stream that you provided. Check that the audio stream is valid and try your request again.` I wonder if this is expected behavior or something that needs a separate fix.

Alan4506 (dismissed):

packages/smithy-aws-event-stream/src/smithy_aws_event_stream/aio/__init__.py
```diff
@@ -85,6 +85,20 @@ async def close(self) -> None:
             return
         self._closed = True
 
+        # Send a signed empty frame to signal stream completion.
```
jonathan343: Could we wrap the signed-empty-frame write in a `try/finally` so the writer is always closed?

It may also be worth making the test assert that the writer is closed, or using AsyncBytesProvider to decode the written frame and verify the terminator payload is actually empty.

jonathan343: Do you know if this is safe for even non-bidirectional APIs?

jonathan343: I just realized all AWS operations that use input streaming are also bi-directional

alexgromero: Addressed in d1fc05770c77a212e3c0715445290bd6f3c4e4bb.

alexgromero: Yep, thanks for confirming! Also worth noting the Java SDK sends the trailing empty frame unconditionally for all signed event streams, no bidi check, so this appears to be part of the SigV4 protocol itself rather than a bidi-specific behavior.

jonathan343 (approved):

PR #692, opened 2026-05-04
Add SigV4 event stream empty-frame signing

jonathan343:
Related Issue: https://github.com/awslabs/aws-sdk-python/issues/46
Related PR: https://github.com/smithy-lang/smithy-python/pull/691

### Summary
Adds `AsyncEventSigner.sign_empty` for producing SigV4-signed Amazon Event Stream terminator frames with zero-byte payloads.

### Details
Normal event signing wraps an encoded event message as the payload. Some AWS event stream services require a final signed empty frame before the HTTP body stream closes, which must use an actual zero-byte payload instead.

This keeps normal `sign()` behavior unchanged and shares the common signing logic through a private helper.

### Testing

I ported a modified version https://github.com/smithy-lang/smithy-python/pull/691 that calls `sign_empty` before closing the stream and confirmed that the [simple_file.py](https://github.com/awslabs/aws-sdk-python/blob/develop/clients/aws-sdk-transcribe-streaming/examples/simple_file.py) script ran successfully with just a `stream.close()` following the completion of sending all audio chunks. 

- Added coverage for normal event signing payload wrapping
- Added coverage for signed empty-frame payload behavior

---

### DIFF
--- a/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
+++ b/packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
@@ -835,6 +835,7 @@ class AsyncEventSigner:
         self._prior_signature = initial_signature
         self._signing_lock = asyncio.Lock()
         self._event_encoder_cls = event_encoder_cls
+        self._is_finalized = False
 
     async def sign(
         self,
@@ -859,18 +860,25 @@ class AsyncEventSigner:
     ) -> "EventMessage":
         """Sign an Amazon Event Stream final empty message.
 
-        The terminator must have a zero-byte outer payload. Calling ``sign``
-        with an empty event message is not equivalent because ``sign`` wraps the
-        encoded event message as the payload.
+        The supplied ``event`` must have no headers and an empty payload.
+        Calling ``sign`` with an empty event message is not equivalent because
+        ``sign`` wraps the encoded event message as the payload.
 
         The returned message contains ``:date`` and ``:chunk-signature``
-        headers and advances the running signature chain.
+        headers and advances the running signature chain. After this method
+        returns, the signer cannot sign additional event stream messages.
         """
+        if event.headers or event.payload:
+            raise ValueError(
+                "sign_empty requires an event with no headers and an empty payload."
+            )
+
         return await self._sign_payload(
             event=event,
             payload=b"",
             identity=identity,
             properties=properties,
+            is_final=True,
         )
 
     async def _sign_payload(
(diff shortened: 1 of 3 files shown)


packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -842,10 +842,47 @@ async def sign(
...
+        The terminator must have a zero-byte outer payload. Calling ``sign``
+        with an empty event message is not equivalent because ``sign`` wraps the
+        encoded event message as the payload.
+
+        The returned message contains ``:date`` and ``:chunk-signature``
+        headers and advances the running signature chain.
+        """
+        return await self._sign_payload(
+            event=event,
+            payload=b"",
```
SamRemis, line 878: Should we confirm that the event that is passed in has no payload before signing as if the payload were empty?

This isn't a major concern, but it's a public interface so I think it's worth safeguarding here

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
```diff
@@ -842,10 +842,47 @@ async def sign(
...
+    async def _sign_payload(
+        self,
+        *,
+        event: "EventMessage",
+        payload: bytes,
+        identity: _AWSCredentialsIdentity,
+        properties: SigV4SigningProperties,
     ) -> "EventMessage":
         async with self._signing_lock:
-            # Copy and prepopulate any missing values in the
```
SamRemis, line 847: So this lock on the line above - I don't feel super strongly about this since it's a private lock, but it may make sense for us to safeguard here as well.  

If someone tries to sign a new event _after_ the empty payload has been sent, they will see a confusing service side error.  We may want to raise an error client side letting them know that the stream was closed.

SamRemis (dismissed):

alexgromero (dismissed):

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
jonathan343: Addressed in 779d455ef21937841ce6418e896cc43facf1c262

packages/aws-sdk-signers/src/aws_sdk_signers/signers.py
jonathan343: Addressed in 779d455ef21937841ce6418e896cc43facf1c262

SamRemis (approved):

PR #697, opened 2026-05-13
Fix AWSCRTHTTPClient request bodies for HTTP/1.1 connections

arandito:
### Description
Previously, our `AWSCRTHTTPClient` attached request bodies via the CRT's `request_body_generator` parameter, which is only supported for HTTP/2 connections ([source](https://github.com/awslabs/aws-crt-python/blob/main/awscrt/aio/http.py#L205-L206)). This caused generated clients that only support HTTP/1.1 to throw the following error when making body-bearing requests:
```
smithy_core.exceptions.SmithyError: 2056 (AWS_ERROR_HTTP_MISSING_BODY_STREAM): Given the provided headers (ex: Content-Length), a body is expected.
```

This PR adds logic that routes HTTP/1.1 connections to attach bodies through the `body_stream` parameter on the request itself and keeps the generator path for HTTP/2.

### Testing
- Generated H1-only service (STS) and confirmed calls no longer raise `AWS_ERROR_HTTP_MISSING_BODY_STREAM`
- Confirmed existing client integration tests succeed with changes

### DIFF
--- a/packages/smithy-http/src/smithy_http/aio/crt.py
+++ b/packages/smithy-http/src/smithy_http/aio/crt.py
@@ -6,6 +6,7 @@ from collections.abc import AsyncGenerator, AsyncIterable
 from copy import deepcopy
 from dataclasses import dataclass
 from inspect import iscoroutinefunction
+from io import BytesIO
 from typing import TYPE_CHECKING, Any
 
 from awscrt.exceptions import AwsCrtError
@@ -36,6 +37,7 @@ except ImportError:
 from smithy_core import interfaces as core_interfaces
 from smithy_core.aio import interfaces as core_aio_interfaces
 from smithy_core.aio.types import AsyncBytesReader
+from smithy_core.aio.utils import read_streaming_blob_async
 from smithy_core.exceptions import MissingDependencyError
 
 from .. import Field, Fields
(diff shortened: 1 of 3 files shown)


packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -175,13 +177,18 @@ async def send(
...
 
-            # Convert body to async iterator for request_body_generator
-            body_generator = self._create_body_generator(request.body)
-
-            crt_stream = connection.request(
-                crt_request,
-                request_body_generator=body_generator,
-            )
+            # request_body_generator is HTTP/2-only in CRT; HTTP/1.1 must use body_stream
+            if connection.version == crt_http.HttpVersion.Http2:
```
Alan4506: Nit: do we want to use "is" instead of "==" to align with `if force_http_2 and connection.version is not crt_http.HttpVersion.Http2:`https://github.com/smithy-lang/smithy-python/blob/720c0cb70c183a2055eae7a37f58688e068b928a/packages/smithy-http/src/smithy_http/aio/crt.py#L271

Alan4506 (commented):
LTGM overall. Just wondering if there is a roadmap / tracking issue on the [aws-crt-python](https://github.com/awslabs/aws-crt-python/blob/v0.32.2/awscrt/aio/http.py#L196-L210) side about **when** they will support HTTP/1.1 streaming request bodies, so we can switch this back from buffering.

arandito:
> LTGM overall. Just wondering if there is a roadmap / tracking issue on the [aws-crt-python](https://github.com/awslabs/aws-crt-python/blob/v0.32.2/awscrt/aio/http.py#L196-L210) side about **when** they will support HTTP/1.1 streaming request bodies, so we can switch this back from buffering.

Not yet. I am going to ask their team if this is on their roadmap as a follow-up. I'm not sure if there are protocol-level limitations on their side for why they can't support body generators for H1. This solution is an immediate fix to unblock future H1-only clients but ideally we can revert back to their generators if they add support in the future.

packages/smithy-http/src/smithy_http/aio/crt.py
arandito: Good catch, yea enums should be compared based on identity so I can switch that. I will update this one I get additional feedback in case there are larger changes.

jonathan343 (dismissed):

packages/smithy-http/src/smithy_http/aio/crt.py
arandito: Fixed in [97aef6e](https://github.com/smithy-lang/smithy-python/pull/697/commits/97aef6e824a0593bb07147e996ed8cb62eb460a0)

Alan4506 (approved):

PR #699, opened 2026-05-20
Rename Config and Plugin to service-specific names

ubaskota:
*Issue #, if available:*

*Description of changes:*
Rename generated Config class to service-specific names derived from ServiceTrait sdkId (e.g., `BedrockRuntimeConfig`). Similarly rename Plugin to `<ServiceName>Plugin`. Deprecated aliases with `DeprecationWarning` are generated for backwards compatibility. Generic (Non-AWS SDKs) remain unchanged.

*Testing*
- Verified that all tests pass (`make test-py` and `make test-protocols`)
- Verified that “deprecation warning” is emitted when deprecated `Config` is used
- Verified that both `Config` and `BedrockRuntimeConfig` successfully make requests and receive responses
- Compared the diffs and verified that there are no changes for non-AWS sdks
- Compared the diffs and verified that capybara sdk generates `BedrockRuntimeConfig/BedrockRuntimePlugin` with deprecated aliases as expected

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/CodegenUtils.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/CodegenUtils.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.aws.traits.ServiceTrait;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NullableIndex;
 import software.amazon.smithy.model.node.Node;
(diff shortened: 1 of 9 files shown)


Alan4506 (commented):
LGTM overall. Just curious: why add the changelog entries to `aws-sdk-signers`, `smithy-aws-core`, and `smithy-aws-event-stream`? None of those packages' source actually changes in this PR. Do we have a better way to surface this codegen-level change?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/CodegenUtils.java
```diff
@@ -63,24 +64,46 @@ public final class CodegenUtils {
...
      * @param settings The client settings, used to account for module configuration.
+     * @param model The model containing the service shape.
      * @return Returns the client's configuration object symbol.
      */
-    public static Symbol getConfigSymbol(PythonSettings settings) {
+    public static Symbol getConfigSymbol(PythonSettings settings, Model model) {
+        var service = settings.service(model);
+        var name = service.getTrait(ServiceTrait.class)
+                .map(trait -> StringUtils.capitalize(trait.getSdkId()).replace(" ", "") + "Config")
+                .orElse("Config");
```
SamRemis, line 101: What circumstances does this fallback come into play? 

I'd think you need an SDK ID to generate a client, so I'm not sure this fallback is useful.  Am I missing something?

ubaskota: The fallback is for non-AWS (generic sdks) that don't have a ServiceTrait (no SDK ID). Updated the comment to clarify.

jonathan343 (commented):
Thanks Ujjwal!

Our current documentation generation in aws/aws-sdk-python relies on the current config and plugin names. We'll need to simultaneously ship a change to use the updated names so we don't break our docs.

PR #702, opened 2026-05-26
Implement client error correction

Alan4506:
## Description of changes

This codegen generates Python clients from Smithy service models. For each operation's input and output shape, it produces a Python `@dataclass` whose fields mirror the model members. Members that the service model marks as `@required` are emitted as non-default keyword-only arguments, meaning the dataclass cannot be constructed without them.

After an operation completes, the SDK's deserialization layer reads the response body and calls the dataclass constructor with the parsed values. When the response follows the model, this works fine. **However**, when the response is *missing a `@required` member*, the constructor raises a Python `TypeError` for the missing keyword-only argument. The SDK pipeline catches that, classifies it as non-retryable, and re-raises as `smithy_core.exceptions.SmithyError`. The user sees something like:

```text
smithy_core.exceptions.SmithyError: SomeOutput.__init__() missing 1 required keyword-only argument: 'foo'
```

This happens in two practical situations:

- **A successful response is missing one of the response shape's `@required` members.** For example, [Connect Health](https://github.com/aws/api-models-aws/blob/main/models/connecthealth/service/2025-01-29/connecthealth-2025-01-29.json) `CreateSubscription` returns an HTTP 200 response that omits the `@required lastUpdatedAt` field of `CreateSubscriptionOutput`. The subscription is created server-side, but the SDK call fails because the output dataclass cannot be constructed without `last_updated_at`.
- **An error response is missing one of the error shape's `@required` members.** For example, [Q Business](https://github.com/aws/api-models-aws/blob/main/models/qbusiness/service/2023-11-27/qbusiness-2023-11-27.json) `ChatSync`, when rejecting a request with `ValidationException`, returns an exception body that omits the `@required reason` field of `ValidationException`. The user sees a generic message about a missing kwarg instead of the actual service-side error message: the real reason the request was rejected is masked.

The Smithy spec accounts for this in [§ 3.3.3.1.1 Client error correction](https://smithy.io/2.0/spec/aggregate-types.html#client-error-correction): clients MAY substitute a zero-value default for a missing required member to avoid downtime. This PR implements that policy in the Python codegen. Each shape type gets the spec's prescribed default. The implementation lives in a new `MemberErrorCorrectionGenerator` (a `ShapeVisitor`) plus two hooks in `StructureGenerator`: error correction at the end of the generated `deserialize_kwargs`, and a generated `_smithy_default()` classmethod used when a parent's required nested-struct member is missing on the wire.

**Note**: enum and intEnum currently fall back to placeholder zero values (`""` and `0`). Spec § 3.3.3.1.1 recommends (SHOULD) that these types define a structured unknown variant for receiving wire values not known at SDK-generation time. Implementing it requires redesigning the generated enum/intEnum classes: today they are plain `StrEnum` / `IntEnum` subclasses with no room for an unknown member, so if necessary, a follow-up would need to change `EnumGenerator` / `IntEnumGenerator` (and likely `PythonSymbolProvider`, `MemberDeserializerGenerator`) to emit a wrapper type that exposes something like `MyEnum.unknown(value)` and `is_unknown`. The two enum visitor methods in `MemberErrorCorrectionGenerator` have TODO comments documenting this.

## Testing

`:core:test`, `:core:integ`, `:core:spotlessCheck`, the `aws-query` and `rest-json-1` protocol-test suites all pass.

Also, two real-world cases were verified before/after:

**Amazon Connect Health `CreateSubscription`.** Before this PR:
```text
  File ".../smithy_http/aio/protocols.py", line 152, in deserialize_response
    return operation.output.deserialize(deserializer)
  File ".../aws_sdk_connecthealth/models.py", line 1491, in deserialize
    return cls(**cls.deserialize_kwargs(deserializer))
TypeError: CreateSubscriptionOutput.__init__() missing 1 required keyword-only argument: 'last_updated_at'

The above exception was the direct cause of the following exception:
  File ".../smithy_core/aio/client.py", line 258, in _execute_request
    raise SmithyError(e) from e
smithy_core.exceptions.SmithyError: CreateSubscriptionOutput.__init__() missing 1 required keyword-only argument: 'last_updated_at'
```

After this PR, `last_updated_at` is filled with `datetime.fromtimestamp(0, tz=timezone.utc)`, the call returns normally.

**Amazon Q Business `ChatSync`.** Before this PR:

```text
smithy_core.exceptions.SmithyError: ValidationException.__init__() missing 1 required keyword-only argument: 'reason'
```

After this PR:

```text
aws_sdk_qbusiness.models.ValidationException: Please provide a unique clientToken that can be used as an idempotency attribute.
```

`reason` is filled with `""`, the `ValidationException` constructs cleanly, and the actual validation message reaches the caller.

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
@@ -13,6 +13,25 @@ import software.amazon.smithy.python.codegen.PythonSettings;
 import software.amazon.smithy.python.codegen.SymbolProperties;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
+/**
+ * Renders intEnums as an {@code IntEnum} subclass.
+ *
+ * <p>Beyond the named members, the generated class has:
+ * <ul>
+ *   <li>{@code is_unknown} — public. {@code True} when the value didn't come
+ *       from a known member: either the service returned a value newer than
+ *       this SDK or client error correction filled in a placeholder.</li>
+ *   <li>{@code _missing_} — invoked when deserializing a value the SDK
+ *       doesn't recognize.</li>
+ *   <li>{@code _unknown} — invoked from {@link MemberErrorCorrectionGenerator}
+ *       to fill a missing required member.</li>
+ *   <li>{@code __eq__} / {@code __hash__} — overridden so an unknown value
+ *       is not equal to any known member, even if its underlying integer
+ *       happens to match one.</li>
+ * </ul>
+ *
+ * @see <a href="https://smithy.io/2.0/spec/simple-types.html#intenum">Smithy spec: intEnum</a>
+ */
 @SmithyInternalApi
 public final class IntEnumGenerator implements Runnable {
 
@@ -27,6 +46,7 @@ public final class IntEnumGenerator implements Runnable {
         var enumSymbol = directive.symbol().expectProperty(SymbolProperties.ENUM_SYMBOL);
         directive.context().writerDelegator().useShapeWriter(directive.shape(), writer -> {
             writer.addStdlibImport("enum", "IntEnum");
+            writer.addStdlibImport("typing", "Self");
             writer.openBlock("class $L(IntEnum):", "", enumSymbol.getName(), () -> {
                 directive.shape().getTrait(DocumentationTrait.class).ifPresent(trait -> {
                     writer.writeDocs(trait.getValue(), directive.context());
@@ -35,11 +55,48 @@ public final class IntEnumGenerator implements Runnable {
                 for (MemberShape member : directive.shape().members()) {
                     var name = directive.symbolProvider().toMemberName(member);
                     var value = member.expectTrait(EnumValueTrait.class).expectIntValue();
-                    writer.write("$L = $L\n", name, value);
+                    writer.write("$L = $L", name, value);
                     member.getTrait(DocumentationTrait.class).ifPresent(trait -> {
                         writer.writeDocs(trait.getValue(), directive.context());
                     });
                 }
+
+                writer.write("""
+
+                        @classmethod
+                        def _unknown(cls, value: int) -> "Self":
+                            pseudo = int.__new__(cls, value)
+                            pseudo._name_ = f"<smithy-unknown:{value}>"
+                            pseudo._value_ = value
+                            return pseudo
+
+                        @classmethod
+                        def _missing_(cls, value: object) -> "Self | None":
+                            if isinstance(value, int):
+                                return cls._unknown(value)
+                            return None
+
+                        @property
+                        def is_unknown(self) -> bool:
+                            \"""True if this value was not known at SDK generation time.\"""
+                            return self._name_ not in type(self).__members__
+
+                        def __eq__(self, other: object) -> bool:
+                            if self.is_unknown:
+                                return (
+                                    isinstance(other, type(self))
+                                    and other.is_unknown
+                                    and self._value_ == other._value_
+                                )
+                            if isinstance(other, type(self)) and other.is_unknown:
+                                return False
+                            return super().__eq__(other)
+
+                        def __hash__(self) -> int:
+                            if self.is_unknown:
+                                return hash(("<smithy-unknown>", type(self).__name__, self._value_))
+                            return super().__hash__()
+                        """);
             });
         });
     }
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
@@ -10,25 +10,17 @@ import software.amazon.smithy.model.traits.DocumentationTrait;
 import software.amazon.smithy.model.traits.EnumValueTrait;
 import software.amazon.smithy.python.codegen.GenerationContext;
 import software.amazon.smithy.python.codegen.PythonSettings;
-import software.amazon.smithy.python.codegen.SymbolProperties;
+import software.amazon.smithy.python.codegen.SmithyPythonDependency;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
 /**
  * Renders intEnums as an {@code IntEnum} subclass.
  *
- * <p>Beyond the named members, the generated class has:
- * <ul>
- *   <li>{@code is_unknown} — public. {@code True} when the value didn't come
- *       from a known member: either the service returned a value newer than
- *       this SDK or client error correction filled in a placeholder.</li>
- *   <li>{@code _missing_} — invoked when deserializing a value the SDK
- *       doesn't recognize.</li>
- *   <li>{@code _unknown} — invoked from {@link MemberErrorCorrectionGenerator}
- *       to fill a missing required member.</li>
- *   <li>{@code __eq__} / {@code __hash__} — overridden so an unknown value
- *       is not equal to any known member, even if its underlying integer
- *       happens to match one.</li>
- * </ul>
+ * <p>The {@code UnknownEnumMixin} base from smithy-core handles values that
+ * weren't known at generation time: deserializing an unrecognized value (or
+ * filling a missing required member during client error correction, see
+ * {@link MemberErrorCorrectionGenerator}) produces a pseudo-member with
+ * {@code is_unknown} set rather than raising.
  *
  * @see <a href="https://smithy.io/2.0/spec/simple-types.html#intenum">Smithy spec: intEnum</a>
  */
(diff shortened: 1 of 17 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/MemberErrorCorrectionGenerator.java
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Emits the Python expression used to fill a missing required member during client error
+ * correction.
+ *
+ * @see <a href="https://smithy.io/2.0/spec/aggregate-types.html#client-error-correction">Smithy
+ *     spec: Client error correction</a>
+ */
+@SmithyInternalApi
+public final class MemberErrorCorrectionGenerator extends ShapeVisitor.DataShapeVisitor<Boolean> {
```
alexgromero: non-blocking nit: any reason this is `<Boolean>` instead of `<Void>`? The other DataShapeVisitor subclasses (MemberSerializerGenerator, MemberDeserializerGenerator) use `<Void>` and the return value here doesn't seem to be consumed.

alexgromero (commented):
Thanks Yuxuan, mostly looks good! Would be great to see some testing for the new error correction path though.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/StructureGenerator.java
```diff
@@ -375,14 +381,78 @@ def _consumer(schema: Schema, de: ShapeDeserializer) -> None:
...
+    private void generateSmithyDefaultMethod() {
+        for (MemberShape member : requiredMembers) {
+            var target = model.expectShape(member.getTarget());
+            if (!MemberErrorCorrectionGenerator.hasDefault(target)) {
+                return;
+            }
+        }
+        writer.write("""
+                @classmethod
+                def _smithy_default(cls) -> Self:
```
jonathan343, line 464: It seems like we generate `_smithy_default` for every structure, including cases where no generated code can call it, such as structures with no required members or structures that are never used as the target of a required nested structure member.

Could we narrow this so the helper is only emitted when it's needed to create a missing required structure member? This will eliminate some dead code that looks like the following:

```python
@classmethod
def _smithy_default(cls) -> Self:
    return cls()
```

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/MemberErrorCorrectionGenerator.java
```diff
@@ -0,0 +1,212 @@
...
+        writer.writeInline("\"\"");
+        return true;
+    }
+
+    @Override
+    public Boolean intEnumShape(IntEnumShape shape) {
+        // TODO: the Smithy spec recommends intEnum types define an unknown variant. If a
+        //   future change adds an unknown variant to the generated intEnum class (e.g.
+        //   MyIntEnum.unknown(value)), revisit this to emit it instead of the bare 0.
+        writer.writeInline("0");
```
jonathan343: I think the safer short-term fix is to omit client error correction for `enum` and `intEnum` until we have a proper unknown-value representation for them. Longer term, we should add first-class unknown enum/intEnum support and then use that here.

jonathan343 (commented):
Thanks @Alan4506! Let a few comments I think are important to address before merging. +1 to testing as well.

Also, we'll want to patch bump the code generator incase there are not other changes when we release this. Otherwise, our release infra won't generate the new SDK clients.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/MemberErrorCorrectionGenerator.java
Alan4506: Thank you! Addressed in https://github.com/smithy-lang/smithy-python/pull/702/changes/0055ee6ed112e4ddaa2ce9e6a2ec0a06c16b4f41.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/StructureGenerator.java
Alan4506: Thank you! Addressed in https://github.com/smithy-lang/smithy-python/pull/702/changes/0055ee6ed112e4ddaa2ce9e6a2ec0a06c16b4f41.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/MemberErrorCorrectionGenerator.java
Alan4506: In the new commit https://github.com/smithy-lang/smithy-python/pull/702/changes/0055ee6ed112e4ddaa2ce9e6a2ec0a06c16b4f41, I added unknown-variant support to enum/intEnum instead of omitting error correction. If you'd rather keep the PR small I can revert these changes as well!

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
```diff
@@ -35,11 +55,48 @@ public void run() {
...
+                            if isinstance(value, int):
+                                return cls._unknown(value)
+                            return None
+
+                        @property
+                        def is_unknown(self) -> bool:
+                            \"""True if this value was not known at SDK generation time.\"""
+                            return self._name_ not in type(self).__members__
+
+                        def __eq__(self, other: object) -> bool:
```
jonathan343: Can you help me understand why these overrides are needed? This breaks the native contract inherited from `StrEnum` and `IntEnum`. An unknown value no longer compares equal to it's raw string/int, and hash no longer matches. Our main goal here is ensuring forwards compatibility. When these from unknown to know, we should make sure customer code doesn't break

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
```diff
@@ -35,11 +55,48 @@ public void run() {
...
+                    writer.write("$L = $L", name, value);
                     member.getTrait(DocumentationTrait.class).ifPresent(trait -> {
                         writer.writeDocs(trait.getValue(), directive.context());
                     });
                 }
+
+                writer.write("""
+
+                        @classmethod
+                        def _unknown(cls, value: int) -> "Self":
```
jonathan343: By doing this in the codegen layer, this means that we are now copying this code into every enum in a generated client. We should try to avoid this when possible since it adds unnecessary bloat to our clients. Sine this logic is identical for all enums, we should define a mixin class to live in one of our runtime packages:

```python
class UnknownEnumMixin:
    @classmethod
    def _missing_(cls, value):
        ...
```

Then, the generator can just do `class SomeEnum(UnknownEnumMixin, StrEnum):`

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/StructureGenerator.java
```diff
@@ -375,14 +390,99 @@ def _consumer(schema: Schema, de: ShapeDeserializer) -> None:
...
+                    return cls(${C|})
+                """,
+                writer.consumer(w -> writeSmithyDefaultArguments()));
+    }
+
+    /**
+     * Returns true if any structure in the model has a python-required member whose target
+     * is this shape.
+     */
+    private boolean isRequiredStructMemberTarget() {
```
jonathan343: This scans every structure's members for every structure being generated which doesn't scale well. Smithy has `KnowledgeIndex` that precomputes the set of shapes targeted by required members once per model and can turn this into a set lookup

jonathan343 (changes requested):
Hey, let some comments we should address regarding the new approach for handling unknown enums.

Also there is a few hundred newlines generated that seem unnecessary. We should track that down and remove it.

I have a branch used as a POC for the changes I've suggested. It might not be perfect, but it works and may help https://github.com/smithy-lang/smithy-python/commits/feat/client-error-correction-jonathan/

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/IntEnumGenerator.java
Alan4506: Good call! I added these overrides because of a disambiguation concern: if an intEnum has a known `0` member and error correction also fills an unknown `0`, then `status == 0` can't tell them apart (although users are expected to use `is_unknown` to check unknowns, rather than `==`). But I agree the price to address this edge case is probably too high since it breaks the native contract. If you think this is not a big concern then I'll remove the overrides.

Alan4506:
> Hey, let some comments we should address regarding the new approach for handling unknown enums.
> 
> Also there is a few hundred newlines generated that seem unnecessary. We should track that down and remove it.
> 
> I have a branch used as a POC for the changes I've suggested. It might not be perfect, but it works and may help https://github.com/smithy-lang/smithy-python/commits/feat/client-error-correction-jonathan/

Thanks @jonathan343! Addressed your feedback in my new commits.

jonathan343 (approved):
Thanks for addressing my previous comments!

I'm approving since I think this PR is in a good state. However, there is still one concern I think we should follow-up with (here or in a new PR):

Some shapes are used as both responses and requests. When servers omit a required enum member, client error correction will fill in a placeholder (e.g. `EnumClass._corrected("")` or `IntEnumClass._corrected(-1)`). If a response has this type of corrected value, it will be re-serialized in a request as `""` or `-1`. I think we should ideally handle this similar to unknown union members which throw `SerializationError("Unknown union variants may not be serialized.")` in this case. 

I don't think there are any AWS services that model these values, but something we should create an issue for to be addressed as a follow-up.

Alan4506:
> Thanks for addressing my previous comments!
> 
> I'm approving since I think this PR is in a good state. However, there is still one concern I think we should follow-up with (here or in a new PR):
> 
> Some shapes are used as both responses and requests. When servers omit a required enum member, client error correction will fill in a placeholder (e.g. `EnumClass._corrected("")` or `IntEnumClass._corrected(-1)`). If a response has this type of corrected value, it will be re-serialized in a request as `""` or `-1`. I think we should ideally handle this similar to unknown union members which throw `SerializationError("Unknown union variants may not be serialized.")` in this case.
> 
> I don't think there are any AWS services that model these values, but something we should create an issue for to be addressed as a follow-up.

@jonathan343 Thanks! I've opened an issue https://github.com/smithy-lang/smithy-python/issues/722 to track your concern as a follow-up.

PR #706, opened 2026-05-29
Add docstring for HTTPClient header validation contract

Alan4506:
*Description of changes:*
Add docstring for `HTTPClient` to specify: if a request's header fields contain characters prohibited by the HTTP specifications (e.g. CR, LF), it is implementer's responsibility that they must reject the request. The default clients (`AWSCRTHTTPClient`, `AIOHTTPClient`) already satisfy this via their underlying libraries.

### DIFF
--- a/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
+++ b/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
@@ -64,7 +64,14 @@ class HTTPResponse(Response, Protocol):
 
 
 class HTTPClient(ClientTransport[HTTPRequest, HTTPResponse], Protocol):
-    """An asynchronous HTTP client interface."""
+    """An asynchronous HTTP client interface.
+
+    Header field names and values are not validated before reaching this
+    layer. Implementations of ``HTTPClient`` are responsible for this
+    validation: if a request's fields contain characters prohibited by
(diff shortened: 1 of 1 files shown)


packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py
```diff
@@ -64,7 +64,14 @@ def consume_body(self) -> bytes:
 
 
 class HTTPClient(ClientTransport[HTTPRequest, HTTPResponse], Protocol):
-    """An asynchronous HTTP client interface."""
+    """An asynchronous HTTP client interface.
+
+    Header field names and values are not validated before reaching this
+    layer. Implementations of ``HTTPClient`` are responsible for this
+    validation: if a request's fields contain characters prohibited by
+    the HTTP specifications (such as CR or LF), the request MUST be
```
jonathan343: > the HTTP specifications

Why not call out RFC 9110 directly? If someone want's to create an implement of `HTTPClient`, we should make it easy to find which spec we're talking about

Alan4506: Thanks. Addressed in https://github.com/smithy-lang/smithy-python/pull/706/changes/4ba367ec612a9b85efa5af51a590c39efdc47b02!

jonathan343 (approved):

PR #714, opened 2026-06-08
fix(smithy-aws-core): omit Result wrapper for empty awsQuery output

Alan4506:
**Problem:**

The awsQuery protocol returns XML responses wrapped in two layers: an outer `<{OperationName}Response>` element, and inside it an `<{OperationName}Result>` element and a `<ResponseMetadata>` element. For example, Amazon SNS `CreateTopic` returns:

```xml
<CreateTopicResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <CreateTopicResult>
    <TopicArn>arn:aws:sns:us-east-1:111122223333:my-topic</TopicArn>
  </CreateTopicResult>
  <ResponseMetadata>
    <RequestId>...</RequestId>
  </ResponseMetadata>
</CreateTopicResponse>
```

To deserialize this, `AwsQueryClientProtocol._response_wrapper_elements` tells the XML deserializer to descend through both wrapper elements before reading members. The current implementation always returns both:

```python
def _response_wrapper_elements(
    self,
    operation: APIOperation[SerializeableShape, DeserializeableShape],
) -> tuple[str, str]:
    return (
        f"{operation.schema.id.name}Response",
        f"{operation.schema.id.name}Result",
    )
```

The problem is that some operations' output shapes do not have members, so they do not get a `<{OperationName}Result>` element from the service. The response body contains only `<ResponseMetadata>` inside the outer wrapper. Requiring `<{OperationName}Result>` then fails.

This was found while testing Amazon SNS `SetTopicAttributes` and `DeleteTopic` operations. For example:

```xml
<SetTopicAttributesResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <ResponseMetadata>
    <RequestId>...</RequestId>
  </ResponseMetadata>
</SetTopicAttributesResponse>
```

The call failed with:

```
XMLParseError: Error parsing XML: Expected element 'SetTopicAttributesResult', got 'ResponseMetadata'
```

**Description of changes:**

In `_response_wrapper_elements`, only include the `<{OperationName}Result>` wrapper when the operation's output shape has members. Operations with empty output are unwrapped using just the outer `<{OperationName}Response>`. The return type is changed from `tuple[str, str]` to `tuple[str, ...]` since it now returns either one or two wrapper names.

**Testing:**

- Added a unit test in `tests/unit/aio/test_protocols.py` that deserializes an empty-output awsQuery response, and asserts it returns the output shape instead of raising.
- Generated Amazon SNS client locally, and verified its operation worked well: CreateTopic / GetTopicAttributes / SetTopicAttributes / ListTopics / DeleteTopic / Subscribe.

### DIFF
--- a/packages/smithy-aws-core/.changes/next-release/smithy-aws-core-bugfix-05435b3fb9a54b86b4ebb4d61b0d13ae.json
+++ b/packages/smithy-aws-core/.changes/next-release/smithy-aws-core-bugfix-05435b3fb9a54b86b4ebb4d61b0d13ae.json
@@ -1,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Fixed awsQuery response deserialization for operations with no output members."
+}
--- a/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
@@ -364,10 +364,14 @@ class AwsQueryClientProtocol(HttpClientProtocol):
     def _response_wrapper_elements(
         self,
         operation: APIOperation[SerializeableShape, DeserializeableShape],
-    ) -> tuple[str, str]:
+    ) -> tuple[str, ...]:
+        name = operation.schema.id.name
+        # Operations with no output members will omit the <OpResult> element.
+        if not operation.output_schema.members:
+            return (f"{name}Response",)
         return (
-            f"{operation.schema.id.name}Response",
-            f"{operation.schema.id.name}Result",
+            f"{name}Response",
+            f"{name}Result",
         )
 
     def _error_wrapper_elements(self) -> tuple[str, ...]:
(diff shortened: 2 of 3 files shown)


arandito:
Thanks @Alan4506! The fix looks good and makes sense. My only concern is dropping the added unit test and instead using the newly added Smithy protocol test for this exact case: https://github.com/smithy-lang/smithy/pull/3106. The protocol tests should be the source of truth for protocol behavior. Can you confirm the new test is passing?

We have a PR open to upgrade us to smithy 1.71.0 that is blocked because this new test is failing: https://github.com/smithy-lang/smithy-python/pull/703. This fix should unblock that as well.

Alan4506:
> Thanks @Alan4506! The fix looks good and makes sense. My only concern is dropping the added unit test and instead using the newly added Smithy protocol test for this exact case: [smithy-lang/smithy#3106](https://github.com/smithy-lang/smithy/pull/3106). The protocol tests should be the source of truth for protocol behavior. Can you confirm the new test is passing?
> 
> We have a PR open to upgrade us to smithy 1.71.0 that is blocked because this new test is failing: #703. This fix should unblock that as well.

Thanks @arandito! I've removed the unit test and verfied that [the newly added Smithy protocol test](https://github.com/smithy-lang/smithy/pull/3106):
- passed with our fix. 
- failed without our fix.

arandito (approved):
:shipit:

PR #716, opened 2026-06-10
Add API Gateway Accept header customization

Alan4506:
## Context

While generating an API Gateway client from its [model](https://github.com/aws/api-models-aws/blob/main/models/api-gateway/service/2015-07-09/api-gateway-2015-07-09.json) locally and running integration tests, several operations failed to deserialize their responses. For example, `CreateRestApi` returns a `createdDate` member modeled as a timestamp (epoch-seconds under `restJson1`), but the response returned an ISO-8601 string, so timestamp deserialization failed:

```
smithy_json._private.deserializers.JSONTokenError: Error parsing JSON.
Expected token of type `number` at path `createdDate`, but found: `string`: 2026-06-10T17:17:25Z
```

The root cause is API Gateway specific. Unless the client sends `Accept: application/json`, API Gateway responds with `application/hal+json`, in which timestamps are serialized as ISO-8601 strings instead of the epoch-seconds that the `restJson1` protocol expects.

This is a documented API Gateway client requirement, not general protocol behavior. The Smithy spec calls it out as a service-specific customization: [Amazon API Gateway Customizations](https://smithy.io/2.0/aws/customizations/apigateway-customizations.html). Other Smithy SDKs already implement this same customization in their codegen (e.g. smithy-rs, smithy-go, smithy-kotlin). This change adds the equivalent to smithy-python.

## Changes

- **Runtime interceptor** (`smithy-aws-core`): a new `ApiGatewayAcceptHeaderInterceptor` that sets `Accept: application/json` on every request in `modify_before_signing`.
- **Codegen integration** (`codegen/aws/core`): a new `ApiGatewayIntegration`, registered with the `PythonIntegration` SPI. When the service being generated is API Gateway, it emits a plugin that registers the interceptor on the client; for all other services it does nothing. It lives under a new `customizations/apigateway/` package, leaving room for the other service customizations the [Smithy spec](https://smithy.io/2.0/aws/customizations/index.html#) lists (e.g. Glacier, S3, Machine Learning).

Generated clients for the API Gateway service now include an `accept_header.py` plugin that appends the interceptor.

## Testing

- New unit tests (`smithy-aws-core`): verify the interceptor sets `Accept: application/json`, and that it overrides any pre-existing `Accept` header. Both pass.
- Existing unit tests in `smithy-aws-core` still pass.
- Regenerated the API Gateway client and confirmed the generated `client.py` registers the `accept_header_plugin` and that the plugin appends `ApiGatewayAcceptHeaderInterceptor`.
- With the customization in place, the following script (AWS credentials must be set up first) can run without deserialization issue:

```python
import asyncio

from aws_sdk_apigateway.client import AsyncAPIGatewayClient
from aws_sdk_apigateway.config import Config
from aws_sdk_apigateway.models import CreateRestApiInput, DeleteRestApiInput

REGION = "us-east-1"


async def main() -> None:
    client = AsyncAPIGatewayClient(
        config=Config(
            endpoint_uri=f"https://apigateway.{REGION}.amazonaws.com",
            region=REGION,
        )
    )

    # The response includes the `createdDate` timestamp member, which is what
    # fails to deserialize when API Gateway returns application/hal+json.
    created = await client.create_rest_api(
        input=CreateRestApiInput(name="accept-header-demo")
    )
    print(created.id, created.created_date)

    await client.delete_rest_api(input=DeleteRestApiInput(rest_api_id=created.id))


asyncio.run(main())
```

Without the customization this raises `JSONTokenError: Expected token of type 'number' at path 'createdDate', but found: 'string'`. 


packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ApiGatewayAcceptHeaderInterceptor(Interceptor[Any, Any, HTTPRequest, None]):
+    """Sets the Accept header to application/json on API Gateway requests."""
+
+    def modify_before_signing(
+        self, context: RequestContext[Any, HTTPRequest]
+    ) -> HTTPRequest:
+        request = context.transport_request
+        request.fields.set_field(Field(name="Accept", values=["application/json"]))
```
jonathan343: There is an [operation](https://github.com/aws/aws-models/blob/master/apigateway/smithy/model.json#L7170) that has a member which targets the `Accept` header. Looks like it accepts a `application/yaml` value. If a customer provides this and we are unconditionally setting this for them which replaced their requested value. We should check if this headers is set, if it is we don't replace it. If it isn't set, then we set `application/json`. 

Botocore handler for reference: https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L719-L722

packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
```diff
@@ -0,0 +1,18 @@
...
+
+from smithy_core.interceptors import Interceptor, RequestContext
+from smithy_http import Field
+from smithy_http.aio.interfaces import HTTPRequest
+
+
+class ApiGatewayAcceptHeaderInterceptor(Interceptor[Any, Any, HTTPRequest, None]):
+    """Sets the Accept header to application/json on API Gateway requests."""
+
+    def modify_before_signing(
```
jonathan343: Just curious, do other SDKs also do this logic under `modify_before_signing`?

packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
jonathan343, line 1: Can this entire interceptor be written at the Java layer?

Although it's a small amount of code, I think service specific interceptors should live inside the service package. If a customer only uses s3, they shouldn't need to pull in code that doesn't apply to them. 

Again, not a big deal now, but would be nice if we can.

jonathan343 (changes requested):
I ran into this issue while doing my own testing with this client and was glad to see this PR existed!

Overall, the changes look good to me. I have a few suggestions. Let me know if you have any questions!

packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
Alan4506: Not all of them since the other SDKs don't share the same interceptor model. But [smithy-rs](https://github.com/smithy-lang/smithy-rs/blob/main/aws/rust-runtime/aws-inlineable/src/apigateway_interceptors.rs) does this in the same hook: its `AcceptHeaderInterceptor` also sets the header in `modify_before_signing`. This PR mirrors the smithy-rs structure.

packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
Alan4506: Thanks. Addressed in 8662b3cfd2b6795d240e20aa13a157fcbd491e34. I've verfied the generated `accept_header.py` looks like the following:
```python
# Code generated by smithy-python-codegen DO NOT EDIT.

from typing import Any

from smithy_core.interceptors import Interceptor, RequestContext
from smithy_http import Field
from smithy_http.aio.interfaces import HTTPRequest

from .config import Config


class _AcceptHeaderInterceptor(Interceptor[Any, Any, HTTPRequest, None]):
    """Sets the Accept header to application/json if not already present."""

    def modify_before_signing(
        self, context: RequestContext[Any, HTTPRequest]
    ) -> HTTPRequest:
        request = context.transport_request
        if "Accept" not in request.fields:
            request.fields.set_field(Field(name="Accept", values=["application/json"]))
        return request


def accept_header_plugin(config: Config):
    config.interceptors.append(_AcceptHeaderInterceptor())
```

The testing script in the PR description still passes. 

Let me know if you prefer this version - we don't modify `smithy-aws-core` this time.

packages/smithy-aws-core/src/smithy_aws_core/interceptors/api_gateway.py
Alan4506: Thanks!. Addressed in 8662b3cfd2b6795d240e20aa13a157fcbd491e34.

jonathan343 (approved):
LGTM!

PR #717, opened 2026-06-12
Validate Transport Support For Bidirectional Streaming

jonathan343:
### Problem

Only the AWS CRT HTTP client can perform true bidirectional (duplex) streaming in the Python ecosystem, since interleaving request and response data in practice requires HTTP/2. Code generation already defaults services with event streams to `AWSCRTHTTPClient`, but nothing stops a user from overriding `config.transport` with `AIOHTTPClient` or a custom transport that can't do this. When they then invoke a bidirectional operation, the request is sent and the stream appears to hang until it eventually dies with an error such as:  `smithy_core.exceptions.SmithyError: Server disconnected`

Nothing in that error points at the actual problem (the configured transport), so this is painful to debug.

### Solution

This PR makes duplex support an explicit, opt-in transport capability. A new `SUPPORTS_DUPLEX_STREAMING` class attribute on `ClientTransport` defaults to `False`, and `RequestPipeline.duplex_stream` now fails fast with an actionable `UnsupportedTransportError` before any request work happens. `AWSCRTHTTPClient` declares support, `AIOHTTPClient` explicitly does not (it buffers the full response before returning and aiohttp has no HTTP/2 support).

## Example

Invoking a bidirectional operation with a non-duplex transport, with this PR:

```python
config = Config(transport=AIOHTTPClient())
stream = await client.streaming_operation(input=..., config=config)
```

**Without this PR:** the request is sent, the stream stalls, and eventually fails mid-stream with `SmithyError: Server disconnected`.

**With this PR:** the call fails immediately, before anything is sent:

```
smithy_core.exceptions.UnsupportedTransportError: The configured transport (AIOHTTPClient) does not support duplex (bidirectional) event streaming, which is required by the com.example#StreamingOperation operation. Use a transport that does, such as smithy_http.aio.crt.AWSCRTHTTPClient. Custom transports that support duplex streaming must set SUPPORTS_DUPLEX_STREAMING to True.
```

**Why opt-in rather than opt-out?** A survey of the Python HTTP client landscape shows most clients cannot interleave request and response data, so absence of a declaration is treated as "not supported". The pipeline reads the attribute with `getattr(transport, "SUPPORTS_DUPLEX_STREAMING", False)`, so existing custom transports that predate the attribute keep working for everything except duplex operations, which were already broken for them.

**Why fail at call time instead of client construction?** A non-duplex transport is perfectly valid for every other operation on the same client. The capability is only required when a duplex operation is actually invoked, so that is where validation happens.

---

### DIFF
--- a/packages/smithy-core/src/smithy_core/aio/client.py
+++ b/packages/smithy-core/src/smithy_core/aio/client.py
@@ -12,7 +12,12 @@ from .. import URI
 from ..auth import AuthParams
 from ..deserializers import DeserializeableShape, ShapeDeserializer
 from ..endpoints import EndpointResolverParams
-from ..exceptions import ClientTimeoutError, RetryError, SmithyError
+from ..exceptions import (
+    ClientTimeoutError,
+    RetryError,
+    SmithyError,
+    UnsupportedTransportError,
+)
 from ..interceptors import (
     InputContext,
     Interceptor,
@@ -197,6 +202,18 @@ class RequestPipeline[TRequest: Request, TResponse: Response]:
         :param output_event_type: The event type to receive in the output stream.
         :param event_deserializer: The method used to deserialize events.
         """
+        # The transport is assumed not to support duplex streaming unless it
+        # explicitly declares otherwise.
+        if not getattr(self.transport, "SUPPORTS_DUPLEX_STREAMING", False):
+            raise UnsupportedTransportError(
+                f"The configured transport ({type(self.transport).__name__}) does "
+                f"not support duplex (bidirectional) event streaming, which is "
+                f"required by the {call.operation.schema.id} operation. Use a "
+                f"transport that does, such as "
+                f"smithy_http.aio.crt.AWSCRTHTTPClient. Custom transports that "
+                f"support duplex streaming must set SUPPORTS_DUPLEX_STREAMING "
+                f"to True."
+            )
         request_future = Future[RequestContext[I, TRequest]]()
         execute_task = asyncio.create_task(self._execute_request(call, request_future))
         request_context = await request_future
--- a/packages/smithy-core/src/smithy_core/aio/client.py
+++ b/packages/smithy-core/src/smithy_core/aio/client.py
@@ -204,15 +204,15 @@ class RequestPipeline[TRequest: Request, TResponse: Response]:
         """
         # The transport is assumed not to support duplex streaming unless it
         # explicitly declares otherwise.
-        if not getattr(self.transport, "SUPPORTS_DUPLEX_STREAMING", False):
+        if getattr(self.transport, "SUPPORTS_DUPLEX_STREAMING", False) is not True:
             raise UnsupportedTransportError(
                 f"The configured transport ({type(self.transport).__name__}) does "
                 f"not support duplex (bidirectional) event streaming, which is "
                 f"required by the {call.operation.schema.id} operation. Use a "
                 f"transport that does, such as "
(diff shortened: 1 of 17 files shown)


packages/smithy-core/tests/unit/aio/test_client.py
arandito, line 1: Non-blocking: We should definitely look at making a shared testing harness to wire up these mocks rather than needing to create them all inline. I can see cases where we need to test pipeline behavior in the future and will need to recreate this all over again. Doesn't block this PR but something we should consider soon.

arandito (dismissed):
LGTM!

packages/smithy-core/src/smithy_core/aio/client.py
```diff
@@ -197,6 +202,18 @@ async def duplex_stream[
         :param output_event_type: The event type to receive in the output stream.
         :param event_deserializer: The method used to deserialize events.
         """
+        # The transport is assumed not to support duplex streaming unless it
+        # explicitly declares otherwise.
+        if not getattr(self.transport, "SUPPORTS_DUPLEX_STREAMING", False):
```
SamRemis: This is interesting - it's kind of a blind "set this attribute".  Is there a reason we aren't just adding this as a default on the [HTTPClient](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py#L66) class?  Or ifwe don't want to tie it to HTTP, the [ClientTransport protocol](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/src/smithy_core/aio/interfaces/__init__.py#L88)?

Adding a default makes more sense to me than doing a getattr call.  What do you think?

See the [python docs](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-http/src/smithy_http/aio/interfaces/__init__.py#L66) for an example of how to do this (look at the value key in the second example)

jonathan343: This PR does add the default on `ClientTransport`. I kept the `getattr(..., False)` intentionally because `ClientTransport` is a structural `Protocol`, so Python does not enforce that custom transports actually inherit the protocol or have the new attribute at runtime. Existing transports may satisfy the old transport shape but not define `SUPPORTS_DUPLEX_STREAMING`. In that case, missing should mean unsupported, and we still want to raise `UnsupportedTransportError` rather than `AttributeError`.

packages/smithy-core/tests/unit/aio/test_client.py
jonathan343: Addressed in 456627a6544abcfad2e2fe653ce88b46acd35aec

packages/smithy-core/src/smithy_core/aio/client.py
SamRemis: Ahh my bad, didn't read below when I posted this.  Should have completed the review first.

Thank you

SamRemis (dismissed):

arandito (dismissed):

jonathan343:
### Updates

- Rebased this branch onto the latest from develop and resolved all merge conflicts.
- Moved `SUPPORTS_DUPLEX_STREAMING` from the base `ClientTransport` protocol to a dedicated `DuplexClientTransport protocol`, preserving structural compatibility for transports without the capability.
- Duplex support remains opt-in and defaults to false. aiohttp fails fast, while CRT and mock transports explicitly opt in.

@arandito @SamRemis please take another look when you get the chance. Let me know if there are any questions or concerns!

arandito (dismissed):

arandito (approved):

PR #723, opened 2026-06-21
Address async retry refactor review follow-ups

jonathan343:
## Overview 

Minor follow-ups from the async `RetryStrategy` refactor (#671), addressing review feedback we deferred to a separate PR.

## Changes

- Await the async transport `send` and `deserialize` calls in the `designs/retries.md` example.
- Import `RetryStrategyType` from `..retries` in `aio/retries.py` instead of redefining the `Literal` alias.
- Fix the dangling `RetryStrategy` cross-reference in the `RetryToken` docstring, since the protocol moved to `smithy_core.aio.interfaces.retries`.
- Expand the changelog entry to note the module relocations, and add a trailing newline.

---

### DIFF
--- a/designs/retries.md
+++ b/designs/retries.md
@@ -146,14 +146,14 @@ example:
 try:
     retry_token = await retry_strategy.acquire_initial_retry_token()
 except RetryError:
-    transport_response = transport_client.send(serialized_request)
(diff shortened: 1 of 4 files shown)


SamRemis (approved):
Thanks, Jonathan!

PR #725, opened 2026-06-24
fix: generate enum fields as str/int so raw values pass type checkers

Alan4506:
*Description of changes:*

When we generate a Python dataclass for a Smithy structure, each member that targets an enum shape gets a type annotation. Before #702, these fields were annotated as `str | None` (or `int | None` for intEnums), so users could pass either a raw string like `"RETRIEVAL_MODE"` or the enum constant `ChatMode.RETRIEVAL_MODE`: both worked and both passed type checkers.

#702 changed `genericEnum` in `PythonSymbolProvider` to return the enum class symbol directly, so those fields became `ChatMode | None`. This is more precise, but it means passing a raw string (https://github.com/awslabs/aws-sdk-python/pull/54 for example) now fails static type checkers (e.g., `ty`) with errors like:

```
error[invalid-argument-type]: Argument is incorrect
  --> tests/integration/test_bidirectional_streaming.py:36:38
   |
36 |             value=ConfigurationEvent(chat_mode="RETRIEVAL_MODE")
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected `ChatMode | None`, found `Literal["RETRIEVAL_MODE"]`
   |
```

This PR restores the original `str`/`int` field typing while preserving all #702 functionality (error correction, unknown-value handling). 

*Testing:*

Regenerated the `qbusiness` and `connecthealth` clients locally. Verfied their integration tests (https://github.com/awslabs/aws-sdk-python/pull/54 and https://github.com/awslabs/aws-sdk-python/pull/62) still passed, which means the #702 functionality is preserved. Type checker `ty` also passed with no issues.

### DIFF
--- a/codegen/build.gradle.kts
+++ b/codegen/build.gradle.kts
@@ -15,5 +15,5 @@
 
 allprojects {
     group = "software.amazon.smithy.python"
-    version = "0.3.1"
+    version = "0.3.2"
 }
--- a/codegen/build.gradle.kts
+++ b/codegen/build.gradle.kts
@@ -15,5 +15,5 @@
 
 allprojects {
     group = "software.amazon.smithy.python"
-    version = "0.5.0"
+    version = "0.5.1"
 }
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/HttpProtocolTestGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/HttpProtocolTestGenerator.java
@@ -766,10 +766,6 @@ public final class HttpProtocolTestGenerator implements Runnable {
                 writer.writeInline(CodegenUtils.getDatetimeConstructor(writer, parsed));
             } else if (inputShape.isFloatShape() || inputShape.isDoubleShape()) {
                 writer.writeInline("float($L)", node.getValue());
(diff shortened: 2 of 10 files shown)


arandito (commented):
Thanks Yuxuan!

After comparing these changes with #702, I noticed we still have the incorrect behavior in [`HttpProtocolTestGenerator.java`](https://github.com/smithy-lang/smithy-python/pull/702/changes#diff-0d0a5ac7742321aadcb3ac3c480f645f76aa6eb742b0df4b5e663bd6bb307d3c). Should we revert those call sites as well?

Alan4506:
> Thanks Yuxuan!
> 
> After comparing these changes with #702, I noticed we still have the incorrect behavior in [`HttpProtocolTestGenerator.java`](https://github.com/smithy-lang/smithy-python/pull/702/changes#diff-0d0a5ac7742321aadcb3ac3c480f645f76aa6eb742b0df4b5e663bd6bb307d3c). Should we revert those call sites as well?

Thanks @arandito! Yes, they should be reverted as well. Addressed in https://github.com/smithy-lang/smithy-python/pull/725/changes/cde3b5a2b0d70713c1cf4d8086c5bfd1b7aae7de.

arandito (approved):
:shipit:

PR #729, opened 2026-06-24
Add support for parsing and merging AWS config and credentials files

ubaskota:
*Issue #, if available:*

*Description of changes:*
Add support for parsing and merging AWS config and credentials files (`~/.aws/config` and `~/.aws/credentials`) as part of the config resolution pipeline.

Changes:
  - Custom line-by-line parser that handles inline comments, sub-properties, and property continuations
  - Standardizer that normalizes section prefixes, handles [default] vs [profile default] precedence, and separates
  profiles/sso-sessions/services 
  - `ParsedConfigFile` class for querying the merged result
  - Async config loading with path resolution from environment variables and defaults
  - Added support to run conformance tests (`config-file-parser-tests.json` and `config-file-location-tests.json`)+ unit tests for the query layer and file I/O edge cases

Testing:
  - Verified that the newly added and existing tests pass
  - Manually tested against a real `~/.aws/config` file to verify end-to-end parsing behavior

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
@@ -1,0 +1,76 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+from smithy_aws_core.config.file_parser import (
+    FileType,
+    parse_config_file,
+    standardize,
+)
+from smithy_aws_core.config.parsed_config_file import ParsedConfigFile
+
+_DEFAULT_CONFIG_FILE = "~/.aws/config"
+_DEFAULT_CREDENTIALS_FILE = "~/.aws/credentials"
+_CONFIG_FILE_ENV_VAR = "AWS_CONFIG_FILE"
+_CREDENTIALS_FILE_ENV_VAR = "AWS_SHARED_CREDENTIALS_FILE"
+
+
+def resolve_config_paths(
+    config_file_path: str | None = None,
+    credentials_file_path: str | None = None,
+) -> tuple[str, str]:
+    """Resolve the final config and credentials file paths.
+
+    Resolution order for each path:
+    1. Explicit argument (if provided)
+    2. Environment variable (AWS_CONFIG_FILE / AWS_SHARED_CREDENTIALS_FILE)
+    3. Default (~/.aws/config / ~/.aws/credentials)
+
+    The ~ is expanded to the user's home directory.
+
+    :param config_file_path: Override path for config file.
+    :param credentials_file_path: Override path for credentials file.
+    :returns: Tuple of (resolved_config_path, resolved_credentials_path).
+    """
+    config_path = (
+        config_file_path or os.environ.get(_CONFIG_FILE_ENV_VAR) or _DEFAULT_CONFIG_FILE
+    )
+    credentials_path = (
+        credentials_file_path
+        or os.environ.get(_CREDENTIALS_FILE_ENV_VAR)
+        or _DEFAULT_CREDENTIALS_FILE
+    )
+
+    return (
+        os.path.expanduser(os.path.expandvars(config_path)),
+        os.path.expanduser(os.path.expandvars(credentials_path)),
+    )
+
+
+async def load_config(
+    config_file_path: str | None = None,
+    credentials_file_path: str | None = None,
+) -> ParsedConfigFile:
+    """Load and merge AWS config and credentials files.
+
+    Parses both files, standardizes them, and returns a merged
+    ParsedConfigFile ready for querying.
+
+    :param config_file_path: Override path for config file.
+        Defaults to AWS_CONFIG_FILE env var or ~/.aws/config.
+    :param credentials_file_path: Override path for credentials file.
+        Defaults to AWS_SHARED_CREDENTIALS_FILE env var or ~/.aws/credentials.
+    :returns: A ParsedConfigFile with merged profiles from both files.
+    """
+    config_path, credentials_path = resolve_config_paths(
+        config_file_path, credentials_file_path
+    )
+
+    raw_config = await parse_config_file(config_path)
+    raw_credentials = await parse_config_file(credentials_path)
+
+    std_config = standardize(raw_config, FileType.CONFIG)
+    std_credentials = standardize(raw_credentials, FileType.CREDENTIALS)
+
+    return ParsedConfigFile(std_config, std_credentials)
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
@@ -2,13 +2,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+from pathlib import Path
 
 from smithy_aws_core.config.file_parser import (
     FileType,
     parse_config_file,
     standardize,
 )
-from smithy_aws_core.config.parsed_config_file import ParsedConfigFile
+from smithy_aws_core.config.merged_config import MergedConfig
 
 _DEFAULT_CONFIG_FILE = "~/.aws/config"
 _DEFAULT_CREDENTIALS_FILE = "~/.aws/credentials"
@@ -16,10 +17,10 @@ _CONFIG_FILE_ENV_VAR = "AWS_CONFIG_FILE"
 _CREDENTIALS_FILE_ENV_VAR = "AWS_SHARED_CREDENTIALS_FILE"
 
 
-def resolve_config_paths(
-    config_file_path: str | None = None,
-    credentials_file_path: str | None = None,
-) -> tuple[str, str]:
+def _resolve_config_paths(
+    config_file_path: Path | None = None,
+    credentials_file_path: Path | None = None,
+) -> tuple[Path, Path]:
     """Resolve the final config and credentials file paths.
 
     Resolution order for each path:
@@ -34,43 +35,42 @@ def resolve_config_paths(
     :returns: Tuple of (resolved_config_path, resolved_credentials_path).
     """
     config_path = (
-        config_file_path or os.environ.get(_CONFIG_FILE_ENV_VAR) or _DEFAULT_CONFIG_FILE
-    )
+        config_file_path
+        or Path(os.environ.get(_CONFIG_FILE_ENV_VAR, _DEFAULT_CONFIG_FILE))
+    ).expanduser()
+
     credentials_path = (
         credentials_file_path
-        or os.environ.get(_CREDENTIALS_FILE_ENV_VAR)
-        or _DEFAULT_CREDENTIALS_FILE
-    )
+        or Path(os.environ.get(_CREDENTIALS_FILE_ENV_VAR, _DEFAULT_CREDENTIALS_FILE))
+    ).expanduser()
 
-    return (
-        os.path.expanduser(os.path.expandvars(config_path)),
-        os.path.expanduser(os.path.expandvars(credentials_path)),
-    )
+    return config_path, credentials_path
 
 
 async def load_config(
-    config_file_path: str | None = None,
-    credentials_file_path: str | None = None,
-) -> ParsedConfigFile:
+    config_file_path: Path | None = None,
+    credentials_file_path: Path | None = None,
+) -> MergedConfig:
     """Load and merge AWS config and credentials files.
 
     Parses both files, standardizes them, and returns a merged
-    ParsedConfigFile ready for querying.
+    MergedConfig ready for querying.
 
     :param config_file_path: Override path for config file.
         Defaults to AWS_CONFIG_FILE env var or ~/.aws/config.
     :param credentials_file_path: Override path for credentials file.
         Defaults to AWS_SHARED_CREDENTIALS_FILE env var or ~/.aws/credentials.
-    :returns: A ParsedConfigFile with merged profiles from both files.
+    :returns: A MergedConfig with merged profiles from both files.
     """
-    config_path, credentials_path = resolve_config_paths(
+
+    config_path, credentials_path = _resolve_config_paths(
         config_file_path, credentials_file_path
     )
 
-    raw_config = await parse_config_file(config_path)
-    raw_credentials = await parse_config_file(credentials_path)
+    raw_config = await parse_config_file(str(config_path))
+    raw_credentials = await parse_config_file(str(credentials_path))
 
     std_config = standardize(raw_config, FileType.CONFIG)
     std_credentials = standardize(raw_credentials, FileType.CREDENTIALS)
 
-    return ParsedConfigFile(std_config, std_credentials)
+    return MergedConfig(std_config, std_credentials)
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/exceptions.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/exceptions.py
@@ -1,0 +1,8 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Config related exceptions"""
+
+
+class ConfigParseError(Exception):
+    """Raised when a config file cannot be parsed due to invalid syntax."""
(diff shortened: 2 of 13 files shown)


packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        config_file_path or os.environ.get(_CONFIG_FILE_ENV_VAR) or _DEFAULT_CONFIG_FILE
+    )
+    credentials_path = (
+        credentials_file_path
+        or os.environ.get(_CREDENTIALS_FILE_ENV_VAR)
+        or _DEFAULT_CREDENTIALS_FILE
+    )
+
+    return (
+        os.path.expanduser(os.path.expandvars(config_path)),
```
SamRemis: [Trying to reason about this, genuine question]

This method strikes me as a little odd.  If someone passes both the credentials and config path, we don't take them literally and just expand the user for them.  

I don't know if this needs to be a public method.  We could have all of this logic inline in load_config, or make it private, no?

SamRemis: [change requested]

To keep the public facing interface smaller, could we move this logic inline to load_config?

Also, for now can we get rid of the expandvars?  We shouldn't expand environment vars if other SDKs don't, and I don't see it in the specification.

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,360 @@
...
+        sso_sessions=sso_sessions,
+        services=services,
+    )
+
+
+async def _read_file(path: str) -> str | None:
+    """Read file content asynchronously.
+
+    Returns None if the file doesn't exist or can't be opened.
+    Per the SEP: inaccessible files are treated as empty.
```
SamRemis: [Change requested]
There's a couple places we reference "SEP", which we shouldn't do in public code.  Can you reword this?

packages/smithy-aws-core/src/smithy_aws_core/config/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        os.path.expanduser(os.path.expandvars(credentials_path)),
+    )
+
+
+async def load_config(
```
SamRemis, line 50: [discussion]

While I like this as an entry point, there's a few things I'd want to consider changing, but you're more into this than me so maybe I'm missing something. 

First, the `ParsedConfigFile` is a strange name for an object that holds two parsed files (config + credentials).  Maybe we should rename to `MergedConfig`? 

Second (and this one is more spitballing), how would you feel about accepting a path object instead of a string so we don't need to normalize?

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,360 @@
...
+        content = await asyncio.to_thread(Path(path).read_text, encoding="utf-8")
+        return content
+    except (FileNotFoundError, PermissionError, OSError, UnicodeDecodeError):
+        return None
+
+
+def parse_content(content: str) -> RawParsedSections:
+    """Parse config file content from a string into raw sections.
+
+    Note: This function is public only for direct use by unit tests
```
SamRemis: [Still thinking about how to fix this one, but commenting cause I think it's right]

If this should be internal, we should make it an internal method.  If that makes it harder to test, that's a sign that we should reconsider our testing strategy.  This probably looks like writing to temp files as the config files for testing.

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,360 @@
...
+
+def _parse_section_line(line: str, line_num: int) -> str | None:
+    """Parse a section definition line like [profile foo]."""
+    stripped = line.strip()
+    bracket_end = stripped.find("]")
+    if bracket_end == -1:
+        raise ConfigParseError(f"Line {line_num}: Section definition must end with ']'")
+    inner = stripped[1:bracket_end].strip()
+    if not inner:
+        raise ConfigParseError(f"Line {line_num}: Section name cannot be empty")
```
SamRemis, line 225: [Discussion for posterity, no change requested]

I feel weird commenting this because I like what you did here _better_ than the specification, but the spec does say that this is wrong for sso-session profiles: 
> If the session name is empty, the section and all following properties must be ignored.

Service profiles have a similar line:
> If the services configuration name is empty, the section and all following properties MUST be ignored.

I'm not asking you to change this for two reasons:
1. If we ever want to conform with the specification, we can always come back and update this code
2. It's a better customer experience to let them know we are ignoring their profile's content.

Maybe there's something I'm missing as to why that was part of the spec, but for now, let's leave it as is unless you (or anyone else) disagrees.  Also, depending on how you read the word "ignored", we could consider raising an error ignoring this since those values would never be used...

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,360 @@
...
+import os
+import re
+from dataclasses import dataclass, field
+from enum import Enum
+from pathlib import Path
+
+from smithy_aws_core.config.exceptions import ConfigParseError
+
+# Type aliases
+type PropertyValue = str | dict[str, str]
```
SamRemis: [Discussion]

Instead of having `ProfileProperties`, a ProfileMap could have a `Profile`; each `Profile` could have `ScalarProperties` and `GroupedProperties`.  The union feels a bit awkward to have to consider while coding around the `StandardizedOutput` object.

What do you think?

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,360 @@
...
+async def _read_file(path: str) -> str | None:
+    """Read file content asynchronously.
+
+    Returns None if the file doesn't exist or can't be opened.
+    Per the SEP: inaccessible files are treated as empty.
+    """
+    try:
+        content = await asyncio.to_thread(Path(path).read_text, encoding="utf-8")
+        return content
+    except (FileNotFoundError, PermissionError, OSError, UnicodeDecodeError):
```
SamRemis: [Suggestion]

For usability, maybe we should add a warning on `PermissionError` and `OSError` so the user knows that we couldn't read from the requested file.  Also, a UnicodeDecodeError should raise and not silently fail.

Thoughts?

SamRemis (changes requested):

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,375 @@
...
+
+    :param raw_sections: Raw sections from parse_config_file().
+    :param file_type: Whether this is a config or credentials file.
+    :returns: StandardizedOutput with profiles, sso_sessions, and services.
+    """
+    profiles: ProfileMap = {}
+    sso_sessions: ProfileMap = {}
+    services: ProfileMap = {}
+
+    has_profile_default = False
```
SamRemis, line 84: [discussion/suggestion]

It feels a bit odd to have a method that has so many if statements gated behind the filetype flag.  How would you feel about getting rid of that flag and instead having `standardize_config_file` and `standardize_credentials_file` methods similar to the parser methods?

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,375 @@
...
+    - [default] vs [profile default] precedence
+    - Invalid profile/property name filtering
+    - Duplicate profile merging
+    - Separating profiles, sso-sessions, and services
+
+    :param raw_sections: Raw sections from parse_config_file().
+    :param file_type: Whether this is a config or credentials file.
+    :returns: StandardizedOutput with profiles, sso_sessions, and services.
+    """
+    profiles: ProfileMap = {}
```
SamRemis: [nit]

A better name for this would be `SectionMap` or similar since profiles is only the first of these three

packages/smithy-aws-core/tests/unit/config/test_merged_config.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def test_returns_none_for_missing(self):
+        cf = _make_config_file(config_profiles={})
+        assert cf.get_profile("nonexistent") is None
+
+
+class TestMerge:
+    """Tests for credentials/config merge behavior."""
+
+    def test_credentials_override_config_for_duplicate_key(self):
```
SamRemis, line 121: Isn't this assertion already covered by `test_credentials_override_config`?

packages/smithy-aws-core/tests/unit/config/test_config_file_io.py
```diff
@@ -0,0 +1,100 @@
...
+class TestFileIssues:
+    """Tests that missing files are treated as empty."""
+
+    @pytest.mark.asyncio
+    async def test_config_not_found_returns_empty(self, tmp_path: Path):
+        nonexistent = str(tmp_path / "does_not_exist")
+        result = await parse_config_file(nonexistent)
+        assert result == {}
+
+    @pytest.mark.asyncio
```
SamRemis: This test is a superset of the test above it; we should just call parse in this test and make it all one test IMO.

SamRemis (approved):

SamRemis (changes requested):
My bad, clicked approve by mistake.  Only the test comments are blocking

packages/smithy-aws-core/tests/unit/config/test_merged_config.py
SamRemis: Also covered by the JSON "Duplicate properties between files uses credentials property" test

ubaskota: I deleted one.

packages/smithy-aws-core/tests/unit/config/test_config_file_io.py
ubaskota: I decided to remove `test_config_not_found_returns_empty`

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
ubaskota: updated

SamRemis (approved):

packages/smithy-aws-core/src/smithy_aws_core/config/merged_config.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        merged: SectionMap = {}
+
+        for name, profile in config_profiles.items():
+            merged[name] = Section(
+                properties=dict(profile.properties),
+                sub_properties={k: dict(v) for k, v in profile.sub_properties.items()},
+            )
+
+        for name, profile in credentials_profiles.items():
```
jonathan343: Could we represent each section with a single standardized mapping? Storing scalar and grouped values separately allows the same property to hold conflicting values after merging.

For example:

```ini
# ~/.aws/config
[profile p]
x = config
```

```ini
# ~/.aws/credentials
[p]
x =
  nested = credentials
```

Running this through the current implementation produces:

```python
Section(
    properties={"x": "config"},
    sub_properties={"x": {"nested": "credentials"}},
)
```

The credentials value should replace the config value entirely. Botocore’s final merged profile is:

```
{'x': {'nested': 'credentials'}}
```

Credentials should replace the complete `x` property, so `get("p", "x")` should return `None`. Botocore produces `{"x": {"nested": "credentials"}}` for this case. A single mapping such as `dict[str, str | dict[str, str]]` would prevent this conflicting state and make whole-property precedence straightforward.

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,375 @@
...
+    profiles: SectionMap,
+) -> None:
+    stripped = section_name.strip()
+
+    if stripped.startswith("profile"):
+        return
+    if stripped.startswith("sso-session"):
+        return
+    if stripped.startswith("services"):
+        return
```
jonathan343: These checks incorrectly discard valid credentials profiles such as:

```ini
[services-prod]
aws_access_key_id = key
```

Could we remove the three `startswith()` checks and rely directly on `_is_valid_identifier()` before merging the profile? This is simpler, matches botocore, and avoids dropping valid names that merely begin with a reserved term.

jonathan343 (changes requested):
Thanks Ujjwal. Commenting on a few edge cases that seem like bugs to me. Still getting through others, but this is what I have so far.

packages/smithy-aws-core/src/smithy_aws_core/config/exceptions.py
```diff
@@ -0,0 +1,8 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Config related exceptions"""
+
+
+class ConfigParseError(Exception):
```
jonathan343, line 7: *This is a general error comment, not specific to this file*

---

Parse errors should identify which source file failed. `load_config()` parses both the config and credentials files, but the current exception only includes the line number.

For example, using an unclosed section header in either file:

```
[profile p
```

produces the same error in both cases:

```
ConfigParseError: Line 1: Section definition must end with ']'
```

Botocore distinguishes the files:

```
ConfigParseError: Unable to parse config file: /.../config
ConfigParseError: Unable to parse config file: /.../credentials
```

Could `parse_config_file()` add `file_path` when re-raising errors from `_parse_content()`? This would preserve the current line-specific detail while identifying the file:

```
Unable to parse config file '/.../credentials':
Line 1: Section definition must end with ']'
```

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,375 @@
...
+    bracket_end = stripped.find("]")
+    if bracket_end == -1:
+        raise ConfigParseError(f"Line {line_num}: Section definition must end with ']'")
+    inner = stripped[1:bracket_end].strip()
+    if not inner:
+        raise ConfigParseError(f"Line {line_num}: Section name cannot be empty")
+    return inner
+
+
+def _parse_property_line(line: str, line_num: int) -> tuple[str | None, str]:
```
jonathan343, line 229: Ignoring an invalid property currently allows its continuation lines to modify the previous valid property.

For example:

```ini
[profile p]
region = us-east-1
invalid key = ignored
  continuation
```

produces:

```python
Section(
    properties={"region": "us-east-1\ncontinuation"},
    sub_properties={},
)
```

The expected result is:

```
Section(
    properties={"region": "us-east-1"},
    sub_properties={},
)
```

Botocore does not filter the invalid property name, but it correctly associates continuation with invalid key rather than region.

Could we track when an invalid property has been ignored and also discard its continuation lines? This preserves our spec's silent-ignore behavior without corrupting the previous property.

packages/smithy-aws-core/src/smithy_aws_core/config/merged_config.py
ubaskota: Yup, addressed.

packages/smithy-aws-core/src/smithy_aws_core/config/file_parser.py
```diff
@@ -0,0 +1,371 @@
...
+
+    :param content: The raw config file content as a string.
+    :returns: Raw sections dict {section_name: {key: value}}.
+    :raises ConfigParseError: If the content has invalid syntax.
+    """
+    sections: RawParsedSections = {}
+    current_section: str | None = None
+    current_key: str | None = None
+    in_sub_property: bool = False
+    ignore_continuation: bool = False
```
jonathan343, line 142: Thanks, this fixes the original example, but it still fails when the invalid property is the first property in the section:

```ini
[profile p]
invalid key = ignored
  continuation
```

This currently raises:

```
ConfigParseError: Line 3: Expected a property definition, found continuation
```

`current_key` is checked before `ignore_continuation`, so the continuation is only ignored when an earlier valid property populated `current_key`. Could we check `ignore_continuation` first and reset it when entering a new section? It would also be useful to test this case without a preceding valid property.

jonathan343 (changes requested):
Thanks Ujjwal, this pretty much looks good to me. Had one follow up from a previews review and a comment on the missing changelog entry. Can you add a feature entry to smithy-aws-core with something like below:

> Add utilities for parsing and merging shared AWS configuration and credentials files. These utilities are not yet integrated into generated clients.

jonathan343 (approved):

PR #733, opened 2026-07-02
Implement updated standard retry behavior

Alan4506:
*Description of changes:*

Updates the standard retry behavior:

- **Throttling backoff:** throttling errors use a larger base backoff (1000ms) than other errors (50ms), based on `error.is_throttling_error`.
- **Retry quota:** `RETRY_COST` is now 14, throttling retries cost 5, and the timeout-specific cost is removed.
- **DynamoDB defaults:** DynamoDB / DynamoDB Streams clients default to 4 max attempts and a 25ms base backoff, applied per-value (a value configured by the customer always wins). The generic runtime stays AWS-agnostic; the DynamoDB knowledge lives in an AWS codegen integration.
- **Long-polling:** long-polling operations back off before giving up even when the quota is exhausted. Detection checks the `smithy.api#longPoll` trait at runtime through `APIOperation.long_polling`.
- **`x-amz-retry-after`:** honored when returned by the service, capped at 5 seconds above the normal backoff; invalid/missing values fall back to normal backoff and the standard HTTP `Retry-After` is ignored.

*Note on long-polling detection:*

Some service models don't apply the `smithy.api#longPoll` trait yet. We deliberately do not hard-code those operations. Doing this would require codegen interface changes for temporary behavior. The trait lookup is the long-term design regardless. Until those service models apply the trait, those operations won't get long-polling backoff.

*Testing:*

Added unit tests for each change (throttling vs non-throttling backoff, quota costs, the quota-exhaustion backoff contract, and `x-amz-retry-after` parsing/clamping/wiring). Also updated affected functional tests. All Python checks pass (`make check-py`, `make test-py`) and the Java codegen builds successfully.

*Follow-up:*

The awsJson protocol's `_create_error` is not on `develop` yet (it lives on the [json-rpc branch](https://github.com/smithy-lang/smithy-python/tree/clean-json-rpc-v1)). Once that merges, it needs the same `parse_retry_after(response)` wiring that awsQuery already has, so `x-amz-retry-after` is honored for awsJson services (e.g. DynamoDB) too.


codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsLongPollingIntegration.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Marks the known long-polling operations so the generic client generator
+ * applies long-polling retry behavior to them.
+ *
+ * <p>These operations are hard-coded until the {@code aws.api#longPoll} trait
+ * ships in service models. Once it ships, this can check for the trait instead.
+ */
+public final class AwsLongPollingIntegration implements PythonIntegration {
+
+    private static final Map<String, Set<String>> LONG_POLLING_OPERATIONS = Map.of(
```
SamRemis: [Question, maybe blocking?]

Does it really make sense for us to do this now before the trait is ready?  I'd rather not have to track the separate work of updating this generator to function based on the trait and just update it when the rest of the SDKs do.  

If we are already committed to shipping this, it's fine, but it is extra effort on our part for no customer benefit.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -240,7 +258,8 @@ private void writeSharedOperationInit(
                     auth_scheme_resolver=config.auth_scheme_resolver,
                     supported_auth_schemes=config.auth_schemes,
                     endpoint_resolver=config.endpoint_resolver,
-                    retry_strategy=retry_strategy,
+                    retry_strategy=retry_strategy,${?isLongPolling}
+                    is_long_polling=True,${/isLongPolling}
```
SamRemis: [blocking, change requested]

This information is retry specific.  Plumbing it through every operation call works fine in this instance, but this type of solution isn't scalable.  This method signature would grow _very_ quickly if we needed to update the pipeline method's signature for every branching decision that we make as part of the pipeline call.  

We should talk about the right way to plumb through the information.  We already have the `APIOperation` object in scope.  Since this will eventually be a modeled trait, doesn't it make more sense for us to add it to the operation object as opposed to growing the pipeline method's signature?

packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -147,23 +203,41 @@ def __init__(
...
 
         :param token_scope: This argument is ignored by this retry strategy.
+        :param is_long_polling: Whether the operation is a long-polling operation.
+            Long-polling operations back off before returning even when the retry
+            quota is exhausted.
         """
         retry_delay = self.backoff_strategy.compute_next_backoff_delay(0)
-        return StandardRetryToken(retry_count=0, retry_delay=retry_delay)
+        return StandardRetryToken(
+            retry_count=0, retry_delay=retry_delay, is_long_polling=is_long_polling
```
SamRemis: [Change requested, blocking but open to discussion]

The concept of a token represents the state of a given attempt.  This concept is unrelated to whether or not the operation is used for long polling.  

Plumbing this through the retry token means that `refresh_retry_token_for_retry` needs to copy it forward unchanged.  This is more boilerplate and more complex than it needs to be.  

I'm trying to think of a better alternative.  Off the top of my head, we could add the operation's schema to these calls, maybe via a retries_context like concept?  Or just put this long_polling in a context so that it can grow without growing the method signature?

packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -22,8 +22,31 @@ class RetryStrategyResolver:
...
+    ):
+        """Initialize the resolver.
+
+        :param default_max_attempts: The maximum number of attempts to use when the
+            customer did not configure one. Only applies to standard mode.
+        :param default_backoff_scale: The base backoff scale in seconds used for
+            non-throttling errors when the customer did not configure one. Only applies
+            to standard mode.
+        """
+        self._default_max_attempts = default_max_attempts
```
SamRemis: [Question/discussion]

Did you consider making this a service-level interceptor instead of passing it in here?  The interceptor could set these values if they're unset and only apply to DDB.  I think that's more inline with the intention of the library's original authors.

You may need to wait until we have provenance tracking to support this though.

SamRemis (dismissed):

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -29,6 +30,30 @@
 from ..serializers import HTTPRequestSerializer
 from .interfaces import HTTPErrorIdentifier, HTTPRequest, HTTPResponse
 
+_LOGGER = logging.getLogger(__name__)
+
+_RETRY_AFTER_HEADER = "x-amz-retry-after"
```
SamRemis: [Blocking]

We should not add aws-specific logic to smithy-http or other generic non-AWS specific code.  We need to find a way to hook into this from the AWS specific code.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsLongPollingIntegration.java
Alan4506: After a deeper look into other SDKs, I think it makes more sense to keep long polling in this PR, implemented as trait-first with a hardcoded fallback:

```java
if (operation.hasTrait(LongPollTrait.class)) {
    return true;
}
return HARDCODED_LONG_POLLING_OPERATIONS...  // SQS/SFN/SWF
```

- The `smithy.api#longPoll` trait shipped in Smithy 1.69 (smithy-lang/smithy#3019), so `hasTrait(LongPollTrait)` compiles and works today. The service models don't carry the trait yet, so it currently falls through to the hardcoded table.
- This trait-first + fallback method matches some other Smithy SDKs, such as Go, Kotlin, and Swift. They all check the trait first and fall back to a hardcoded table.
- Once the service models carry the trait, we can just delete the fallback table.

packages/smithy-http/src/smithy_http/aio/protocols.py
Alan4506: Fixed. The `x-amz-retry-after` parsing moved to `smithy_aws_core.utils.parse_retry_after`. smithy-http now only has a neutral `_retry_after` hook that returns `None`, which the AWS protocols override.

packages/smithy-core/src/smithy_core/aio/retries.py
Alan4506: Good call. I agree that a service-level interceptor can be a reasonable refactor in the future, once we have provenance tracking. I can make a follow-up for it.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
Alan4506: Good call. Dropped `is_long_polling` from the pipeline signature entirely. The long-polling flag is now carried on the operation `context` (`smithy_core.retries.LONG_POLLING`, a `PropertyKey[bool]`) instead of a dedicated parameter. The generator only emits the key on operations that need it.

packages/smithy-core/src/smithy_core/aio/retries.py
Alan4506: Agreed, and this is the same change as above. `StandardRetryToken.is_long_polling` is gone, so `refresh_retry_token_for_retry` no longer copies it forward. `acquire_initial_retry_token` and `refresh_retry_token_for_retry` now take an optional `context`. They read `LONG_POLLING` from it to decide whether the operation should still back off when the retry quota is exhausted.

packages/smithy-aws-core/.changes/next-release/smithy-aws-core-feature-881860b5bda049d1b000983dd2a3bd0e.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "feature",
```
arandito: Why does this need to be a minor bump?

packages/smithy-http/src/smithy_http/aio/protocols.py
arandito, line 1: Looks like `smithy-http` is missing a changelog

packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -123,18 +132,42 @@ def __deepcopy__(self, memo: Any) -> "SimpleRetryStrategy":
...
+    _MAX_BACKOFF: float = 20
+    """Upper bound (seconds) for the computed backoff, applied before jitter."""
+
     def __init__(
         self,
         *,
         backoff_strategy: retries_interface.RetryBackoffStrategy | None = None,
+        throttling_backoff_strategy: retries_interface.RetryBackoffStrategy
+        | None = None,
+        default_backoff_scale: float | None = None,
```
arandito: Why do we need `default_backoff_scale` to be passed directly for overrides? It's a bit awkward because callers can now set it either with the new parameter or through `backoff_strategy.backoff_scale_value`.  If they set both, `default_backoff_scale` is silently dropped.

The backoff value is part of backoff strategies so it makes sense to define it in the original abstraction. It may also cause confusion on whether its the default for both non-throttling and throttling errors; right now its just the default for non-throttling.

What do you think about making the DynamoDB plugin construct an instance of `ExponentialRetryBackoffStrategy` with a 25 ms backoff value and passing it in as the `backoff_strategy` value? The trade off is the possibility that DynamoDB's `max_backoff` and `jitter_type` drift in the future, but I think this keeps the retry strategy user interface cleaner and more intuitive.

packages/smithy-core/src/smithy_core/aio/retries.py
```diff
@@ -147,20 +180,37 @@ def __init__(
...
+                max_backoff=self._MAX_BACKOFF,
+                jitter_type=ExponentialBackoffJitterType.FULL,
+            )
+        )
         self.max_attempts = max_attempts
         self._retry_quota = retry_quota or StandardRetryQuota()
 
     async def acquire_initial_retry_token(
-        self, *, token_scope: str | None = None
+        self, *, token_scope: str | None = None, context: TypedProperties | None = None
```
arandito: If `context` is not used within `acquire_initial_retry_token` for the standard or simple strategies, why did we add it to the base protocol?

packages/smithy-core/tests/unit/aio/test_retries.py
```diff
@@ -96,14 +102,160 @@ async def test_standard_retry_does_not_retry(error: Exception | CallError) -> No
...
+    )
+    error = CallError(is_retry_safe=True, retry_after=10.0)
+    token = await strategy.acquire_initial_retry_token()
+    token = await strategy.refresh_retry_token_for_retry(
+        token_to_renew=token, error=error
+    )
+    assert token.retry_delay == 6.0
+
+
+async def test_standard_non_throttling_uses_default_backoff_scale() -> None:
```
arandito, line 145: I don't think this test actually exercises the default backoff behavior as it claims. It just returns the backoff strategy's `backoff_scale_value` which in this case is configured to 0.05. Should we instead omit the `backoff_strategy` and confirm the default?

packages/smithy-core/tests/unit/aio/test_retries.py
```diff
@@ -96,14 +102,160 @@ async def test_standard_retry_does_not_retry(error: Exception | CallError) -> No
...
+    error = CallError(is_retry_safe=True, is_throttling_error=False)
     token = await strategy.acquire_initial_retry_token()
     token = await strategy.refresh_retry_token_for_retry(
         token_to_renew=token, error=error
     )
-    assert token.retry_delay == 5.5
+    assert token.retry_delay == pytest.approx(0.05)  # type: ignore
+
+
+async def test_standard_throttling_uses_throttling_backoff_scale() -> None:
```
arandito, line 156: Same here, should we be testing the behavior when `throttling_backoff_strategy` is not set?

packages/smithy-core/tests/unit/aio/test_retries.py
```diff
@@ -166,3 +318,14 @@ async def test_retry_strategy_resolver_rejects_invalid_type() -> None:
...
+
+async def test_resolver_no_service_defaults_uses_strategy_defaults() -> None:
+    resolver = RetryStrategyResolver()
+
+    strategy = await resolver.resolve_retry_strategy(retry_strategy=None)
+
+    assert isinstance(strategy, StandardRetryStrategy)
+    assert strategy.max_attempts == 3
+    delay = strategy.backoff_strategy.compute_next_backoff_delay(1)
+    assert 0 <= delay <= 0.05
```
arandito, line 317: Should we also confirm the throttling strategy default backoff delay here?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -256,11 +266,14 @@ private void writeSharedOperationInit(
                     protocol=config.protocol,
                     transport=config.transport
                 )
-                call = $4T(
+                ${?isLongPolling}operation_context = $4T({"config": config})
```
arandito: Building on Sam's [comment](https://github.com/smithy-lang/smithy-python/pull/733#discussion_r3538504160), I still think we can improve how we pass this information into the `ClientCall`.

The trait should be detectable from `APIOperation` without needing to code generate any context. We already have precedence for operation trait detection for the [idempotency token trait](https://github.com/smithy-lang/smithy-python/blob/5c25af99413f0ff404437e66da1aaaeb73ae0347/packages/smithy-core/src/smithy_core/schemas.py#L310) plus streaming traits.

We also already have a reference to an `APIOperation` object inside of `RequestPipeline._retry` through `call.operation`.

We can do something like this:
```
except RetryError as retry_error:
    if call.operation.long_polling and retry_error.retry_after is not None:
        await sleep(retry_error.retry_after)
    raise output_context.response
```
I think we may need to unconditionally set `retry_error.retry_after` for all quota exhaustion errors to avoid passing in the operation into `refresh_retry_token_for_retry`. Since we are already adding `retry_after` to the `RetryError` contract, I think that's a reasonable tradeoff.

arandito (commented):
Thanks @Alan4506! I left a couple comments. My main concerns related to how we are detecting long polling operations and simplifying the `StandardRetryStrategy` constructor. Let me know if you have any questions!

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
Alan4506: Good call! I agree that this aligns with our existing runtime design, so I've addressed this in my latest commit. And I've removed the temporary hard-coded fallback for services that don't apply the trait, since it would introduce codegen interface changes for temporary behavior.

packages/smithy-aws-core/.changes/next-release/smithy-aws-core-feature-881860b5bda049d1b000983dd2a3bd0e.json
Alan4506: I think enhancement is more appropriate. Addressed in my latest commit. Let me know if you think otherwise!

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDynamoDbRetryIntegration.java
```diff
@@ -64,7 +64,10 @@ and source_of("max_attempts") == $4T.DEFAULT
                         if max_attempts is not None
                         else _DYNAMODB_DEFAULT_MAX_ATTEMPTS
                     ),
-                    default_backoff_scale=_DYNAMODB_DEFAULT_BACKOFF_SCALE,
+                    backoff_strategy=$5T(
```
arandito, line 68: nit: can we explicitly set `max_backoff` to a constant that equals 20? in case the underlying default changes, we want to make that explicit here.

packages/smithy-core/src/smithy_core/exceptions.py
```diff
@@ -90,9 +91,9 @@ class DiscriminatorError(SmithyError):
 class RetryError(SmithyError):
     """Base exception type for all exceptions raised in retry strategies.
 
-    :param retry_after: An optional delay in seconds that the caller should wait before
-        giving up on retries. Long-polling operations use this to back off even when the
-        retry quota is exhausted.
+    :param retry_after: An optional delay in seconds that would have applied to the
```
arandito: tiny nit : "would have applied" --> "could apply"

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsDynamoDbRetryIntegration.java
```diff
@@ -0,0 +1,140 @@
...
+                        backoff_scale_value=_DYNAMODB_DEFAULT_BACKOFF_SCALE,
+                        jitter_type=$6T.FULL,
+                    ),
+                )
+            """;
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins(GenerationContext context) {
+        final String pluginFile = "retry";
```
arandito: nit: to stay consistent with our runtime libraries, can we call this `retries.py`.

packages/smithy-core/src/smithy_core/traits.py
```diff
@@ -208,6 +208,20 @@ def __post_init__(self):
         assert self.document_value is None
 
 
+@dataclass(init=False, frozen=True)
+class LongPollTrait(Trait, id=ShapeID("smithy.api#longPoll")):
+    """Indicates that the service may hold the request open while waiting for
+    information to become available."""
+
```
arandito, line 215: nit: Let's move the assertions to a `__post_init__` method to stay consistent with other traits. See [XMLNamespaceTrait](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/src/smithy_core/traits.py#L368) for a reference. This ensures it fails at initialization not when we access the property.

arandito (commented):
LGTM :shipit:

arandito (approved):

PR #737, opened 2026-07-09
Add `Async` prefix to generated AWS client names

jonathan343:
## Summary

Generated AWS clients are async, but they're named `<SdkId>Client` (e.g. `BedrockRuntimeClient`). This PR renames them to `Async<SdkId>Client` (e.g. `AsyncBedrockRuntimeClient`) to match the convention and to free up the unprefixed name for the sync clients we plan to add later.

The rename is a breaking change for the clients already published under the old name. To keep existing code working, those clients also emit the old name as a deprecated alias that resolves to the async client and raises a `DeprecationWarning`. The alias is temporary and will be removed in a minor bump sometime in the future.

## Scope

The Async rename applies to all generated AWS clients. The backwards-compat alias is only generated for the clients already published under the unprefixed name in [awslabs/aws-sdk-python](https://github.com/awslabs/aws-sdk-python/tree/main/clients):

Bedrock Runtime, ConnectHealth, Lex Runtime V2, Polly, QBusiness, SageMaker Runtime HTTP2, Transcribe Streaming

New clients released after this change is merged are created with the Async-prefixed and no alias.

## Before

Client Class:
```python
class BedrockRuntimeClient
    ...
```

Usage (no warning):
```
>>> from aws_sdk_bedrock_runtime.client import BedrockRuntimeClient
>>> BedrockRuntimeClient()
<aws_sdk_bedrock_runtime.client.AsyncBedrockRuntimeClient object at 0x102ce64e0>
```

## After

The old name stays importable at runtime (with a DeprecationWarning) and resolves for type checkers via the TYPE_CHECKING alias.

Client Class:
```python
class AsyncBedrockRuntimeClient
    ...

if TYPE_CHECKING:
    # Deprecated alias for backwards compatibility, to be removed.
    BedrockRuntimeClient = AsyncBedrockRuntimeClient

def __getattr__(name: str) -> Any:
    if name == "BedrockRuntimeClient":
        warnings.warn(
            "BedrockRuntimeClient is deprecated, use AsyncBedrockRuntimeClient instead. "
            "This alias will be removed in a future version.",
            DeprecationWarning,
            stacklevel=2,
        )
        return AsyncBedrockRuntimeClient
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

Usage:
```
>>> from aws_sdk_bedrock_runtime.client import AsyncBedrockRuntimeClient
>>> AsyncBedrockRuntimeClient()
<aws_sdk_bedrock_runtime.client.AsyncBedrockRuntimeClient object at 0x102ce7530>
>>> from aws_sdk_bedrock_runtime.client import BedrockRuntimeClient
<stdin>:1: DeprecationWarning: BedrockRuntimeClient is deprecated, use AsyncBedrockRuntimeClient instead. This alias will be removed in a future version.
>>> BedrockRuntimeClient()
<aws_sdk_bedrock_runtime.client.AsyncBedrockRuntimeClient object at 0x102ce64e0>
```

A newly added service (not in the allowlist) generates just the class, no alias:

```
class AsyncWeatherClient:
    ...
```

## Backwards compatibility

Existing customer code using `BedrockRuntimeClient` will continue to work since we're aliasing it to `AsyncBedrockRuntimeClient`. This will be removed in a future version of the SDK in a minor bump before GA.

## Testing

Regenerated all 7 aws-sdk-python clients. I verified all clients have the renamed class and only the existing clients get the aliasing logic.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,27 @@
 
 ## Unreleased
 
+## v0.4.0
+
 ### Breaking Changes
 
-* Removed the `http_client` config option in favor of the generic `transport`.
+* Renamed generated AWS service clients to include `Async` (e.g.
+  `BedrockRuntimeClient` is now `AsyncBedrockRuntimeClient`) to match the
+  Python community convention for async clients. The old name is still
+  available as a deprecated alias that emits a `DeprecationWarning`, and will
+  be removed in an upcoming release. The unprefixed name will later be
+  reintroduced for synchronous clients.
 
-### Features
+<!--
+TODO: Backfill pre-0.4.0 codegen releases into this file. Prior changes were
+accumulated under "Unreleased" and never cut over per release, so the following
+already shipped (in-tree as of the 0.3.0 codegen release) and need to be placed
+under their correct version section:
 
-* Removed code-generated protocol implementations in favor of hand-written
-  implementations based on schemas.
-* Moved documentation for structure members into doc strings after the member's
-  dataclass field declaration.
+* (Breaking) Removed the `http_client` config option in favor of the generic
+  `transport`.
+* (Feature) Removed code-generated protocol implementations in favor of
+  hand-written implementations based on schemas.
+* (Feature) Moved documentation for structure members into doc strings after
+  the member's dataclass field declaration.
+-->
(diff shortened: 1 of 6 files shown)


SamRemis (dismissed):
Looks good; got this generated and running locally.

We could make an allow-list instead so we aren't shipping new clients with deprecated aliases, but as long as we swap this out soon, it's not a big concern.

jonathan343:
Yea, I though of the allow list but got lazy with the decision. Since you also had the same idea, I think it's worth doing. I'll add in the next revision.

SamRemis (approved):

PR #738, opened 2026-07-13
Add support for config resolver

ubaskota:
*Issue #, if available:*

*Description of changes:*
This change implements the async config resolution pipeline with `AsyncAwsConfig` using a `resolve()` classmethod as the only construction path. It adds `SharedConfigContext` for holding env vars and memoized config file data, `FieldSpec` for per-field resolution metadata, and provenance tracking via `ConfigSource`. Resolvers for region and `retry_strategy_options` are included along with validators for both fields.

*Testing:*
- Added unit and end-to-end tests covering env var resolution, profile resolution, explicit overrides, defaults, provenance tracking, validator enforcement, and construction blocking, and verified that they all pass.
- All existing tests pass 

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
@@ -1,0 +1,176 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from typing import Any, ClassVar, Self
+
+from smithy_aws_core.config.context import FileSystem, SharedConfigContext
+from smithy_aws_core.config.exceptions import ConfigError
+from smithy_aws_core.config.resolvers import resolve_region, resolve_retry_config
+from smithy_aws_core.config.types import UNSET, ConfigSource, FieldSpec, Resolved
+from smithy_aws_core.config.validators import (
+    validate_region,
+    validate_retry_strategy_options,
+)
+from smithy_core.retries import RetryStrategyOptions
+
+
+@dataclass(kw_only=True)
+class AsyncAwsConfig:
+    """Base configuration class for all AWS services.
+
+    Fields are resolved asynchronously from multiple sources (env vars,
+    config files, defaults) through the resolve() classmethod.
+
+    Do not instantiate directly — use:
+        config = await AsyncAwsConfig.resolve()
+    """
+
+    region: str | None = None
+    retry_strategy_options: RetryStrategyOptions | None = None
+
+    _ctx: SharedConfigContext | None = field(default=None, repr=False, compare=False)
+    _sources: dict[str, ConfigSource] = field(  # type: ignore[assignment]
+        default_factory=dict,
+        repr=False,
+        compare=False,
+    )
+
+    _FIELDS: ClassVar[dict[str, FieldSpec]] = {
+        "region": FieldSpec(
+            default=None,
+            resolver=resolve_region,
+            validator=validate_region,
+        ),
+        "retry_strategy_options": FieldSpec(
+            default_factory=RetryStrategyOptions,
+            resolver=resolve_retry_config,
+            validator=validate_retry_strategy_options,
+        ),
+    }
+
+    def __post_init__(self) -> None:
+        """Block direct construction. Use resolve() instead."""
+        raise ConfigError(
+            f"{type(self).__name__} cannot be constructed directly. "
+            f"Use `await {type(self).__name__}.resolve(...)` instead."
+        )
+
+    @classmethod
+    async def resolve(
+        cls,
+        *,
+        profile: str | None = None,
+        env: Mapping[str, str] | None = None,
+        fs: FileSystem | None = None,
+        config_file_path: str | None = None,
+        credentials_file_path: str | None = None,
+        **overrides: Any,
+    ) -> Self:
+        """Resolve a config object from environment, config files, and defaults.
+
+        This is the only supported way to create a config instance.
+
+        :param profile: Override the active profile name.
+        :param env: Override the environment variable mapping.
+        :param fs: Override the filesystem abstraction.
+        :param config_file_path: Override path for config file.
+        :param credentials_file_path: Override path for credentials file.
+        :param overrides: Explicit field values that skip resolution.
+        :returns: A fully-resolved config instance.
+        """
+        ctx = SharedConfigContext(
+            profile_name=profile,
+            env=env,
+            fs=fs,
+            config_file_path=config_file_path,
+            credentials_file_path=credentials_file_path,
+        )
+
+        # Create the instance bypassing __post_init__ check
+        instance = cls._create_instance()
+        instance._ctx = ctx
+
+        # Resolve each field
+        await instance._resolve_fields(overrides)
+
+        return instance
+
+    def source_of(self, field_name: str) -> ConfigSource | None:
+        """Get the source that provided a field's value.
+
+        :param field_name: The config field name.
+        :returns: The ConfigSource, or None if not tracked.
+        """
+        return self._sources.get(field_name)
+
+    def resolution_context(self) -> SharedConfigContext | None:
+        """Get the resolution context used to create this config.
+
+        :returns: The SharedConfigContext, or None if not available.
+        """
+        return self._ctx
+
+    @classmethod
+    def _create_instance(cls) -> Self:
+        """Create an instance that bypasses construction blocking."""
+        instance = object.__new__(cls)
+        object.__setattr__(instance, "_sources", {})
+        object.__setattr__(instance, "_ctx", None)
+        for field_name in cls._FIELDS:
+            object.__setattr__(instance, field_name, UNSET)
+        return instance
+
+    async def _resolve_fields(self, overrides: dict[str, Any]) -> None:
+        """Run the resolution pipeline for all fields."""
+        unknown = set(overrides) - set(self._FIELDS)
+        if unknown:
+            raise ConfigError(
+                f"Unknown config field(s): {sorted(unknown)}. "
+                f"Valid fields are: {sorted(self._FIELDS)}"
+            )
+
+        for field_name, spec in self._FIELDS.items():
+            # check for overrides first
+            if field_name in overrides:
+                value = overrides[field_name]
+                setattr(self, field_name, value)
+                self._sources[field_name] = ConfigSource.OVERRIDE
+            # check in resolver
+            elif spec.resolver is not None:
+                result: Resolved[Any] = await spec.resolver(self._ctx)
+                if result.value is not UNSET:
+                    setattr(self, field_name, result.value)
+                    self._sources[field_name] = result.source
+                else:
+                    # If resolver returned UNSET, fall back to default
+                    self._apply_default(field_name, spec)
+
+            else:
+                # No resolver, use default directly
+                self._apply_default(field_name, spec)
+
+            # Run validator for all sources
+            if spec.validator is not None:
+                spec.validator(getattr(self, field_name))
+
+    def _apply_default(self, field_name: str, spec: FieldSpec) -> None:
+        """Apply the default value for a field."""
+        if spec.default_factory is not None:
+            value = spec.default_factory()
+        else:
+            value = spec.default
+        setattr(self, field_name, value)
+        self._sources[field_name] = ConfigSource.DEFAULT
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Track provenance when fields are set with plugins after construction"""
+        super().__setattr__(name, value)
+        # Mark as override only if the field is in _FIELDS and was already resolved
+        if (
+            name in self.__class__._FIELDS
+            and hasattr(self, "_sources")
+            and name in self._sources
+        ):
+            self._sources[name] = ConfigSource.OVERRIDE
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/context.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/context.py
@@ -1,0 +1,139 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+import logging
+import os
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Any, Protocol, runtime_checkable
+
+from smithy_aws_core.config import load_config
+from smithy_aws_core.config.merged_config import MergedConfig
+
+logger = logging.getLogger(__name__)
+
+
+@runtime_checkable
+class FileSystem(Protocol):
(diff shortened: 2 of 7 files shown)


packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+
+    :param ctx: The shared resolution context.
+    :param env_vars: Environment variable names to check, in order.
+    :param profile_keys: Config file profile keys to check, in order.
+    :returns: Resolved int value with source, or Resolved(value=UNSET) if not found.
+    """
+    result = await _resolve_str(ctx, env_vars=env_vars, profile_keys=profile_keys)
+    if result.value is UNSET:
+        return Resolved(value=UNSET, source=ConfigSource.DEFAULT)  # type: ignore[arg-type]
+    if result.value is None:
```
Alan4506, line 61: Nit: Is this dead code? `_resolve_str` may never return `Resolved` with a `None` value?

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+    if mode_result.value is not UNSET and mode_result.value is not None:
+        kwargs["retry_mode"] = mode_result.value
+        sources_found.append(mode_result.source)
+
+    if attempts_result.value is not UNSET and attempts_result.value is not None:
+        kwargs["max_attempts"] = attempts_result.value
+        sources_found.append(attempts_result.source)
+
+    source = (
+        _strongest_source(*sources_found) if sources_found else ConfigSource.DEFAULT
```
Alan4506, line 129: Since we only have `source_of("retry_strategy_options")`, do we lose the ability to tell where `retry_mode` vs `max_attempts` each came from? E.g. `retry_mode` from env but `max_attempts` from profile would both just report `ENV`? I forgot if we discussed this yesterday - please refresh my memory if we did. I'm curious about this because it may affect another [PR](https://github.com/smithy-lang/smithy-python/pull/733#discussion_r3538673741) - which depends on provenance tracking, and needs to know if the `max_attempts` itself has been set by customer or not.  If we've decided to keep provenance at the options level (rather than per-field) for now, then the other PR may consider a different way to determine if the customer has set the `max_attempts` manually.

ubaskota: Yes, for composite resolvers like `retry_strategy_options`, we report the strongest source across the fields that compose it (example: if `retry_mode` comes from `ENV` and `max_attempts` from `PROFILE`, the source is `ENV`). This is something the team decided upon. If your other PR needs to know specifically whether `max_attempts` was user-set, we can discuss that.

packages/smithy-aws-core/tests/unit/config/test_resolver.py
```diff
@@ -0,0 +1,285 @@
...
+        ctx = SharedConfigContext(
+            env={"AWS_RETRY_MODE": "standard", "AWS_MAX_ATTEMPTS": "10"},
+        )
+        result = await resolve_retry_config(ctx)
+        assert result.value.retry_mode == "standard"
+        assert result.value.max_attempts == 10
+        assert result.source == ConfigSource.ENV
+
+    @pytest.mark.asyncio
+    async def test_uses_defaults_when_nothing_found(self):
```
jonathan343, line 98: There are a few tests in this file that build `SharedConfigContext(env={})` with no file_path overrides. This causes them to eventually read the real `~/.aws/config` and `~/.aws/credentials` files on the runners machine. If the machine's `[default]` profile sets `retry_mode` or `max_attempts`, the assertions will fail. 

This issue affects the following tests:
- `test_uses_defaults_when_nothing_found`
- `test_partial_resolution_uses_defaults_for_missing`
- `test_max_attempts_unset_when_not_found`.

packages/smithy-aws-core/tests/unit/config/test_resolver.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+class TestAsyncAwsConfigResolve:
+    @pytest.mark.asyncio
+    async def test_resolves_region_from_env(self):
+        config = await AsyncAwsConfig.resolve(env={"AWS_REGION": "us-west-2"})
+        assert config.region == "us-west-2"
+
+    @pytest.mark.asyncio
+    async def test_resolves_retry_from_env(self):
```
jonathan343, line 142: This test is also environment dependent. It relies on the test runner's machine to define `region = ...` in the `~/.aws/config` file. If it doesn't it will fail with the following error:

```
smithy_aws_core.config.exceptions.ConfigError: Invalid value for 'region': None. Region is required and must be set.
```

Tests are run on various different dev, CI, and containerized setups, so they need to be deterministic and not reliant on the environment they're run in.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+        env_vars=("AWS_MAX_ATTEMPTS",),
+        profile_keys=("max_attempts",),
+    )
+
+    # Build kwargs for values that were resolved
+    kwargs: dict[str, Any] = {}
+    sources_found: list[ConfigSource] = []
+
+    if mode_result.value is not UNSET and mode_result.value is not None:
+        kwargs["retry_mode"] = mode_result.value
```
jonathan343, line 121: We need to validate that the configured `retry_mode` value is one we accept/support. I'd recommend validating against `typing.get_args(RetryStrategyType)` or something similar so it stays aligned with the actual supported list. If something is configured that isn't in that list, we should raise an error.

Here is an example of me being able to configure a mode that doesn't exist:

config file:

```ini
# ~/.aws/config
[default]
retry_mode = fake-mode
```

```python
(smithy-python) $ python -m asyncio
asyncio REPL 3.12.13 (main, Mar 25 2026, 03:16:06) [Clang 22.1.1 ] on darwin
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from smithy_aws_core.config.aws_config import AsyncAwsConfig
>>> config = await AsyncAwsConfig.resolve()
>>> config
AsyncAwsConfig(region='us-west-2', retry_strategy_options=RetryStrategyOptions(retry_mode='fake-mode', max_attempts=None))
```


For reference, this is what botocore raises:

```
botocore.exceptions.InvalidRetryModeError: Invalid value provided to "mode": "fake-mode" must be one of: ('legacy', 'standard', 'adaptive')
```

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+        sources_found.append(mode_result.source)
+
+    if attempts_result.value is not UNSET and attempts_result.value is not None:
+        kwargs["max_attempts"] = attempts_result.value
```
jonathan343, line 125: I also think we should validate the these ranges. Here is an example of me sucessfully configuring `max_attempts = -1` which shouldn't be allowed:

config file:

```ini
# ~/.aws/config
[default]
max_attempts = -1
```

```python
(smithy-python) $ python -m asyncio
asyncio REPL 3.12.13 (main, Mar 25 2026, 03:16:06) [Clang 22.1.1 ] on darwin
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from smithy_aws_core.config.aws_config import AsyncAwsConfig
>>> config = await AsyncAwsConfig.resolve()
>>> config
AsyncAwsConfig(region='us-west-2', retry_strategy_options=RetryStrategyOptions(retry_mode='standard', max_attempts=-1))
```

Botocore error for reference:

```
botocore.exceptions.InvalidMaxRetryAttemptsError: Value provided to "max_attempts": -1 must be an integer greater than or equal to 1.
```

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,42 @@
...
+
+import re
+
+from smithy_aws_core.config.exceptions import ConfigError
+from smithy_core.retries import RetryStrategyOptions
+
+_REGION_PATTERN = re.compile(r"^(?![0-9]+$)(?!-)[a-zA-Z0-9-]{1,63}(?<!-)$")
+
+
+def validate_region(value: object) -> None:
```
jonathan343, line 12: It seems like we're only validating on the initial `AsyncAwsConfig.resolve(...)` call. When users set values directly after resolution, I feel like we should be revalidating. 

Example 1: config rejects a bad value at resolution time:

```python
(smithy-python) $ python -m asyncio
asyncio REPL 3.12.13 (main, Mar 25 2026, 03:16:06) [Clang 22.1.1 ] on darwin
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from smithy_aws_core.config.aws_config import AsyncAwsConfig
>>> config = await AsyncAwsConfig.resolve(region="bad-value!")
  File "/Users/gytndd/.local/share/uv/python/cpython-3.12.13-macos-aarch64-none/lib/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/gytndd/.local/share/uv/python/cpython-3.12.13-macos-aarch64-none/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py", line 96, in resolve
    await instance._resolve_fields(overrides)
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py", line 156, in _resolve_fields
    spec.validator(getattr(self, field_name))
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/validators.py", line 26, in validate_region
    raise ConfigError(
smithy_aws_core.config.exceptions.ConfigError: Invalid value for 'region': 'bad-value!'. Must be a valid AWS region identifier.
```

Example 2: we config a valid region, then successfully overwrite with a bad region:

```python
>>> config = await AsyncAwsConfig.resolve(region="us-east-1")
>>> config.region
'us-east-1'
>>> config.region = "bad-value!"
>>> config.region
'bad-value!'
```

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+        return Resolved(value=UNSET, source=ConfigSource.DEFAULT)  # type: ignore[arg-type]
+    if result.value is None:
+        return Resolved(value=None, source=result.source)
+    try:
+        return Resolved(value=int(result.value), source=result.source)
+    except (ValueError, TypeError):
+        raise ConfigError(
+            f"Invalid integer value {result.value!r} for config key. "
+            "Expected a valid integer."
+        )
```
jonathan343, line 69: nit - We should catch the error and raise from it:

```python
except (ValueError, TypeError) as e:
    raise ConfigError(...) from e
```

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -0,0 +1,176 @@
...
+            f"{type(self).__name__} cannot be constructed directly. "
+            f"Use `await {type(self).__name__}.resolve(...)` instead."
+        )
+
+    @classmethod
+    async def resolve(
+        cls,
+        *,
+        profile: str | None = None,
+        env: Mapping[str, str] | None = None,
```
jonathan343, line 65: It doesn't seem like we're gaining the full value of `env` since it doesn't get propagated to the methods lower in the call stack. Specifically, it gets lost through `parsed_profiles() → load_config() → _resolve_config_paths()` which just use `os.environ`. 

Could we thread `ctx.env` all the way through `load_config`/`_resolve_config_paths` (an optional `env` param defaulting to `os.environ`) so all env lookups share one source?

packages/smithy-aws-core/src/smithy_aws_core/config/context.py
```diff
@@ -0,0 +1,139 @@
...
+        :param env: Environment variable mapping. Defaults to os.environ.
+        :param fs: Filesystem abstraction for reading config files.
+            Defaults to real disk I/O.
+        :param http_client: HTTP client for network-based resolvers
+            (e.g., IMDS).
+        :param config_file_path: Override path for config file.
+        :param credentials_file_path: Override path for credentials file.
+        """
+        self._env: Mapping[str, str] = env if env is not None else os.environ
+        self._fs: FileSystem = fs if fs is not None else DefaultFileSystem()
```
jonathan343, line 88: I don't really get why this file abstraction was introduced. It seems to be stored here but is never used. Can you explain to me why this was introduced and what value it's supposed to bring?

jonathan343 (changes requested):
Thanks Ujjwal. Overall, everything seems functional after testing with different direct, env var, and config file setups.

I let some comments calling out some gaps and bugs in the implementation. Let me know if you have any questions. I'll continue to think about if this is the optimal interface we want to expose.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
arandito: I have some concerns about this after investigating an env mapping override for our new credential chain.

Does the `env` mapping provide any value outside of environment injection for tests? Correct me if I'm wrong, but I don't think we want or need to expose this for customers to provide custom mappings in the config. Should we not always use the os environment? 

If the above is true, keeping this adds unnecessary plumbing which can be replaced by `monkeypatch` in tests.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
ubaskota: Removed

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
ubaskota: Ideally yes,` env` provides value beyond tests. When multiple clients share a config, a captured env snapshot ensures all clients see the same environment state regardless of whether the values are changed mid-process. However, we're not yet sure about the chances of environment variables changing while an application is running and whether that's something our users wold need. Plus, if it does happen, what should the correct behavior be (fail vs use the original value). After discussing internally, we decided to drop it for now and add it back in the future if needed.

packages/smithy-aws-core/src/smithy_aws_core/config/context.py
ubaskota: The `FileSystem` abstraction exists for test isolation. `NullFileSystem` simulates absent config files (so the resolver falls through to defaults), and `FakeFileSystem` provides in-memory file content for testing profile-based resolution without hitting disk.

PR #742, opened 2026-07-16
Resolve REST JSON modeled errors by shape name

jonathan343:
## Overview

This PR fixes REST JSON modeled error resolution when a wire error identifier contains a namespace that differs from the modeled error's namespace.

After an exact lookup fails, REST JSON matches the identifier's shape name against the operation's modeled errors and uses the matched error's full shape ID.

## Background

Smithy 1.72.0 added a protocol test for error identifiers containing an unrecognized namespace in [smithy-lang/smithy#3157](https://github.com/smithy-lang/smithy/pull/3157).

The new test surfaced a failure in the Smithy dependency update from [#739](https://github.com/smithy-lang/smithy-python/pull/739). We preserved the namespace received over the wire, causing the exact modeled error lookup to fail and return a generic `CallError` instead of `FooError`.

```
Failed: Expected 'FooError' exception to be thrown, but received CallError: Unknown error for operation aws.protocoltests.restjson#GreetingWithErrors - status: 500 - id: aws.different.namespace#FooError
```

The REST JSON specification requires clients to ignore namespaces and URI suffixes when resolving error identifiers. This change applies that normalization before looking up the modeled error.

## Testing

Updated Smithy from `1.71.0` to `1.72.1`, enabling the protocol test introduced by [smithy-lang/smithy#3157](https://github.com/smithy-lang/smithy/pull/3157). All tests passed when running `make test-protocols`.

---

### DIFF
--- a/packages/smithy-aws-core/tests/unit/aio/test_protocols.py
+++ b/packages/smithy-aws-core/tests/unit/aio/test_protocols.py
@@ -10,7 +10,6 @@ from smithy_aws_core.aio.protocols import (
     AWSErrorIdentifier,
     AWSJSONDocument,
     AwsQueryClientProtocol,
-    RestJsonClientProtocol,
 )
 from smithy_aws_core.traits import AwsQueryTrait
 from smithy_core.deserializers import ShapeDeserializer
@@ -21,7 +20,7 @@ from smithy_core.prelude import STRING
 from smithy_core.schemas import APIOperation, Schema
 from smithy_core.serializers import ShapeSerializer
 from smithy_core.shapes import ShapeID, ShapeType
-from smithy_core.traits import HTTPTrait, Trait
+from smithy_core.traits import Trait
 from smithy_core.types import TypedProperties
 from smithy_http import Fields, tuples_to_fields
 from smithy_http.aio import HTTPRequest, HTTPResponse
@@ -39,10 +38,6 @@ from smithy_json import JSONSettings
         (
             "com.test#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate",
             "com.test#FooError",
-        ),
-        (
-            "com.other#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate",
-            "com.other#FooError",
         ),
         ("", None),
         (":", None),
@@ -80,12 +75,6 @@ def test_aws_error_identifier(header: str | None, expected: ShapeID | None) -> N
                 "__type": "com.test#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate"
             },
             "com.test#FooError",
-        ),
-        (
-            {
-                "__type": "com.other#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate"
-            },
-            "com.other#FooError",
         ),
         ({"code": "FooError"}, "com.test#FooError"),
         ({"code": "com.test#FooError"}, "com.test#FooError"),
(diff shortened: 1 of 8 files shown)


packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -39,6 +40,10 @@
             "com.test#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate",
             "com.test#FooError",
         ),
+        (
```
arandito: I'm not sure what value this test adds compared to case above. The identifier's behavior doesn't branch based off the namespace. `parse_error_code` just parses the wire code into a `ShapeId` agnostic of what namespace it has.

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -76,6 +81,12 @@ def test_aws_error_identifier(header: str | None, expected: ShapeID | None) -> N
             },
             "com.test#FooError",
         ),
+        (
```
arandito: Same here. The discriminator's behavior also doesn't branch based off the namespace being different.

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -120,6 +131,14 @@ def test_aws_json_document_discriminator(
     shape_type=ShapeType.SERVICE,
     traits=[AwsQueryTrait(None)],
 )
+_REST_JSON_SERVICE_SCHEMA = Schema.collection(
+    id=ShapeID("com.test#RestJsonService"),
+    shape_type=ShapeType.SERVICE,
+)
+_CROSS_NAMESPACE_ERROR_SCHEMA = Schema.collection(
```
arandito: These tests are bit confusing to follow because they do not directly reproduce the behavior we are fixing, which is when a server returns an error code with an unknown namespace.

Currently, the error is modeled in a shared namespace, and the resolver is exercised only when the wire code contains an unqualified shape name. Could we instead model the error in the service namespace, return the same shape name with an unknown namespace, and verify that it resolves to the modeled error?

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
```diff
@@ -178,6 +215,57 @@ def _mock_operation(
...
+            id="body-unqualified",
+        ),
+        pytest.param(
+            [("content-type", "application/json")],
+            b'{"__type":"com.shared#CrossNamespaceError"}',
+            id="body-qualified",
+        ),
+    ],
+)
+async def test_rest_json_resolves_cross_namespace_modeled_error(
```
arandito: General question because I keep seeing us try to add tests to this file for protocol behavior fixes. 

Should we delegate testing protocol behavior to the Smithy protocol tests to keep a single source of truth instead of our own unit tests?

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
jonathan343: Yea, intention of this test was for testing errors with a namespace that differs from the one used by the operation (e.g. `com.other` instead of `com.test`). But that doesn't really seem to apply here.  Removed in 36e1b91f8e4d958b50163ea8e27d94138ad247a3

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
jonathan343: Removed in 36e1b91f8e4d958b50163ea8e27d94138ad247a3

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
jonathan343: Removed them in favor of protocol tests in 36e1b91f8e4d958b50163ea8e27d94138ad247a3

packages/smithy-aws-core/tests/unit/aio/test_protocols.py
jonathan343: Yea, I agree these are verbose and we can rely on protocol tests for now. To ensure this addresses the new protocol test, I've bumped the smithy version in 36e1b91f8e4d958b50163ea8e27d94138ad247a3.

arandito (approved):
:shipit:

PR #743, opened 2026-07-17
fix: synchronize aiohttp constraint in uv.lock

jonathan343:
## Overview

This PR updates `uv.lock` to match the `aiohttp` constraint declared by `smithy-http`.

## Background

Dependabot PR #710 upgraded the resolved `aiohttp` version to 3.14.0, but also changed the lockfile metadata constraint from `>=3.11.12,<4.0` to `>=3.14.0,<4.0`. The source constraint in `packages/smithy-http/pyproject.toml` was not changed.

As a result, running `make install` causes `uv sync` to detect the stale metadata and modify `uv.lock`.

This has already been reported to dependabot in https://github.com/dependabot/dependabot-core/issues/14937

## Testing

- Ran `uv lock --check` successfully.

---

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -753,7 +753,7 @@ awscrt = [
 
 [package.metadata]
 requires-dist = [
-    { name = "aiohttp", marker = "extra == 'aiohttp'", specifier = ">=3.14.0,<4.0" },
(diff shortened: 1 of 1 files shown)


Alan4506 (approved):
LGTM. Thanks @jonathan343!

PR #744, opened 2026-07-19
Add experimental Python codegen CLI scaffold

jonathan343:
## Overview

This PR lays the foundation for a Python-native Smithy code generator. It adds an experimental `smithy-python` package with the command-line interface that will eventually generate Python clients and standalone types packages.

The CLI currently validates its inputs and then exits with an explicit “generation is not implemented yet” error. The existing Java generator remains the primary one while the Python implementation is developed incrementally.

## Background

The Python generator will consume [Smithy JSON AST](https://smithy.io/2.0/spec/json-ast.html) models and produce packages compatible with the existing Smithy Python runtime. It is distributed separately from those runtime packages and is only needed during code generation.

Smithy’s `run` plugin invokes the generator as an external process, allowing it to integrate with standard Smithy builds without being loaded into the Smithy CLI or implemented in Java. Read [Smithy run plugin docs](https://smithy.io/2.0/guides/smithy-build-json.html#run-plugin) for more info.

## Changes Summary

- Created a new `smithy-python` package.
- Add `generate client` and `generate types` commands.
- Support direct invocation using `--model` and `--output`.
  - Note: This list of params will grow as in subsequent PRs
- Support Smithy `run` plugin invocation through standard input and the provided plugin environment.
- Validate invocation inputs and distinguish argument errors from I/O and generation failures.
- Add unit tests covering CLI behavior, environment parsing, and the error hierarchy.
- Document the generator architecture, CLI contract, and migration strategy.

### CLI smoke tests

```console
$ smithy-python --version
smithy-python 0.0.0
```

```console
$ smithy-python generate --help
usage: smithy-python generate [-h] {client,types} ...

positional arguments:
  {client,types}
    client        Generate a client package
    types         Generate a standalone types package

options:
  -h, --help      show this help message and exit
```

Both generation commands accept a model and reach the expected placeholder behavior:

```console
$ echo '{}' | smithy-python generate client --output /tmp/smithy-python-client
smithy-python: error: client generation is not implemented yet
```

```console
$ echo '{}' | smithy-python generate types --output /tmp/smithy-python-types
smithy-python: error: types generation is not implemented yet
```

Both generation commands exit with status 1, as expected for the current implementation.

---


packages/smithy-python/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# Changelog
```
jonathan343, line 1: Note for reviewers: Our release infra will automatically handle version bumping and changelog management.

designs/codegen/cli.md
JordonPhillips, line 1: I haven't talked much about this publicly since I'm still working on implementations and a blog post, but check a look at [shape closures](https://smithy.io/2.0/spec/shape-closures.html). This is a way to drive generation that doesn't root everything in a service.

What it means for you is relatively little, other than that you should tolerate generating without a set service. Current type codegen generates a synthetic one, but you shouldn't operate that way.

Computing a closure is beyond your scope as a non-java plugin. I would suggest rather that you generate what's given to you and put the onus on customers to use smithy-build transforms to narrow it down to what they want. Maybe later you could implement everything needed to compute it and relax that restriction.

designs/codegen/cli.md
```diff
@@ -0,0 +1,95 @@
+# Code Generator CLI
+
+The `smithy-python` command is the process interface described in the
+[Python Code Generation](index.md) overview. It supports direct use and
+invocation from Smithy's
+[`run` plugin](https://smithy.io/2.0/guides/smithy-build-json.html#run-plugin).
+
+## Commands
+
+Generation is organized by artifact type:
```
JordonPhillips, line 10: But what if you could mix artifacts. You don't now - but you could. Types+client or types+server is an obvious combo that's easy to do without much difficulty.

```console
smithy-python generate --modes client,types
```

You could default that to just be client since most will want that.

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,140 @@
...
+    artifacts = generate.add_subparsers(dest="artifact", required=True)
+    for name, help_text in (
+        ("client", "Generate a client package"),
+        ("types", "Generate a standalone types package"),
+    ):
+        artifact = artifacts.add_parser(name, help=help_text)
+        artifact.add_argument(
+            "--model",
+            type=Path,
+            help="Read the JSON AST from a file instead of standard input",
```
JordonPhillips: Something like "The Smithy JSON AST model file to use for code generation. If not set, the model is read from standard input."

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,140 @@
...
+
+    commands = parser.add_subparsers(required=True)
+    generate = commands.add_parser("generate", help="Generate Python source")
+    artifacts = generate.add_subparsers(dest="artifact", required=True)
+    for name, help_text in (
+        ("client", "Generate a client package"),
+        ("types", "Generate a standalone types package"),
+    ):
+        artifact = artifacts.add_parser(name, help=help_text)
+        artifact.add_argument(
```
JordonPhillips: This seems like a use-case for [parent parsers](https://docs.python.org/3/library/argparse.html#parents).

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,140 @@
...
+            "--model",
+            type=Path,
+            help="Read the JSON AST from a file instead of standard input",
+        )
+        artifact.add_argument(
+            "--output",
+            type=Path,
+            help="Output directory for direct invocation",
```
JordonPhillips: Best to mention here where the default value comes from.

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,140 @@
...
+
+_GENERATION_NOT_IMPLEMENTED: Final = (
+    "smithy-python: error: {artifact} generation is not implemented yet\n"
+)
+
+
+@dataclass(frozen=True, slots=True)
+class _Invocation:
+    artifact: str
+    model_source: bytes
```
JordonPhillips, line 27: A readable would be better here I think. You don't want to have to buffer all of that only to pass it into json.reads or whatever else.

JordonPhillips (dismissed):

designs/codegen/cli.md
```diff
@@ -0,0 +1,95 @@
...
+                "run::python-client": {
+                    "command": ["smithy-python", "generate", "client"]
+                }
+            }
+        }
+    }
+}
+```
+
+Artifact-specific options will be appended to `command` after they are defined.
```
Alan4506, line 56: This is more for my own understanding: the existing Java generator receives `service`, `module`, and `moduleVersion` as plugin settings for each generated client. So in the future are we going to add options like the following?
```json
{
  "run::python-client": {
    "command": [
      "smithy-python",
      "generate",
      "client",
      "--service",
      "com.amazonaws.bedrockruntime#AmazonBedrockFrontendService",
      "--module",
      "aws_sdk_bedrock_runtime",
      "--module-version",
      "0.8.0"
    ]
  }
}
```

jonathan343: Yup. I need to look into [Jordon's comment](https://github.com/smithy-lang/smithy-python/pull/744#discussion_r3614606335) to see if `--service` will still be needed. I'll also consider using `--package-name` which we derive the module name from w/ an optional override (.e.g a `aws-sdk-s3` package name generates a  `aws_sdk_s3` module name). I also think `--module-version` should become `--package-version`.

We may also grow this overtime to add new options like `--http-client aiohttp | httpx | awscrt`

JordonPhillips: Note that you can also put these into environment variables. The [run plugin](https://smithy.io/2.0/guides/smithy-build-json.html#run-plugin) supports putting stuff there. So you could also accept them from the environment if you want.

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Agreed. Updated the wording to:

> Output directory for generated files. Defaults to the Smithy run plugin's output directory (SMITHY_PLUGIN_DIR) when invoked by Smithy.

designs/codegen/cli.md
jonathan343: I'm a bit hesitant to combine multiple modes into one command since that complicates mode-specific options. I'm wondering if we instead could do something like `smithy-python generate client ... --with-types` or similar. 

But honestly, I'm not sure what you mean by client+types compared to what we have now. Can you elaborate a bit?

packages/smithy-python/src/smithy_python/cli.py
jonathan343: I don't think a readable would provide much more value here. `json.load` just does `fp.read()` internally, and we need the whole AST in memory anyway to resolve shape references. Using an open readable also means owning when to close it. We can probably do that cleanly with a context manager, but seems unnecessary. Additionally, the file and stdin cases close differently so we'd also have to keep track of that.

Lmk if I'm missing something and happy to re-evaluate. Since this is all under the hood, we can also always add if later if we find a good reason to.

designs/codegen/cli.md
jonathan343: Nice, thanks for making us aware of this.

I agree with you. I think not requiring `--service <service_id>` is also an improved user experience. It's pretty easy to detect on our end. For the AWS services, we've automated our infra to do `smithy select --selector service` to find it for us. I think we can do that at the Python layer here. The only issue I see here is if someone defines two services (not sure if smithy allows it). But in that case, we'd ask them to specify.

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Addressed in 0dcae939127a368c2163e3b0259b48167712d45e

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Addressed in 0dcae939127a368c2163e3b0259b48167712d45e

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Addressed in 0dcae939127a368c2163e3b0259b48167712d45e

jonathan343 (commented):
Thanks @JordonPhillips!

I addressed a few of your comments. Others might need clarification or further discussion.

jonathan343:
I'm so used to reviewing other people's PRs I accidentally submitted my responses as a review. Sorry if it looks weird. I see duplicate responses now.

designs/codegen/cli.md
JordonPhillips: Yeah, that's basically how it works on the java side right now. You can ask that there be exactly one service. It may be possible to generate for multiple services, but there's no expectation.

designs/codegen/cli.md
JordonPhillips: The idea is to support related types that aren't part of the service's operations. For example, say a service wants to define the events they send out through SNS. Those wouldn't be part of the service shape closure, so if you're filtering down to that then you exclude those shapes.

packages/smithy-python/src/smithy_python/cli.py
JordonPhillips: > `json.load` just does `fp.read()` internally

Why is python like this.

As long as you discard the string as soon as possible in favor of a loaded dict or whatever then it's fine.

designs/codegen/cli.md
jonathan343: Got it, that makes sense. The gap is that the Java client generator only emits the service closure (`directive.connectedShapes()`), so shapes like your SNS events example just get dropped silently today.

My first thought was to have the Python generator skip the closure entirely and generate every data shape in the AST. Before committing to that I ran it over all 431 published AWS models. It broke healthlake out of the box (leaked, unconnected shapes in a second namespace colliding with real ones) and pulled in orphan `Cfn*` types for two others, while the service closure reproduced the existing SDK surface on all 431. So I'm keeping the service closure as the default, which also matches what Java, smithy-ruby, and smithy-php do. The CLI prints a note when unconnected shapes get left out so it's not silent anymore.

Everything else you called out is in: `--service` is optional (we auto-detect when there's one), we don't synthesize a service, and `types` works without a service by just generating every data shape.

That does mean the "types outside the closure" case is deferred for now. I left room in the design doc for an all-shapes option or reading `shapeClosures` metadata later. I'd like to get moving on the core implementation and we can re-evaluate the interface before we cut a major version. Lmk if you see this as a blocker though and we can hash it out now.

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,153 @@
...
+        output_dir = plugin_dir
+    else:
+        if output_path is None:
+            raise InvalidInvocationError("Direct invocation requires --output")
+        output_dir = output_path
+
+    if model_path is not None:
+        if not model_path.is_file():
+            raise InvalidInvocationError(f"Model path is not a file: {model_path}")
+        model_source = model_path.read_bytes()
```
stobrien89: nit: missing model files exits 2 when design says I/O errors exit 1

stobrien89 (approved):
just one minor comment

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Addressed in 85d815192006d3e4ec674c23ea02584aa8b67dff. Thanks!

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,158 @@
...
+        prog="smithy-python",
+        description="Generate Python source from Smithy models.",
+    )
+    parser.add_argument(
+        "--version",
+        action="version",
+        version=f"%(prog)s {__version__}",
+    )
+
+    commands = parser.add_subparsers(required=True)
```
arandito: nit: Are we planning to add more commands apart from `generate`? If we are, lets add `dest="command"` when adding the subcommands.

packages/smithy-python/src/smithy_python/cli.py
```diff
@@ -0,0 +1,158 @@
...
+
+    try:
+        args = parser.parse_args(argv)
+    except SystemExit as error:
+        return error.code if isinstance(error.code, int) else 1
+
+    try:
+        _resolve_invocation(
+            args,
+            environ=os.environ if environ is None else environ,
```
arandito: tiny nit: The current path resolves environ twice: here and when `PluginEnvironment.from_environ` is called in `_resolve_invocation`. Is there a reason `_resolve_invocation` needs this to be resolved beforehand? If not, can we consolidate it?

arandito (commented):
Had a tiny nit but looks good to me.

packages/smithy-python/src/smithy_python/cli.py
jonathan343: Addressed in c7e2f6738697cf2b7a5c23cc42ebc367c1519222. This was actually done in the stacked PR, pulled it in here: c7e2f6738697cf2b7a5c23cc42ebc367c1519222

arandito (approved):

PR #751, opened 2026-07-23
Add async resolver for AWS config values

ubaskota:
*Issue #, if available:*

*Description of changes:*
This change implements the async config resolution pipeline with `AsyncAwsConfig` using a `resolve()` classmethod as the only construction path. It adds `SharedConfigContext` for holding memoized config file data, `FieldSpec` for per-field resolution metadata, and provenance tracking via `ConfigSource`. Resolvers for `region` and `retry_strategy_options` are included along with validators for both fields.

*Testing:*
- Added unit and end-to-end tests covering env var resolution, profile resolution, explicit overrides, defaults, provenance tracking, FileSystem, validator enforcement, and construction blocking, and verified that they all pass.
- All existing tests pass

Note: This is a revision to the old PR (https://github.com/smithy-lang/smithy-python/pull/738) that was automatically closed after the feature branch was merged.

**Key design points:**

  - `AsyncAwsConfig` cannot be instantiated directly. Calling `AsyncAwsConfig()` raises a `ConfigError`. All construction must go through `await AsyncAwsConfig.resolve()`, which ensures every field is resolved and validated before the object is returned.
  - All config fields are resolved upfront in one call. There is no lazy resolution.
  - Resolution follows a strict precedence: explicit overrides > environment variables > config file profile > default value.
  - Every field tracks where its value came from (provenance), queryable via `config.source_of("field_name")`.

### Usage

  **Direct instantiation is blocked:**

  ```python
from smithy_aws_core.config import AsyncAwsConfig

config = AsyncAwsConfig()
# raises ConfigError: "AsyncAwsConfig cannot be constructed directly.
#   Use `await AsyncAwsConfig.resolve(...)` instead."
  ```

  **Resolve from environment variables:**

  If `AWS_REGION=us-west-2` is set in the environment:

```python
config = await AsyncAwsConfig.resolve()
print(config.region)  # "us-west-2"
```

  **Resolve from config file:**

  If `~/.aws/config` contains:

```
[profile default]
region = us-east-1
retry_mode = standard
max_attempts = 10
```

  And no environment variables override these:

```python
config = await AsyncAwsConfig.resolve()
print(config.region)        # "us-east-1"
print(config.retry_mode)    # "standard"
print(config.max_attempts)  # 10
```

**Override specific values:**
  
Overrides passed to `resolve()` take highest precedence — the value is validated but skips the `env/file` resolution chain entirely:
```python
config = await AsyncAwsConfig.resolve(region="eu-west-1", max_attempts=5)
print(config.region)  # "eu-west-1" (not from env or file)
```

**Use a named profile:**

```python
config = await AsyncAwsConfig.resolve(profile="some_name")
```

If the profile doesn't exist:

```python
config = await AsyncAwsConfig.resolve(profile="nonexistent")
# raises ProfileNotFoundError: "Profile 'nonexistent' (from the profile argument)
#   not found in config file."
```

**Provenance tracking:**
```python
config = await AsyncAwsConfig.resolve()
config.source_of("region")        # ConfigSource.ENV
config.source_of("retry_mode")    # ConfigSource.DEFAULT
config.source_of("max_attempts")  # ConfigSource.PROFILE
```

**Override values after resolution:**

Values can be changed after resolution using plugins. The source updates to OVERRIDE and the new value is validated:

```python
config = await AsyncAwsConfig.resolve()
config.region = "ap-southeast-1"
config.source_of("region")  # ConfigSource.OVERRIDE

config.region = "some-value!"  # raises ConfigValidationError
```

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
@@ -97,6 +97,12 @@ class AsyncAwsConfig:
             credentials_file_path=credentials_file_path,
         )
 
+        # Validate profile exists if explicitly requested
+        if profile is not None:
+            config_file = await ctx.parsed_profiles()
+            if profile not in config_file.profiles:
+                raise ConfigError(f"Profile '{profile}' not found in config file.")
+
         # Create the instance bypassing __post_init__ check
         instance = cls._create_instance()
         instance._ctx = ctx
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
@@ -1,19 +1,20 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from collections.abc import Mapping
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Self
 
-from smithy_aws_core.config.context import FileSystem, SharedConfigContext
-from smithy_aws_core.config.exceptions import ConfigError
+from smithy_core.exceptions import ConfigError, ConfigValidationError
+from smithy_core.retries import RetryStrategyOptions
+
+from smithy_aws_core.config.context import SharedConfigContext
+from smithy_aws_core.config.filesystem import FileSystem
 from smithy_aws_core.config.resolvers import resolve_region, resolve_retry_config
 from smithy_aws_core.config.types import UNSET, ConfigSource, FieldSpec, Resolved
 from smithy_aws_core.config.validators import (
     validate_region,
     validate_retry_strategy_options,
 )
-from smithy_core.retries import RetryStrategyOptions
 
 
 @dataclass(kw_only=True)
@@ -62,7 +63,6 @@ class AsyncAwsConfig:
         cls,
         *,
         profile: str | None = None,
-        env: Mapping[str, str] | None = None,
         fs: FileSystem | None = None,
         config_file_path: str | None = None,
         credentials_file_path: str | None = None,
@@ -73,7 +73,6 @@ class AsyncAwsConfig:
         This is the only supported way to create a config instance.
 
         :param profile: Override the active profile name.
-        :param env: Override the environment variable mapping.
         :param fs: Override the filesystem abstraction.
         :param config_file_path: Override path for config file.
         :param credentials_file_path: Override path for credentials file.
@@ -82,7 +81,6 @@ class AsyncAwsConfig:
         """
         ctx = SharedConfigContext(
             profile_name=profile,
-            env=env,
             fs=fs,
             config_file_path=config_file_path,
             credentials_file_path=credentials_file_path,
@@ -126,7 +124,7 @@ class AsyncAwsConfig:
         """Run the resolution pipeline for all fields."""
         unknown = set(overrides) - set(self._FIELDS)
         if unknown:
-            raise ConfigError(
+            raise ConfigValidationError(
                 f"Unknown config field(s): {sorted(unknown)}. "
                 f"Valid fields are: {sorted(self._FIELDS)}"
             )
@@ -166,11 +164,14 @@ class AsyncAwsConfig:
 
     def __setattr__(self, name: str, value: Any) -> None:
         """Track provenance when fields are set with plugins after construction"""
-        super().__setattr__(name, value)
         # Mark as override only if the field is in _FIELDS and was already resolved
         if (
             name in self.__class__._FIELDS
             and hasattr(self, "_sources")
             and name in self._sources
         ):
+            spec = self.__class__._FIELDS[name]
+            if spec.validator is not None:
+                spec.validator(value)
             self._sources[name] = ConfigSource.OVERRIDE
+        super().__setattr__(name, value)
--- a/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
@@ -4,16 +4,21 @@
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Self
 
-from smithy_core.exceptions import ConfigError, ConfigValidationError
 from smithy_core.retries import RetryStrategyOptions
 
-from smithy_aws_core.config.context import SharedConfigContext
-from smithy_aws_core.config.filesystem import FileSystem
-from smithy_aws_core.config.resolvers import resolve_region, resolve_retry_config
-from smithy_aws_core.config.types import UNSET, ConfigSource, FieldSpec, Resolved
-from smithy_aws_core.config.validators import (
+from .context import SharedConfigContext
+from .exceptions import ConfigError, ConfigValidationError
+from .filesystem import FileSystem
+from .resolvers import (
+    resolve_max_attempts,
+    resolve_region,
+    resolve_retry_mode,
+)
+from .types import UNSET, ConfigSource, FieldSpec, Resolved
+from .validators import (
+    validate_max_attempts,
     validate_region,
-    validate_retry_strategy_options,
+    validate_retry_mode,
 )
 
 
@@ -29,7 +34,8 @@ class AsyncAwsConfig:
     """
 
     region: str | None = None
-    retry_strategy_options: RetryStrategyOptions | None = None
+    retry_mode: str | None = None
+    max_attempts: int | None = None
 
     _ctx: SharedConfigContext | None = field(default=None, repr=False, compare=False)
     _sources: dict[str, ConfigSource] = field(  # type: ignore[assignment]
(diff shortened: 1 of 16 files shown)


packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,86 @@
...
+    :raises ConfigValidationError: If the value is not a RetryStrategyOptions instance.
+    """
+    if not isinstance(value, RetryStrategyOptions):
+        raise ConfigValidationError(
+            f"Invalid value for 'retry_strategy_options': {value!r}. "
+            f"Must be RetryStrategyOptions, got {type(value).__name__}."
+        )
+
+
+def validate_retry_mode(retry_mode: str):
```
jonathan343, line 35: I'm realizing now that by only accepting `standard`, we don't allow customers to use existing config files and environments they use for products like boto3. I'm thinking if someone sets `legacy` can just map that to standard and throw a warning telling them that we don't support `legacy` (and never will). I think that is better than erroring out. Especially if they have legacy set for something like boto3 which does support it. Maybe doing the same with `adaptive` make sense until we actually support it. Because we have provenance now, we can event check: if they pass in `"adaptive"` to the config object itself, that is a hard failure as it is now. However, if it comes from the env or config file, we warn and default to standard as mentioned above. And of course, typos and garbage inputs should be completely rejected anywhere. 

Let me know if you think there is a better path forward

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,86 @@
...
+            "Invalid value for 'region': None. Region is required and must be set."
+        )
+    if not isinstance(value, str) or not _REGION_PATTERN.match(value):
+        raise ConfigValidationError(
+            f"Invalid value for 'region': {value!r}. "
+            "Must be a valid AWS region identifier."
+        )
+
+
+def validate_retry_strategy_options(value: object) -> None:
```
jonathan343: Edit: See [this comment](https://github.com/smithy-lang/smithy-python/pull/751#discussion_r3647774933) first 

This seems to allow `retry_strategy_options=RetryStrategyOptions(retry_mode="bogus", max_attempts=-5)` which is something we should reject. Why not re-use the validation methods below for mode and max attempts?

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
```diff
@@ -0,0 +1,86 @@
...
+    :param max_attempts: The max attempts value (string or int)
+
+    :raises ConfigValidationError: If the value is less than 1 or cannot be converted to an integer
+    """
+    try:
+        max_attempts = int(max_attempts)
+    except (ValueError, TypeError):
+        raise ConfigValidationError(
+            f"max_attempts must be a number, got {type(max_attempts).__name__}",
+        )
```
jonathan343, line 76: Anytime we catch an exception and raise a new one, we should use `raise ... from e`

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -0,0 +1,177 @@
...
+from smithy_aws_core.config.resolvers import resolve_region, resolve_retry_config
+from smithy_aws_core.config.types import UNSET, ConfigSource, FieldSpec, Resolved
+from smithy_aws_core.config.validators import (
+    validate_region,
+    validate_retry_strategy_options,
+)
+
+
+@dataclass(kw_only=True)
+class AsyncAwsConfig:
```
jonathan343, line 27: We should expose any public classes / methods we expect customers to use directly in `config/__init__.py` like below:

```python
from smithy_aws_core.config.aws_config import AsyncAwsConfig

__all__ = [
    "AsyncAwsConfig",
    ...
]
```

`AsyncAwsConfig` for sure should go there + anything else you think people will use directly.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Fields are resolved asynchronously from multiple sources (env vars,
+    config files, defaults) through the resolve() classmethod.
+
+    Do not instantiate directly — use:
+        config = await AsyncAwsConfig.resolve()
+    """
+
+    region: str | None = None
+    retry_strategy_options: RetryStrategyOptions | None = None
```
jonathan343: `retry_mode` and `max_attempts` are resolved independently but bundled into this one field, which causes two problems:
- Provenance: mixed sources collapse via `_strongest_source`, so `retry_mode` from env + `max_attempts` from profile reports just `ENV`.
- Validation: You are currently doing an isinstance check which doesn't validated the nested values that we really care about

I also think from the customer perspective, `retry_mode` and `max_attempts` is more straight forward and doesn't require you having to import and create a new `RetryStrategyOptions` object. 

Can we split them into two top-level fields? Is there something preventing that?

packages/smithy-core/src/smithy_core/exceptions.py
```diff
@@ -116,3 +116,15 @@ class UnsupportedStreamError(SmithyError):
 
 class EndpointResolutionError(SmithyError):
     """Exception type for all exceptions raised by endpoint resolution."""
+
```
jonathan343: I think you may have misunderstood what I was suggesting previously. The errors should live in `smithy_aws_core` since that's what uses them. However, they should inherit from `SmithyError` which lives in `smithy_core`. This pattern exists in smithy_aws_event_stream as well if that helps you better understand: https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-aws-event-stream/src/smithy_aws_event_stream/exceptions.py

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
jonathan343, line 1: nit, not blocking - the rest of this package seems to use relative imports instead of absolute ones. We should be consistent

jonathan343 (changes requested):

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
ubaskota: Yes, we can. I've split them into two separate top-level fields (`retry_mode` and `max_attempts`), each with its own resolver, validator, and provenance tracking. `RetryStrategyOptions` will be constructed on the client side from these individual values.

packages/smithy-aws-core/src/smithy_aws_core/config/validators.py
ubaskota: I agree with this. It has been updated to map "legacy" and "adaptive" to "standard" with a warning message.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -0,0 +1,188 @@
...
+    Fields are resolved asynchronously from multiple sources (env vars,
+    config files, defaults) through the resolve() classmethod.
+
+    Do not instantiate directly — use:
+        config = await AsyncAwsConfig.resolve()
+    """
+
+    region: str | None = None
+    retry_mode: str | None = None
+    max_attempts: int | None = None
```
jonathan343, line 39: We are promising to return `int | None` for `max_attempts`, however, this isn't true when a customer passes in a `str`. We cannot violate that contract.

```python
>>> config = await AsyncAwsConfig.resolve(max_attempts="5")
>>> type(config.max_attempts)
<class 'str'>
```

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -0,0 +1,188 @@
...
+        raise ConfigError(
+            f"{type(self).__name__} cannot be constructed directly. "
+            f"Use `await {type(self).__name__}.resolve(...)` instead."
+        )
+
+    @classmethod
+    async def resolve(
+        cls,
+        *,
+        profile: str | None = None,
```
jonathan343, line 77: When a customer provides a `profile` value that doesn't actually exist, we throw a misleading error:

```python
>>> c>>> config = await AsyncAwsConfig.resolve(profile="FOOBAR")  from smithy_aws_core.config import AsyncAwsConfig
>>> config = await AsyncAwsConfig.resolve(profile="FOOBAR")
  File "/Users/gytndd/.local/share/uv/python/cpython-3.12.13-macos-aarch64-none/lib/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/gytndd/.local/share/uv/python/cpython-3.12.13-macos-aarch64-none/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py", line 105, in resolve
    await instance._resolve_fields(overrides)
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py", line 165, in _resolve_fields
    spec.validator(getattr(self, field_name))
  File "/Users/gytndd/dev/GitHub/temp6/smithy-python/packages/smithy-aws-core/src/smithy_aws_core/config/validators.py", line 24, in validate_region
    raise ConfigValidationError(
smithy_aws_core.config.exceptions.ConfigValidationError: Invalid value for 'region': None. Region is required and must be set.
```

A common case will be when customers make a type in the profile name. We should provide a more accurate/helpful error.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    result = await _resolve_str(
+        ctx,
+        env_vars=("AWS_RETRY_MODE",),
+        profile_keys=("retry_mode",),
+    )
+    if result.value == "legacy":
+        warnings.warn(
+            "'legacy' retry mode is not supported, using 'standard' instead.",
+            DeprecationWarning,
```
jonathan343: Usually `DeprecationWarning` means something works for this package today and won't in the future. We should use `UserWarning` instead since this is by design.

jonathan343 (approved):
LGTM, thanks Ujjwal. 

There are some UX things we'll want to clean up. But this is a good start and we have other work that depends on this, so I'm approving.

PR #752, opened 2026-07-24
fix: handle explicit null member defaults for enums and intEnums

jonathan343:
## Summary

Fixes a Python codegen crash when a structure member overrides an enum or intEnum default with an explicit `@default(null)`.

## Problem

A member can set `@default(null)` to mark itself nullable even when its target enum declares a non-null default. This is valid Smithy that appears in real AWS models (e.g. `bedrock-agentcore-control`'s `UpdatePolicyRequest$enforcementMode`). In existing codegen, `StructureGenerator.getDefaultValue` sent these into the enum/intEnum branches, which read the default as a string/number without a null check and threw:

```
ExpectationNotMetException: Expected string, but found null.
```

## Solution

Guard for a null default at the top of `getDefaultValue` so every non-document type resolves to `None` before the typed branches run (documents are excluded, their null default is a non-None `Document(None)`).

Added `defaultNull{Enum,IntEnum,Blob,Timestamp}` members to the integration model to cover this path.

---

### DIFF
--- a/codegen/core/src/it/resources/META-INF/smithy/main.smithy
+++ b/codegen/core/src/it/resources/META-INF/smithy/main.smithy
@@ -381,6 +381,8 @@ structure Defaults {
     @required
     requiredDefaultBlob: Blob = "c3BhbQ=="
 
+    defaultNullBlob: Blob = null
+
     // timestamp
     @required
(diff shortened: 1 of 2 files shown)


arandito (approved):
:shipit:

PR #753, opened 2026-07-27
fix: avoid chunked framing for bodiless aiohttp requests

jonathan343:
## Summary

Fixes `AIOHTTPClient` requests that contain no payload but are transmitted as chunked streams.

This was observed with Amazon Polly’s `DescribeVoices` operation. The request reached the service, but Polly never returned response headers. The same operation completed normally when using `AWSCRTHTTPClient`.

## Root cause

Smithy represents an omitted request payload as an empty `AsyncBytesReader`. Before this change, `AIOHTTPClient` passed that reader to aiohttp as request data.

Because aiohttp cannot determine the length of an arbitrary async iterable, it selected chunked transfer encoding:

```
# removed unrelated additional headers
GET /v1/voices?IncludeAdditionalLanguageCodes=false HTTP/1.1
Host: polly.us-east-1.amazonaws.com
Content-Type: application/octet-stream
Transfer-Encoding: chunked
```

Although an empty chunked request is permitted by HTTP, Polly did not return response headers for the observed request shape. The client consequently continued waiting until canceled or timed out.

The CRT transport does not produce this HTTP/1.1 chunked framing. Its HTTP/1.1 path omits an empty buffered body stream, while HTTP/2 does not use chunked transfer encoding.

## Changes

`AIOHTTPClient` now distinguishes between an empty seekable body and a stream with data:

- Empty seekable bodies are passed to aiohttp as `None`.
- Nonempty bodies are restored to their original position before transmission.
- Nonseekable bodies continue through the existing streaming path.
- Requests with explicit `Content-Length` or `Transfer-Encoding` headers retain their existing framing behavior.
- No request body is read or buffered to determine whether it is empty.

For a bodyless operation, aiohttp now emits an ordinary request without payload framing:

```
# removed unrelated additional headers
GET /v1/voices?IncludeAdditionalLanguageCodes=false HTTP/1.1
Host: polly.us-east-1.amazonaws.com
```

## Verification

```python
import asyncio

from aws_sdk_polly.client import PollyClient
from aws_sdk_polly.config import Config
from aws_sdk_polly.models import DescribeVoicesInput


async def main() -> None:
    config = Config(
        region="us-east-1",
        transport=AIOHTTPClient(),
    )
    client = PollyClient(config=config)

    response = await client.describe_voices(DescribeVoicesInput())
    print(f"Received {len(response.voices or [])} voices")

if __name__ == "__main__":
    asyncio.run(main())
```

Observed behavior:

- Before: request stalls while waiting for response headers
- After:  `Received 15 voices`
---

### DIFF
--- a/packages/smithy-http/.changes/next-release/smithy-http-bugfix-4d75836f71714fef8792632f0de556de.json
+++ b/packages/smithy-http/.changes/next-release/smithy-http-bugfix-4d75836f71714fef8792632f0de556de.json
@@ -1,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Fixed `AIOHTTPClient` sending omitted request payloads with chunked transfer encoding."
+}
--- a/packages/smithy-http/src/smithy_http/aio/aiohttp.py
+++ b/packages/smithy-http/src/smithy_http/aio/aiohttp.py
@@ -90,8 +90,13 @@ class AIOHTTPClient(HTTPClient):
             chain.from_iterable(fld.as_tuples() for fld in request.fields)
         )
 
-        body: StreamingBlob = request.body
-        if not isinstance(body, AsyncBytesReader):
+        body: StreamingBlob | None = request.body
+        if (
+            "content-length" not in request.fields
+            and "transfer-encoding" not in request.fields
+        ):
+            body = await self._prepare_body(body)
+        elif not isinstance(body, AsyncBytesReader):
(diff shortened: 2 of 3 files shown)


Alan4506 (approved):
LGTM! Thanks @jonathan343!

jonathan343:
I have not fully convinced myself this is the most correct solution. But since it does fix the issue and is a 2-way door, I'm going to merge.

PR #758, opened 2026-07-30
Add support for remaining config variables from the old to-be-deprecated Config interface and add a new config class

ubaskota:
*Issue #, if available:*

*Description of changes:*
  Adds the remaining AWS-shared config fields to `AsyncAwsConfig`, bringing it to parity with the generated service Config class.

  - **New fields:** Adds support for `endpoint_uri, aws_access_key_id, aws_secret_access_key, aws_session_token, sdk_ua_app_id, user_agent_extra, interceptors, http_request_config, transport, retry_strategy, aws_credentials_identity_resolver`. Resolvable fields wire into the `env > profile > default` resolution pipeline.
  - **Service-specific codegen:** Generates `Async<ServiceId>Config(AsyncAwsConfig)` with service-specific `_FIELDS` that override the base class example: `endpoint_uri` uses a service-aware resolver that checks `AWS_ENDPOINT_URL_<SERVICE_ID>` and the services config section before falling back to global sources.
  - **Dual config support:** The generated config module now contains both the old `Config` (with a deprecation warning) and the new `Async<ServiceId>Config`, so existing users continue to work while new users adopt the async resolution path. The generated client accepts either type via `isinstance` dispatch.
  - **Supporting changes:** Adds `get_service_config()` on `MergedConfig` for services-section lookups, and updates `RetryStrategyResolver` to accept `retry_mode/max_attempts` fallbacks from the config layer.
  
  
  *Testing:*
  - Added unit tests for:
      - `EndpointUriResolver` covering the full precedence chain: `service-specific env var > global env var > service config   section > global profile > unset`.   
    - `MergedConfig.get_service_config()` covering all lookup paths (profile missing, services key missing, service section   not found, multiple services).   
    - `RetryStrategyResolver` fallback behavior: `retry_mode/max_attempts` params used when `retry_strategy` is None, explicit   strategy takes precedence over fallbacks.

*Example Usage:*

**Resolve service config and inspect provenance:**
```
# With AWS_REGION=us-east-1 set in the environment
# and ~/.aws/config containing:
#   [profile default]
#   services = my-services
#
#   [services my-services]
#   bedrock_runtime =
#     endpoint_url = https://bedrock-runtime.us-east-1.amazonaws.com
```

```python
import asyncio
from aws_sdk_bedrock_runtime.config import AsyncBedrockRuntimeConfig

async def main():
  config = await AsyncBedrockRuntimeConfig.resolve()

  print(config.region)                    # "us-east-1"
  print(config.source_of("region"))       # ENV
  print(config.endpoint_uri)              # "https://bedrock-runtime.us-east-1.amazonaws.com"
  print(config.source_of("endpoint_uri")) # PROFILE

asyncio.run(main())
```

**Invalid profile raises a clear error:**

```python
import asyncio
from aws_sdk_bedrock_runtime.config import AsyncBedrockRuntimeConfig

async def main():
  config = await AsyncBedrockRuntimeConfig.resolve(profile="non-existent")
  # raises ProfileNotFoundError:
  #   Profile 'non-existent' (from the profile argument) not found in config file.

asyncio.run(main())
```

Refer to https://github.com/smithy-lang/smithy-python/pull/751 for more examples.

### DIFF
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
@@ -68,12 +68,10 @@ public class AwsAsyncConfigIntegration implements PythonIntegration {
             // to emit references to these classes. If it says no symbol is generated, we
             // must not define one, or the two would disagree.
             var maybeAsyncConfigSymbol = CodegenUtils.getAsyncConfigSymbol(context.settings(), model);
-            var maybeAsyncPluginSymbol = CodegenUtils.getAsyncPluginSymbol(context.settings(), model);
-            if (maybeAsyncConfigSymbol.isEmpty() || maybeAsyncPluginSymbol.isEmpty()) {
+            if (maybeAsyncConfigSymbol.isEmpty()) {
                 return;
             }
             var asyncConfigSymbol = maybeAsyncConfigSymbol.get();
-            var asyncPluginSymbol = maybeAsyncPluginSymbol.get();
 
             final String serviceId = service.getTrait(ServiceTrait.class)
                     .map(ServiceTrait::getSdkId)
@@ -172,19 +170,11 @@ public class AwsAsyncConfigIntegration implements PythonIntegration {
             // Write _FIELDS class variable with service-specific defaults
             writer.openBlock("_FIELDS: ClassVar[dict[str, $T]] = {", fieldSpecSymbol);
 
-            // Plugin-contributed FieldSpec entries.
-            //
-            // These are written *before* the base class spread on purpose. Plugins
-            // declare config properties for the legacy Config object, which has no
-            // base class, so some of them duplicate fields AsyncAwsConfig already
-            // owns and resolves (region, credentials, sdk_ua_app_id, ...). Emitting
-            // them first means the spread below wins for any such duplicate, so a
-            // bare FieldSpec(default=None) can never clobber a base spec that
-            // carries a resolver or validator. Properties the base doesn't declare
-            // (e.g. api_key for @httpApiKeyAuth) survive untouched.
-            //
-            // Reuse the set collected above so these entries stay in step with the
-            // field declarations and duplicate contributions are written once.
+            // Plugin-contributed FieldSpec entries are emitted before the base class
+            // spread. Some duplicate fields already in AsyncAwsConfig._FIELDS (e.g.,
+            // region, sdk_ua_app_id) — these are harmlessly overwritten by the spread
+            // below. Fields unique to this service (e.g., api_key from @httpApiKeyAuth)
+            // survive and participate in the resolution pipeline.
             for (String propertyName : writtenProperties) {
                 writer.write("\"$L\": $T(default=None),", propertyName, fieldSpecSymbol);
             }
@@ -261,18 +251,6 @@ public class AwsAsyncConfigIntegration implements PythonIntegration {
 
             writer.closeBlock("}");
             writer.closeBlock("");
-
-            // Generate the async plugin type alias
-            writer.addStdlibImport("typing", "Callable");
-            writer.addStdlibImport("typing", "TypeAlias");
-            writer.write("");
-            writer.write("");
-            writer.write("$L: TypeAlias = Callable[[$L], None]",
-                    asyncPluginSymbol.getName(),
-                    asyncConfigSymbol.getName());
-            writer.writeDocs(
-                    "A callable that allows customizing the async config object on each request.",
-                    context);
         }
 
         private static void writeAsyncDefaultAuthSchemes(GenerationContext context, PythonWriter writer) {
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
@@ -1,0 +1,320 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.python.aws.codegen;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import software.amazon.smithy.aws.traits.ServiceTrait;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.knowledge.ServiceIndex;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.node.ArrayNode;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.python.codegen.CodegenUtils;
+import software.amazon.smithy.python.codegen.ConfigProperty;
+import software.amazon.smithy.python.codegen.GenerationContext;
+import software.amazon.smithy.python.codegen.RuntimeTypes;
+import software.amazon.smithy.python.codegen.SmithyPythonDependency;
+import software.amazon.smithy.python.codegen.integrations.PythonIntegration;
+import software.amazon.smithy.python.codegen.integrations.RuntimeClientPlugin;
+import software.amazon.smithy.python.codegen.sections.AsyncConfigSection;
+import software.amazon.smithy.python.codegen.writer.PythonWriter;
+import software.amazon.smithy.utils.CodeInterceptor;
+import software.amazon.smithy.utils.CodeSection;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * AWS integration that generates the async config subclass (e.g., AsyncBedrockRuntimeConfig)
+ * inheriting from AsyncAwsConfig with service-specific fields and defaults.
+ */
+@SmithyInternalApi
+public class AwsAsyncConfigIntegration implements PythonIntegration {
+
+    @Override
+    public List<? extends CodeInterceptor<? extends CodeSection, PythonWriter>> interceptors(
+            GenerationContext context
+    ) {
+        return List.of(new AsyncConfigInterceptor(context));
+    }
+
+    private static final class AsyncConfigInterceptor
+            implements CodeInterceptor<AsyncConfigSection, PythonWriter> {
+
+        private final GenerationContext context;
+
+        AsyncConfigInterceptor(GenerationContext context) {
+            this.context = context;
+        }
+
+        @Override
+        public Class<AsyncConfigSection> sectionType() {
+            return AsyncConfigSection.class;
+        }
+
+        @Override
+        public void write(PythonWriter writer, String previousText, AsyncConfigSection section) {
+            // Write any previous content first
+            writer.write(previousText);
+
+            var model = context.model();
+            var service = context.settings().service(model);
+
+            // Gate on the same source of truth the core generators use to decide whether
+            // to emit references to these classes. If it says no symbol is generated, we
+            // must not define one, or the two would disagree.
+            var maybeAsyncConfigSymbol = CodegenUtils.getAsyncConfigSymbol(context.settings(), model);
+            var maybeAsyncPluginSymbol = CodegenUtils.getAsyncPluginSymbol(context.settings(), model);
+            if (maybeAsyncConfigSymbol.isEmpty() || maybeAsyncPluginSymbol.isEmpty()) {
+                return;
+            }
+            var asyncConfigSymbol = maybeAsyncConfigSymbol.get();
+            var asyncPluginSymbol = maybeAsyncPluginSymbol.get();
+
+            final String serviceId = service.getTrait(ServiceTrait.class)
+                    .map(ServiceTrait::getSdkId)
+                    .orElse(context.settings().service().getName());
+
+            // Import AsyncAwsConfig base class
+            var asyncAwsConfigSymbol = Symbol.builder()
+                    .name("AsyncAwsConfig")
+                    .namespace("smithy_aws_core.config.aws_config", ".")
+                    .addDependency(AwsPythonDependency.SMITHY_AWS_CORE)
+                    .build();
+
+            // Import FieldSpec and ClassVar
+            var fieldSpecSymbol = Symbol.builder()
+                    .name("FieldSpec")
+                    .namespace("smithy_aws_core.config.types", ".")
+                    .addDependency(AwsPythonDependency.SMITHY_AWS_CORE)
+                    .build();
+            writer.addStdlibImport("typing", "ClassVar");
+            writer.addStdlibImport("typing", "Any");
+            writer.addStdlibImport("dataclasses", "dataclass");
+
+            writer.write("");
+            writer.write("");
+            // repr=False is required: AsyncAwsConfig defines a __repr__ that filters out
+            // credential fields, and a generated __repr__ on this subclass would shadow it
+            // and leak secrets.
+            writer.write("@dataclass(kw_only=True, repr=False)");
+            writer.openBlock("class $L($T):", asyncConfigSymbol.getName(), asyncAwsConfigSymbol);
+            writer.writeDocs(serviceId + " configuration (async-resolved).", context);
+            writer.write("");
+
+            // Write service-specific field declarations
+            writer.write("endpoint_resolver: $T | None = None", RuntimeTypes.ENDPOINT_RESOLVER);
+            writer.writeDocs("The endpoint resolver used to resolve the final endpoint per-operation "
+                    + "based on the configuration.", context);
+            writer.write("");
+
+            writer.write("protocol: $T | None = None",
+                    Symbol.builder()
+                            .name("ClientProtocol[Any, Any]")
+                            .addReference(Symbol.builder()
+                                    .name("ClientProtocol")
+                                    .namespace("smithy_core.aio.interfaces", ".")
+                                    .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                                    .build())
+                            .build());
+            writer.writeDocs("The protocol to serialize and deserialize requests with.", context);
+            writer.write("");
+
+            var serviceIndex = ServiceIndex.of(context.model());
+            var hasAuth = !serviceIndex.getAuthSchemes(context.settings().service()).isEmpty();
+
+            if (hasAuth) {
+                writer.write("auth_schemes: dict[$T, $T] | None = None",
+                        RuntimeTypes.SHAPE_ID,
+                        Symbol.builder()
+                                .name("AuthScheme[Any, Any, Any, Any]")
+                                .addReference(Symbol.builder()
+                                        .name("AuthScheme")
+                                        .namespace("smithy_core.aio.interfaces.auth", ".")
+                                        .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                                        .build())
+                                .build());
+                writer.writeDocs("A map of auth scheme ids to auth schemes.", context);
+                writer.write("");
+
+                writer.write("auth_scheme_resolver: $T | None = None",
+                        CodegenUtils.getHttpAuthSchemeResolverSymbol(context.settings()));
+                writer.writeDocs("An auth scheme resolver that determines the auth scheme "
+                        + "for each operation.", context);
+                writer.write("");
+            }
+
+            // Plugin-contributed field declarations (e.g., api_key for @httpApiKeyAuth).
+            //
+            // More than one plugin can contribute the same property — region, for
+            // instance, comes from both the auth and regional-endpoints integrations
+            // — so track the names already written and emit each only once.
+            var writtenProperties = new LinkedHashSet<String>();
+            for (PythonIntegration integration : context.integrations()) {
+                for (RuntimeClientPlugin plugin : integration.getClientPlugins(context)) {
+                    if (plugin.matchesService(model, service)) {
+                        for (ConfigProperty property : plugin.getConfigProperties()) {
+                            if (!writtenProperties.add(property.name())) {
+                                continue;
+                            }
+                            writer.write("$L: $T | None = None", property.name(), property.type());
+                            writer.writeDocs(property.documentation(), context);
+                            writer.write("");
+                        }
+                    }
+                }
+            }
+
+            // Write _FIELDS class variable with service-specific defaults
+            writer.openBlock("_FIELDS: ClassVar[dict[str, $T]] = {", fieldSpecSymbol);
+
+            // Plugin-contributed FieldSpec entries.
+            //
+            // These are written *before* the base class spread on purpose. Plugins
+            // declare config properties for the legacy Config object, which has no
+            // base class, so some of them duplicate fields AsyncAwsConfig already
+            // owns and resolves (region, credentials, sdk_ua_app_id, ...). Emitting
+            // them first means the spread below wins for any such duplicate, so a
+            // bare FieldSpec(default=None) can never clobber a base spec that
+            // carries a resolver or validator. Properties the base doesn't declare
+            // (e.g. api_key for @httpApiKeyAuth) survive untouched.
+            //
+            // Reuse the set collected above so these entries stay in step with the
+            // field declarations and duplicate contributions are written once.
+            for (String propertyName : writtenProperties) {
+                writer.write("\"$L\": $T(default=None),", propertyName, fieldSpecSymbol);
+            }
+
+            writer.write("**$T._FIELDS,", asyncAwsConfigSymbol);
+
+            // Everything below deliberately overrides the base class and so must
+            // stay after the spread.
+
+            // endpoint_uri FieldSpec — overrides base class with service-aware resolver
+            var endpointUriResolverSymbol = Symbol.builder()
+                    .name("EndpointUriResolver")
+                    .namespace("smithy_aws_core.config.resolvers", ".")
+                    .addDependency(AwsPythonDependency.SMITHY_AWS_CORE)
+                    .build();
+            var snakeCaseServiceId = serviceId.replace(" ", "_").toLowerCase();
+            writer.write("\"endpoint_uri\": $T(", fieldSpecSymbol);
+            writer.indent();
+            writer.write("default=None,");
+            writer.write("resolver=$T($S),", endpointUriResolverSymbol, snakeCaseServiceId);
+            writer.dedent();
+            writer.write("),");
+
+            // endpoint_resolver FieldSpec
+            var endpointPrefix = service.getTrait(ServiceTrait.class)
+                    .map(ServiceTrait::getEndpointPrefix)
+                    .orElse(context.settings().service().getName());
+            writer.write("\"endpoint_resolver\": $T(", fieldSpecSymbol);
+            writer.indent();
+            writer.write("default_factory=lambda: $T(endpoint_prefix=$S),",
+                    AwsRuntimeTypes.STANDARD_REGIONAL_ENDPOINTS_RESOLVER,
+                    endpointPrefix);
+            writer.dedent();
+            writer.write("),");
+
+            // protocol FieldSpec
+            writer.write("\"protocol\": $T(", fieldSpecSymbol);
+            writer.indent();
+            writer.write("default_factory=lambda: ${C|},",
+                    writer.consumer(w -> context.protocolGenerator().initializeProtocol(context, w)));
+            writer.dedent();
+            writer.write("),");
+
+            // auth_schemes FieldSpec
+            if (hasAuth) {
+                writer.write("\"auth_schemes\": $T(", fieldSpecSymbol);
+                writer.indent();
+                writer.write("default_factory=lambda: ${C|},",
+                        writer.consumer(w -> writeAsyncDefaultAuthSchemes(context, w)));
+                writer.dedent();
+                writer.write("),");
+
+                // auth_scheme_resolver FieldSpec
+                writer.write("\"auth_scheme_resolver\": $T(", fieldSpecSymbol);
+                writer.indent();
+                writer.write("default_factory=$T,",
+                        CodegenUtils.getHttpAuthSchemeResolverSymbol(context.settings()));
+                writer.dedent();
+                writer.write("),");
+            }
+
+            // transport FieldSpec
+            writer.write("\"transport\": $T(", fieldSpecSymbol);
+            writer.indent();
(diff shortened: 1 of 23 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -72,10 +72,23 @@ private void generateService(PythonWriter writer) {
...
+                        self,
+                        config: $1T | $6T | None = None,
+                        plugins: list[$2T] | None = None,
+                    ):
                         $3C
-                        self._config = config or $1T()
+                        if isinstance(config, $6T):
+                            self._config: $1T = config  # type: ignore[assignment]
+                        elif isinstance(config, $1T) or config is None:
+                            self._config = config or $1T()
```
arandito, line 117: The default behavior constructs the deprecated `Config` object and subsequently throws a deprecation warning to users even though they never explicitly use it. IMO it's an awkward user interface to raise deprecation warnings for default behavior, especially when we are trying to get customers to start using the new config. What's stopping us from defaulting to `AsyncAwsConfig.resolve()` or `Async<ServiceId>Config.resolve()`?"

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -385,6 +419,162 @@ private void initializeProperties(PythonWriter writer, Collection<ConfigProperty
...
+     */
+    private void generateAsyncConfig(GenerationContext context, PythonWriter writer, Symbol asyncConfigSymbol) {
+        var model = context.model();
+        var service = context.settings().service(model);
+        final String serviceId = service.getTrait(ServiceTrait.class)
+                .map(ServiceTrait::getSdkId)
+                .orElse(context.settings().service().getName());
+
+        // Import AsyncAwsConfig base class
+        writer.addDependency(SmithyPythonDependency.SMITHY_AWS_CORE);
```
arandito: This import couples our generic code generator with AWS specifics and forces a `smithy-aws-core` import. Can we use interceptor system and an AWS integration to write this section instead?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -72,10 +72,23 @@ private void generateService(PythonWriter writer) {
...
-                    def __init__(self, config: $1T | None = None, plugins: list[$2T] | None = None):
+                    def __init__(
+                        self,
+                        config: $1T | $6T | None = None,
+                        plugins: list[$2T] | None = None,
+                    ):
                         $3C
-                        self._config = config or $1T()
+                        if isinstance(config, $6T):
+                            self._config: $1T = config  # type: ignore[assignment]
```
arandito: This generates:
```
if isinstance(config, AsyncBedrockRuntimeConfig):
    self._config: Config = config  # type: ignore[assignment]
```
If the instance is a `AsyncBedrockRuntimeConfig` object, why are we setting the `self._config` type hint to the deprecated `Config` object?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -385,6 +419,162 @@ private void initializeProperties(PythonWriter writer, Collection<ConfigProperty
...
+                .build();
+        writer.addStdlibImport("typing", "ClassVar");
+        writer.addStdlibImport("typing", "Any");
+        writer.addStdlibImport("dataclasses", "dataclass");
+
+        writer.write("");
+        writer.write("");
+        writer.write("@dataclass(kw_only=True)");
+        writer.openBlock("class $L($T):", asyncConfigSymbol.getName(), asyncAwsConfigSymbol);
+        writer.write("\"\"\"$L configuration (async-resolved).\"\"\"", serviceId);
```
arandito: nit: We should use `writeDocs` here.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -385,6 +419,162 @@ private void initializeProperties(PythonWriter writer, Collection<ConfigProperty
...
+        writer.write("");
+
+        // Write service-specific field declarations
```
arandito: These fields are missing docstrings like the old `Config` object.

Adding them will help with IDE support and for documentation.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -385,6 +419,162 @@ private void initializeProperties(PythonWriter writer, Collection<ConfigProperty
...
+        writer.indent();
+        writer.write("default_factory=lambda: ${C|},",
+                writer.consumer(w -> writeAsyncDefaultAuthSchemes(context, w)));
+        writer.dedent();
+        writer.write("),");
+
+        // auth_scheme_resolver FieldSpec
+        writer.write("\"auth_scheme_resolver\": $T(", fieldSpecSymbol);
+        writer.indent();
+        writer.write("default_factory=HTTPAuthSchemeResolver,");
```
arandito: nit: We should use `CodegenUtils.getHttpAuthSchemeResolverSymbol(context.settings()` instead of hardcoding it to remain consistent.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
arandito, line 1: `IntEnumGenerator`, `EnumGenerator` and `UnionGenerator` all have formatting only changes. Are these intentional?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -249,7 +264,9 @@ private void writeSharedOperationInit(
                     raise $2T("protocol and transport MUST be set on the config to make calls.")
 
                 retry_strategy = await self._retry_strategy_resolver.resolve_retry_strategy(
-                    retry_strategy=config.retry_strategy
+                    retry_strategy=config.retry_strategy,
+                    retry_mode=getattr(config, "retry_mode", None),
```
arandito: Why not `config.retry_mode` and `config.max_attempts` instead of `getattr`?

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -37,6 +50,17 @@ class AsyncAwsConfig:
...
     retry_mode: str | None = None
     max_attempts: int | None = None
+    endpoint_uri: "str | URI | None" = None
+    aws_access_key_id: str | None = field(default=None, repr=False)
+    aws_secret_access_key: str | None = field(default=None, repr=False)
+    aws_session_token: str | None = field(default=None, repr=False)
+    aws_credentials_identity_resolver: "IdentityResolver[AWSCredentialsIdentity, AWSIdentityProperties] | None" = None
+    sdk_ua_app_id: str | None = None
+    user_agent_extra: str | None = None
+    interceptors: list[Any] = field(default_factory=list)  # type: ignore
```
arandito, line 101: If the interceptors list type depends on service specific interceptors, should we define this field in the service config object instead?

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -183,8 +250,88 @@ def _apply_default(self, field_name: str, spec: FieldSpec) -> None:
...
         self._sources[field_name] = ConfigSource.DEFAULT
 
+    async def _resolve_credentials(self, overrides: dict[str, Any]) -> None:
+        """Resolve credential fields atomically from a single source.
+
+        Rules:
+        - If both aws_access_key_id and aws_secret_access_key are overridden,
+          resolve normally
+        - If only one credential is overridden, raise ConfigValidationError.
+        - Otherwise, resolve atomically: if both key and secret are present in
```
arandito: We must not attempt to resolve env/profile credentials during config resolution. The `IdentityChain` handles both environment and profile credentials. Config resolution should only set and validate in-code credentials if `aws_access_key_id` and `aws_secret_access_key` are set.

If we do this, we would skip any credential source that must be checked after environment and before profile credentials (i.e. STS Assume Role with Web Identity).

What we could do if customers set credentials in-code is to automatically set `config.aws_credentials_identity_resolver = StaticCredentialsResolver()` for them. Currently, when customers want to define in-code credentials they also have to set their resolver to `StaticCredentialsResolver()`. To make the user experience easier, we can detect the in-code credentials, validate both are present and automatically set the resolver field to the static resolver. Let me know what you think about this.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -120,3 +120,82 @@ async def resolve_max_attempts(ctx: SharedConfigContext) -> Resolved[int | None]
...
+        ctx,
+        env_vars=("AWS_SDK_UA_APP_ID",),
+        profile_keys=("sdk_ua_app_id",),
+    )
+
+
+class EndpointUriResolver:
+    """Service-aware endpoint URI resolver.
+
+    Resolution order (first match wins):
```
arandito, line 161: Question: Do other SDKs also match the global env first before the service specific config file? Seems odd we take the global over a service specific url.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -37,6 +50,17 @@ class AsyncAwsConfig:
     region: str | None = None
```
arandito, line 52: None of these options have docstrings. Are we planning on adding them to ensure we have IDE support and documentation support?

Curious how much of a pain its going to be since this repo uses the old rst docstring format while our clients use the google style format.

arandito (commented):
Thanks @ubaskota! I left a couple comments but my biggest concern is how we are resolving environment and profile credentials during config resolution. Config resolution should only handle in-code credentials and defer env/profile credentials to the new `IdentityChain`. Let me know if you have any questions!

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
ubaskota: At the time of this implementation, the understanding was that `aws_access_key_id` and `aws_secret_access_key` would be resolved through the config pipeline, as the credential chain was still under development. I'll update it to only validate in-code credentials and auto-set `StaticCredentialsResolver` when both are present, leaving `env/profile` credential resolution to `IdentityChain`.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
ubaskota: Yes, this matches [botocore's order](https://github.com/boto/botocore/blob/731fa05213abd83fb6030be125127361e606d3ca/botocore/configprovider.py#L953). Environment variables (both service-specific and global) take precedence over config file entries.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -72,10 +72,23 @@ private void generateService(PythonWriter writer) {
             }
 
             writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
+            var asyncConfigSymbol = CodegenUtils.getAsyncConfigSymbol(context.settings(), context.model());
             writer.write("""
-                    def __init__(self, config: $1T | None = None, plugins: list[$2T] | None = None):
+                    def __init__(
+                        self,
+                        config: $1T | $6T | None = None,
+                        plugins: list[$2T] | None = None,
```
arandito: Something I realized after reviewing #733 is that we need to start accepting plugins that modify the new async config object and satisfy the new async plugin type. However, that raises an issue: how do we avoid breaking current plugins that use `Config`.

At the very least, we'll need to update our plugins to not strictly accept `Config` ([example](https://github.com/aws/aws-sdk-python/blob/develop/clients/aws-sdk-api-gateway/src/aws_sdk_api_gateway/accept_header.py)) or else existing clients will likely fail type checking.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
ubaskota: The `Config` type annotation was added to suppress pyright errors downstream. I'll change it to `Config | AsyncBedrockRuntimeConfig` instead.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/UnionGenerator.java
ubaskota: Yes, intentional. Those changes are from running spotlessApply to fix formatting violations flagged by the build.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
ubaskota: This is because `retry_mode` and `max_attempts` exist on `AsyncBedrockRuntimeConfig` (inherited from AsyncAwsConfig) but not on the old Config. Since the generated client must work with both old and new configs, `getattr` returns `None` when the attribute doesn't exist instead of raising an AttributeError.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
ubaskota: I think its simpler to leave it in the base class because every service uses it with the same [] default. The only benefit of moving it would be narrower type hints, which can always be aded on the generated subclass later. If you see more benefits on moving it, let’s discuss.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
ubaskota: Yes, I agree. After internal discussion, I updated the Plugin TypeAlias to accept both config types: `Plugin: TypeAlias = Callable[[Config | AsyncBedrockRuntimeConfig], None]` for AWS services. The generated plugin functions (like `aws_user_agent_plugin`) also accept both types now. For non-AWS services, the Plugin type remains `Callable[[Config], None]` since they don't have an async config class.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
ubaskota: We discussed internally and decided to default to `Async<ServiceId>Config`. The issue was that `__init__` can't call an `async` function, so we couldn't resolve it at construction time. Instead, config is set to `None` initially and later resolved via `_ensure_config()` on the first operation call. This way it defaults to the new async config without triggering the deprecation warning, and resolution still happens asynchronously.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -61,38 +61,70 @@ private void generateService(PythonWriter writer) {
...
-                }
-            }
-
             writer.addDependency(SmithyPythonDependency.SMITHY_CORE);
-            writer.write("""
-                    def __init__(self, config: $1T | None = None, plugins: list[$2T] | None = None):
-                        $3C
-                        self._config = config or $1T()
-
-                        client_plugins: list[$2T] = [
```
jonathan343, line 80: Why did you decide to move `client_plugins` this out of the client constructor? It's now generated inside of every operation which means we are re-allocating every time. Unless there is a good reason, I think this should stay in the class constructor as it exists today.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
```diff
@@ -0,0 +1,320 @@
...
+                    Symbol.builder()
+                            .name("ClientProtocol[Any, Any]")
+                            .addReference(Symbol.builder()
+                                    .name("ClientProtocol")
+                                    .namespace("smithy_core.aio.interfaces", ".")
+                                    .addDependency(SmithyPythonDependency.SMITHY_CORE)
+                                    .build())
+                            .build());
+            writer.writeDocs("The protocol to serialize and deserialize requests with.", context);
+            writer.write("");
```
jonathan343, line 123: There are multiple config options that get generated with trailing whitespace in their docstrings:

```
"""The protocol to serialize and deserialize requests with.    """
```

This should be:

```
"""The protocol to serialize and deserialize requests with.    """
```

Can you investigate this bug and compare with the existing `Config` object to see why there is this difference?

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
jonathan343, line 1: This package should have a brief but descriptive changelog entry for the changes being made in this PR. Our packages get version bumped based on pending entries. Right now you're relying on existing entries to get version bumped which we shouldn't do.

packages/smithy-core/src/smithy_core/aio/retries.py
jonathan343, line 1: Should also add an entry here to ensure this is released with the other changes.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
```diff
@@ -22,6 +22,19 @@
...
+                        ua_suffix=config.user_agent_extra,
+                        ua_app_id=config.sdk_ua_app_id,
+                        sdk_version=$3T,
+                        service_id=$4S,
+                    )
+                )
+            """;
+
+    // Variant for services without a generated async config, which must not be referenced.
+    private static final String USER_AGENT_PLUGIN_SYNC_ONLY = """
```
jonathan343, line 37: Which services won't have an async config? Shouldn't they all have the async config right now?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -263,16 +264,31 @@ public void run() {
...
         context.writerDelegator().useFileWriter(plugin.getDefinitionFile(), plugin.getNamespace(), writer -> {
             writer.addStdlibImport("typing", "Callable");
             writer.addStdlibImport("typing", "TypeAlias");
-            writer.write("$L: TypeAlias = Callable[[$T], None]", plugin.getName(), config);
+            if (asyncConfigForPlugin.isPresent()) {
+                writer.write("$L: TypeAlias = Callable[[$T | $T], None]",
+                        plugin.getName(), config, asyncConfigForPlugin.get());
+            } else {
+                writer.write("$L: TypeAlias = Callable[[$T], None]", plugin.getName(), config);
+            }
```
jonathan343, line 302: The plugin API introduced in this PR doesn't make sense to me. Currently I see the following get generated:

```python
AsyncBedrockRuntimePlugin: TypeAlias = Callable[[AsyncBedrockRuntimeConfig], None]
"""
A callable that allows customizing the async config object on each
request.
"""

Plugin: TypeAlias = Callable[[Config | AsyncBedrockRuntimeConfig], None]
"""A callable that allows customizing the config object on each request."""
```
1. I don't see `AsyncBedrockRuntimePlugin` actually being used or referenced anywhere.
2. The `Async` naming prefix seems misleading since there is no async work done by the plugins.
3. The `Callable[[Config | AsyncBedrockRuntimeConfig], None]` signature is not what we want. This will make all plugins need to accept both config options. See below for what I think it should be. 

IMO, during migration, we should generate the following:

```python
Plugin: TypeAlias = (
    Callable[[Config], None]
    | Callable[[AsyncBedrockRuntimeConfig], None]
)
```

After we remove support for `Config` it should just become:

```python
Plugin: TypeAlias = Callable[[AsyncBedrockRuntimeConfig], None]
```

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
```diff
@@ -0,0 +1,320 @@
...
+                            writer.write("$L: $T | None = None", property.name(), property.type());
+                            writer.writeDocs(property.documentation(), context);
+                            writer.write("");
+                        }
+                    }
+                }
+            }
+
+            // Write _FIELDS class variable with service-specific defaults
+            writer.openBlock("_FIELDS: ClassVar[dict[str, $T]] = {", fieldSpecSymbol);
```
jonathan343, line 171: This emits something like below:

```python
    _FIELDS: ClassVar[dict[str, FieldSpec]] = {
        "aws_credentials_identity_resolver": FieldSpec(default=None),
        "region": FieldSpec(default=None),
        "aws_access_key_id": FieldSpec(default=None),
        "aws_secret_access_key": FieldSpec(default=None),
        "aws_session_token": FieldSpec(default=None),
        "user_agent_extra": FieldSpec(default=None),
        "sdk_ua_app_id": FieldSpec(default=None),
        **AsyncAwsConfig._FIELDS,
        "endpoint_uri": FieldSpec(
            default=None, resolver=EndpointUriResolver("bedrock_runtime")
        ),
        "endpoint_resolver": FieldSpec(
            default_factory=lambda: StandardRegionalEndpointsResolver(
                endpoint_prefix="bedrock-runtime"
            )
        ),
        "protocol": FieldSpec(
            default_factory=lambda: RestJsonClientProtocol(
                _SCHEMA_AMAZON_BEDROCK_FRONTEND_SERVICE
            )
        ),
        "auth_schemes": FieldSpec(
            default_factory=lambda: {
                ShapeID("aws.auth#sigv4"): SigV4AuthScheme(service="bedrock")
            }
        ),
        "auth_scheme_resolver": FieldSpec(default_factory=HTTPAuthSchemeResolver),
        "transport": FieldSpec(default_factory=lambda: AWSCRTHTTPClient()),
    }
```

It's not clean to my why we're emitting inherited fields here that I though would be covered by `**AsyncAwsConfig._FIELDS,`. 

I was expecting to see something closer to:

```python
 _FIELDS = {
      **AsyncAwsConfig._FIELDS,
      "endpoint_uri": ...,
      "endpoint_resolver": ...,
      "protocol": ...,
      "auth_schemes": ...,
      "auth_scheme_resolver": ...,
      "transport": ...,
  }
```

jonathan343 (changes requested):
Thanks Ujjwal. I let some comments on the areas I'm most concerned about. Let me know if you have any questions.

Also, you need to rebase this PR with the latest from `develop`.

I'm still investigating some additional cleanup that probably should be done, but wanted to get you some feedback so you have something to work on in the meantime.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
ubaskota: With the new design for lazy resolution, I moved both `client_plugins` and `operation_plugins` into the operation methods. But you're right that client_plugins can remain inside the init function. I will move it back.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsUserAgentIntegration.java
ubaskota: All AWS services have async config. This is for non-AWS Smithy services (those without `@aws.api#service`), which don't get an async config generated, `getAsyncConfigSymbol` returns `Optional.empty()` for them.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
ubaskota: After internal discussion, we decided to drop support for the old `Config` for AWS services and only generate `Async<ServiceId>Config`. This simplifies the plugin type as Plugin is now `Callable[[AsyncBedrockRuntimeConfig], None]`, and `AsyncBedrockRuntimePlugin` has been removed.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
ubaskota: This is intentional. The base class (`AsyncAwsConfig` in `aws_config.py`) declares common config fields with their resolvers and validators. When `AwsAsyncConfigIntegration.java` generates the service-specific config, it iterates all registered integration plugins (like `AwsAuthIntegration`,`AwsUserAgentIntegration`) to emit config fields they contribute. Some of those plugins contribute fields that already exist in the base class, just like the duplicates here.

Functionally, these duplicates are harmless because if a config var is declared in the base class it will overwrite the duplicate ones with the proper specs (`resolvers` and `validators`). Fields that are unique to a service are not overwritten and will be used in config resolution. One option to prevent this was by filtering based on the config vars that are already in the base class, but that'd require us to hardcode the list of those config vars in codegen. That meant creating a second source of truth that needs to be in sync with the variables in the base class. For now, I chose to keep duplicates rather than have two sources of truth.

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -35,8 +50,67 @@ class AsyncAwsConfig:
...
+    """
+
+    aws_secret_access_key: str | None = field(default=None, repr=False)
+    """A secret access key that can be used to sign requests.
+
+    Must be set together with ``aws_access_key_id``.
+    """
+
+    aws_session_token: str | None = field(default=None, repr=False)
+    """An access key ID that identifies temporary security credentials."""
```
jonathan343, line 85: nit - This docstring describes a session token as "An access key ID," is incorrect and confusing since (an access key id is `aws_access_key_id`). I know this description comes from the existing config object, but something like `The session token used with temporary AWS security credentials.` might be more correct here.

packages/smithy-aws-core/src/smithy_aws_core/config/resolvers.py
```diff
@@ -92,6 +92,10 @@ async def resolve_retry_mode(ctx: SharedConfigContext) -> Resolved[str | None]:
         env_vars=("AWS_RETRY_MODE",),
         profile_keys=("retry_mode",),
     )
+
+    if result.value is not UNSET:
+        result = Resolved(value=result.value.lower(), source=result.source)
```
jonathan343, line 97: ~**`retry_mode` case-handling is inconsistent between env/profile and in-code overrides.** Values coming from `AWS_RETRY_MODE`/profile are lowercased here before validation, so `AWS_RETRY_MODE=STANDARD` works. In-code overrides skip this path: `resolve(retry_mode=...)` sets the raw value and `validate_retry_mode` compares case-sensitively against `("standard",)`, so `resolve(retry_mode="Standard")` (or `"STANDARD"`) raises `ConfigValidationError` even though the same string is accepted from the environment. Consider lowercasing the override too (or documenting the constraint).~

I will validate this finding and update w/ my own feedback

packages/smithy-aws-core/src/smithy_aws_core/config/aws_config.py
```diff
@@ -61,8 +135,59 @@ class AsyncAwsConfig:
...
+        Defined on the base class so that every subclass inherits the
+        filtering, rather than relying on each subclass to mark its own
+        credential fields ``repr=False``. Subclasses must be declared with
+        ``@dataclass(repr=False)`` so they inherit this instead of generating
+        their own ``__repr__``.
+        """
+        rendered = ", ".join(
+            f"{f.name}={getattr(self, f.name)!r}"
+            for f in fields(self)
+            if f.repr and f.name not in _CREDENTIAL_FIELDS
```
jonathan343, line 187: ~**`__repr__` still exposes credential material via `aws_credentials_identity_resolver`.** The filter redacts the three string credential fields but not the resolver field, which holds the same secret (the auto-wired `StaticCredentialsResolver` is built from `aws_access_key_id`/`aws_secret_access_key`). It happens not to leak today only because `StaticCredentialsResolver` has no custom `__repr__` and `AWSCredentialsIdentity` is only reachable through it — but `AWSCredentialsIdentity` is a plain `@dataclass` whose repr prints `secret_access_key`. Any resolver that reprs its identity (or a future dataclass-style resolver) would defeat the stated goal ("without exposing credential material").~

I will validate this finding and update w/ my own feedback

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -344,25 +367,66 @@ private void generateConfig(GenerationContext context, PythonWriter writer) {
...
         writer.addLocallyDefinedSymbol(configSymbol);
         writer.addStdlibImport("dataclasses", "dataclass");
-        writer.write("""
-                @dataclass(init=False)
-                class $L:
-                    \"""Configuration for $L.\"""
+        // This class is only deprecated where an async replacement is generated to point
+        // at. For services without one it remains the supported config class.
+        var asyncConfigSymbol = CodegenUtils.getAsyncConfigSymbol(context.settings(), context.model());
+        if (asyncConfigSymbol.isPresent()) {
```
jonathan343, line 373: ~**Dead branch: this deprecated-`Config` path is unreachable.** `generateConfig(...)` is only invoked when `getAsyncConfigSymbol(...).isEmpty()` (non-AWS services), yet this branch runs only when `asyncConfigSymbol.isPresent()` — recomputed from the same inputs. So the `.. deprecated::` shim + `DeprecationWarning` is never emitted. If the intent is to fully drop old `Config` for AWS services (per the commit message), this whole `if` branch is dead code and can be removed; if the intent was to still emit a deprecation shim, it is silently not happening.~

I will validate this finding and update w/ my own feedback

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsAsyncConfigIntegration.java
arandito: I may be missing something here, but why can't we just strip out the config property declarations from the runtime plugins in these aws integrations? We shouldn't keep the old codegen-defined properties if we are moving to maintaining these in our runtime libraries. We're effectively maintaining two sources of truth for AWS-specific config options now.

jonathan343 (commented):
I'm going to merge this into a new `feat/async-config` branch. There are additional changes we need to include in this PR to improve the user experience and not break existing plugins.

I will follow up with PRs on that branch with the fixes and then when we're ready, we will open one from `feat/async-config` to `develop`. This will provide us quicker iteration that since @ubaskota's availability this week will vary.

PR #759, opened 2026-07-30
Update ChainSetup and IdentityChain interface

arandito:
### Description

This PR refactors `ChainSetup` to store the resolved profile name instead of a parsed profile `Section`.

Downstream providers (e.g. assume-role) need `ChainSetup` to store the resolved profile name for cycle tracking. Storing the parsed `Section` alongside it would make the profile reachable two ways, `setup.profile` and `setup.profile_file.get_profile(setup.profile_name)`, duplicating state. Using `profile_name` as the single source of truth lets providers read values consistently through `MergedConfig.get(profile_name, key)`. That call also performs the string validation the previous `_get_string` helpers provided, so each provider no longer needs its own copy of that logic.

This PR also renames `profile_file`/`profile_name_override` to `config_file`/`profile_name` for consistency, and updates the `EcsContainer` module suggestion from `aws-credentials-ecs` to `aws-credentials-http`.

### Testing
- Updated tests with new interfaces
- Ran `make check-py` and `make test-py`.

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
@@ -2,7 +2,6 @@
 #  SPDX-License-Identifier: Apache-2.0
 from smithy_core.interfaces.identity import Identity
 
-from ....config.file_parser import Section
 from ...components import AWSCredentialsIdentity
 from ...static import StaticCredentialsResolver
 from ..ordering import Standard, StandardProvider
@@ -12,11 +11,6 @@ _ACCESS_KEY_ID = "aws_access_key_id"
 _SECRET_ACCESS_KEY = "aws_secret_access_key"  # noqa: S105
 _SESSION_TOKEN = "aws_session_token"  # noqa: S105
 _ACCOUNT_ID = "aws_account_id"
-
-
-def _get_string(profile: Section, key: str) -> str | None:
-    value = profile.properties.get(key)
-    return value if isinstance(value, str) else None
 
 
 class ProfileSessionCredentialsProvider:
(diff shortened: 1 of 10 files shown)


packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
```diff
@@ -74,18 +69,19 @@ async def setup(self, identity_type: type[Identity], setup: ChainSetup) -> None:
...
-        if profile is None:
+        config_file = setup.config_file
+        profile_name = setup.profile_name
+        if config_file is None or profile_name is None:
             return
 
-        access_key_id = _get_string(profile, _ACCESS_KEY_ID)
-        secret_access_key = _get_string(profile, _SECRET_ACCESS_KEY)
+        access_key_id = config_file.get(profile_name, _ACCESS_KEY_ID)
+        secret_access_key = config_file.get(profile_name, _SECRET_ACCESS_KEY)
```
Alan4506, line 78: This runs before `ProfileAssumeRole` and terminates assembly, so a profile with both static keys and `role_arn` resolves to the raw user keys and the role is never assumed. This breaks the standard self-referencing pattern, where the keys are only the source credentials for the STS call:

```ini
[profile deploy]
role_arn = arn:aws:iam::999999999999:role/Deployer
source_profile = deploy
aws_access_key_id = AKIA...          # near-zero-permission user, only sts:AssumeRole
aws_secret_access_key = ...          # real permissions live on the role
```

botocore gives `role_arn` precedence: botocore's assume-role provider runs before the static-keys providers and claims any profile containing `role_arn`, regardless of static keys ([`_has_assume_role_config_vars`](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L1625-L1634)).

Let me know what you think.

arandito: Boto3's credential chain order is legacy and places assume role credentials directly after environment ([docs](https://docs.aws.amazon.com/boto3/latest/guide/credentials.html)).

However, modern SDKs must follow the order defined in the `StandardProvider` enum. This means that self-referencing profiles will not be respected if used in a chain.

They are still respected if you use the assume role resolver on its own.

jonathan343 (approved):
LGTM

PR #764, opened 2026-08-06
smithy-http: disable redirects in aiohttp client

arandito:
### Description
This PR disables automatic HTTP redirect following in the aiohttp client by passing `allow_redirects=False` to `self._session.request(...)`. This ensures that 3xx responses are returned to the caller instead of being followed by the transport.

This aligns the aiohttp transport with SDK convention of not following redirects by default (e.g. [botocore](https://github.com/boto/botocore/blob/develop/botocore/httpsession.py#L477-L487) via `Retry(False)`, and [smithy-kotlin](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt#L117-L119) via OkHttp `followRedirects(false)`). The CRT transport is not updated because the CRT client already does not follow redirects (no behavior or option exists).

### Testing
`make check-py` and `make test-py` pass (1793 passed, 9 skipped).


jonathan343 (approved):

jonathan343:
LGTM. If anyone ever needs this behavior, we can make it an opt in flag. Fine with it as is for now.

PR #765, opened 2026-08-09
Add async client transport cleanup

jonathan343:
## Overview

This PR adds deterministic resource cleanup to generated clients and HTTP transports. Generated clients can now close transport resources explicitly or through async context managers, while closing an unused client does not trigger lazy config resolution or credential I/O.

## Current State

Generated clients do not expose a lifecycle API for their transport. aiohttp sessions and CRT pooled connections can remain open unless callers reach into the configured transport and close it themselves, which can produce `Unclosed client session` and `Unclosed connector` warnings during shutdown.

## New Pattern

Generated clients and HTTP transports now support idempotent `close()` methods and async context managers. A generated client closes its configured transport after setup, rejects operations and context-manager re-entry after closure, and avoids initializing config or transport resources solely to close an unused client.

`AIOHTTPClient.close()` closes its `ClientSession`. `AWSCRTHTTPClient.close()` attempts to close every pooled connection and clears the pool even when an individual connection close fails. Both transports reject sends after closure.

If a service models an operation named `Close`, the operation is generated as `close_()` so it does not collide with the lifecycle `close()` method.

```python
async with AsyncExampleClient(config) as client:
    response = await client.some_operation(input)
```

Clients can also be closed manually:

```python
client = AsyncExampleClient(config)
try:
    response = await client.some_operation(input)
finally:
    await client.close()
```

---

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -143,10 +143,12 @@ final class ClientGenerator implements Runnable {
                         \"\"\"Close this client and any resources held by its transport.\"\"\"
                         if self._closed:
                             return
-                        self._closed = True
-                        await self._ensure_setup()
-                        assert self._config is not None
-                        await $1T(self._config.transport)
+                        async with self._derive_lock:
+                            if self._closed:
+                                return
+                            self._closed = True
+                            if self._setup_done and self._config is not None:
+                                await $1T(self._config.transport)
 
                     async def __aenter__(self) -> Self:
                         if self._closed:
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -134,6 +134,29 @@ final class ClientGenerator implements Runnable {
                         w.popState();
                     }));
 
+            writer.addStdlibImport("typing", "Any");
+            writer.write("""
+
+                    async def close(self) -> None:
+                        \"\"\"Close any resources held by this client's transport.\"\"\"
+                        await self._ensure_setup()
+                        assert self._config is not None
+                        await $1T(self._config.transport)
+
+                    async def __aenter__(self) -> "$2L":
+                        return self
+
+                    async def __aexit__(
+                        self,
+                        exc_type: Any,
+                        exc_value: Any,
+                        traceback: Any,
+                    ) -> None:
+                        await self.close()
+                    """,
+                    RuntimeTypes.ASYNC_CLOSE,
+                    serviceSymbol.getName());
+
             var topDownIndex = TopDownIndex.of(model);
             var eventStreamIndex = EventStreamIndex.of(model);
             for (OperationShape operation : topDownIndex.getContainedOperations(service)) {
@@ -289,7 +312,10 @@ final class ClientGenerator implements Runnable {
                         assert self._config is not None
                         if operation_plugins:
                             # Keep operation-plugin mutations scoped to this call.
-                            config = deepcopy(self._config)
+                            config = deepcopy(
+                                self._config,
+                                {id(self._config.transport): self._config.transport},
+                            )
                             for plugin in operation_plugins:
                                 plugin(config)
                         else:
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
@@ -94,6 +94,7 @@ final class ClientGenerator implements Runnable {
                         self._plugins = plugins
                         self._derive_lock = asyncio.Lock()
                         self._setup_done = False
+                        self._closed = False
                         self._retry_strategy_resolver = $4T()
                         self._client_plugins: list[$2T] = [
                             ${5C|}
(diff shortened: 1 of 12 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
```diff
@@ -242,7 +263,10 @@ private void writeSharedOperationInit(
                 ]
                 if plugins:
                     operation_plugins.extend(plugins)
-                config = deepcopy(self._config)
+                config = deepcopy(
+                    self._config,
+                    {id(self._config.transport): self._config.transport},
```
arandito: [Non-blocking, general comment]
I'm noticing some inconsistency in how we are preserving certain config fields. Our retries strategies each define their own `__deepcopy__` override to return the same instance. This works fine for the retry strategies we maintain but if customers provide their own custom implementations, they will need to always implement their own override. 

This issue also arises for the `config.aws_credentials_identity_resolver` field. We need to preserve the same instance to actually reuse cached credentials. I like your approach of modifying the memo dict in the `deepcopy()` call. I will probably add a memo entry for the identity resolver field in my upcoming PR that integrates the chain into the client.

As a follow up, we should see if its possible to do all of this work inside the config object's own `__deepcopy__`. This should centralize the logic and ensure any retry strategy, transport, or identity resolver is preserved, plus any future fields.

We can wait to do this until we deprecate the existing `Config` object in favor of the new `AsyncAwsConfig`.

packages/smithy-http/src/smithy_http/aio/aiohttp.py
```diff
@@ -159,7 +168,4 @@ async def _marshal_response(
         )
 
     def __deepcopy__(self, memo: Any) -> "AIOHTTPClient":
-        return AIOHTTPClient(
-            client_config=deepcopy(self._config),
-            _session=copy(self._session),
-        )
+        return self
```
arandito: If we already codegen this:
```
config = deepcopy(
    self._config,
    {id(self._config.transport): self._config.transport},
)
```
Do we need to change the override to return itself?

Alan4506 (commented):
Thanks @jonathan343! LGTM overall. Just have a non-blocking question:

`AIOHTTPClient.close()` and `AWSCRTHTTPClient.close()` release different things. The aiohttp one closes the ClientSession, while the CRT one closes and clears the pooled connections. I understand why: aiohttp only exposes `session.close()`, and a closed session cannot be reopened; CRT has no session equivalent, and connections are the only thing it can release.

But the problem is: the `close()` on a generated client now behaves differently per transport. For example, I verified the divergence with the following script:
```python
import asyncio

from smithy_http.aio.crt import AWSCRTHTTPClient

from aws_sdk_apigateway.client import AsyncAPIGatewayClient
from aws_sdk_apigateway.config import Config
from aws_sdk_apigateway.models import GetAccountInput

REGION = "us-east-1"


async def check(label, transport):
    client = AsyncAPIGatewayClient(
        config=Config(
            endpoint_uri=f"https://apigateway.{REGION}.amazonaws.com",
            region=REGION,
            transport=transport,
        )
    )
    await client.get_account(GetAccountInput())
    print(f"{label}: call before close -> ok")
    await client.close()
    try:
        await client.get_account(GetAccountInput())
        print(f"{label}: call after close  -> ok (no error raised)")
    except Exception as e:
        print(f"{label}: call after close  -> {type(e).__name__}: {str(e)[:60]}")


async def main():
    await check("aiohttp", AIOHTTPClient())
    await check("crt    ", AWSCRTHTTPClient())


asyncio.run(main())
```
Output:
```
aiohttp: call before close -> ok
aiohttp: call after close  -> SmithyError: Session is closed
crt    : call before close -> ok
crt    : call after close  -> ok (no error raised)
```

Do you think this divergence is acceptable? Do we need to document the divergence with some comments?

packages/smithy-http/src/smithy_http/aio/crt.py
```diff
@@ -364,7 +376,4 @@ async def _create_body_generator(
                     yield chunk
 
     def __deepcopy__(self, memo: Any) -> "AWSCRTHTTPClient":
-        return AWSCRTHTTPClient(
-            eventloop=self._eventloop,
-            client_config=deepcopy(self._config),
-        )
+        return self
```
arandito: Same comment at the `AIOHTTPClient.__deepcopy__`

arandito (commented):
Overall this PR looks good to me. Left a couple minor comments related to how we preserve config fields when we deepcopy.

packages/smithy-http/src/smithy_http/aio/crt.py
jonathan343: deepcopy related changes were removed from this PR

packages/smithy-http/src/smithy_http/aio/aiohttp.py
jonathan343: deepcopy related changes were removed from this PR

codegen/core/src/main/java/software/amazon/smithy/python/codegen/ClientGenerator.java
jonathan343: deepcopy related changes were removed from this PR

jonathan343:
@Alan4506 Thanks for flagging this. I agree the transport behavior should be consistent, so I updated both implementations to treat `close()` as terminal.

Generated clients now reject operations after closure with `RuntimeError`. Both `AIOHTTPClient.send()` and
`AWSCRTHTTPClient.send()` also reject sends after closure with `SmithyHTTPError`, so CRT will no longer create new
connections after being closed.

I added tests covering sends after closure and context-manager re-entry for both transports.

arandito (approved):
:shipit:

PR #766, opened 2026-08-11
fix: update env provider to skip terminal resolver registration if asume role creds are configured

arandito:
### Description

The `EnvironmentCredentialsProvider` previously registered a terminal resolver whenever `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` were present in the environment. This short-circuited the credential chain even when the active profile was configured to *assume a role* using those environment credentials as the source, meaning the source credentials were handed back directly instead of being used to make the AssumeRole call.

This change makes the provider skip registering a terminal `EnvironmentCredentialsResolver` when a profile is explicitly configure in `IdentityChain.create()` and that profile declares both `role_arn` and `credential_source = Environment`, deferring to assume-role resolution.

Example profile that triggers the skip (`~/.aws/config`):

```
[profile assume]
role_arn = arn:aws:iam::123456789012:role/example
credential_source = Environment
```

With this profile, the environment credentials are treated as the *source* for the AssumeRole call rather than the credentials to return.

### Testing

- Added unit tests in `test_environment.py`:
- Ran `make test-py`- 1881 passed, 9 skipped.
- Ran `make check-py` - ruff, format, and pyright all clean.

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
@@ -4,6 +4,7 @@ import os
 
 from smithy_core.interfaces.identity import Identity
 
+from ....config import load_config
 from ...components import AWSCredentialsIdentity
 from ...environment import EnvironmentCredentialsResolver
 from ..ordering import Standard, StandardProvider
@@ -11,6 +12,8 @@ from ..provider import ChainSetup
 
 _ACCESS_KEY_ID = "AWS_ACCESS_KEY_ID"
 _SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY"  # noqa: S105
+_ROLE_ARN = "role_arn"
+_CREDENTIAL_SOURCE = "credential_source"
 
 
 class EnvironmentCredentialsProvider:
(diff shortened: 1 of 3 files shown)


packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
```diff
@@ -38,4 +41,15 @@ async def setup(
         if not os.getenv(_ACCESS_KEY_ID) or not os.getenv(_SECRET_ACCESS_KEY):
             return
 
+        # Skip environment provider if a profile is explicitly provided and
+        # that profile configures assume role credentials with 'Environment' as the
+        # credential source
+        profile_name = setup.profile_name
+        if profile_name:
+            config = setup.config_file or await load_config()
```
jonathan343, line 49: We aren't caching the config here which means we'll end up doing duplicate work. We need to ensure that we're always only reading the config file only once. Let's also keep this in mind when we integrate with @ubaskota's config resolver work.

Also, might want to make this a method on setup or some other class so we can just reuse loading caching logic.

packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
jonathan343, line 1: I'm just now realizing that when a customer attempts to use the assume role provider and doesn't install `aws-credentials-sts`, they don't get a helpful error message. They just get:

```
smithy_aws_core.identity.chain.exceptions.IdentityChainError: No credential providers were configured to resolve an identity.
```

Not sure if this is another gap in the SEP or a miss on our end. Can you look into this?

jonathan343 (changes requested):

arandito:
Closing in favor of #773

PR #769, opened 2026-08-15
Add awscrt extra to generated pyproject.toml

jonathan343:
## Summary

Generated clients now expose a `[project.optional-dependencies]` awscrt extra so users can opt into the CRT transport (`AWSCRTHTTPClient`) without having to pin a compatible awscrt version themselves. The extra re-exports smithy_http's own awscrt extra, so the version constraint stays sourced from smithy-http.

The set of extras is data-driven through `collectOptionalDependencies`. The awscrt extra is skipped when the client already requires awscrt, such as an http2 service that defaults to the CRT transport, where there is nothing to opt into. Note: These will be swapping to aiohttp as the default soon.

Also drops a dead try/catch in getOptionalDependencies and adds a unit test covering the aiohttp default path.

## Example

This is an example of what gets added to the `pyproject.toml` for affected packages:

```toml
[project.optional-dependencies]
awscrt = [
    "smithy_http[awscrt]~=0.4.0"
]
```

This allows customers to install a compatible version of awscrt through `uv pip install "aws-sdk-sts[awscrt]"`

---

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/SetupGenerator.java
@@ -6,6 +6,7 @@ package software.amazon.smithy.python.codegen.generators;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


arandito (approved):
LGTM

PR #771, opened 2026-08-17
Integrate identity chain into SDK client construction

arandito:
### Description

This PR integrates the new `IdentityChain` into generated AWS clients. A new `AwsIdentityIntegration` builds the default chain into `aws_credentials_identity_resolver` at client setup.

The integration also pins the identity resolver in the config's deep-copy memo (`{id(resolver): resolver}`) so it's shared by reference. Every operation deep-copies the config to isolate per-call plugin mutations, which would otherwise duplicate the resolver on each call; pinning it keeps the chain built once per client and reused across every operation.

> [!IMPORTANT]
> This branch builds on top of #758. **Only the latest commit is the change under review here**. Every earlier commit belongs to #758.

### DIFF
--- a/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsIdentityIntegration.java
+++ b/codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsIdentityIntegration.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.python.aws.codegen;
+
+import java.util.List;
+import software.amazon.smithy.aws.traits.auth.SigV4Trait;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.python.codegen.GenerationContext;
+import software.amazon.smithy.python.codegen.SmithyPythonDependency;
+import software.amazon.smithy.python.codegen.integrations.PythonIntegration;
+import software.amazon.smithy.python.codegen.sections.ClientSetupSection;
+import software.amazon.smithy.python.codegen.writer.PythonWriter;
+import software.amazon.smithy.utils.CodeInterceptor;
+import software.amazon.smithy.utils.CodeSection;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Sets up the default AWS credentials identity chain during client setup.
+ */
+@SmithyInternalApi
+public class AwsIdentityIntegration implements PythonIntegration {
+
+    @Override
+    public List<? extends CodeInterceptor<? extends CodeSection, PythonWriter>> interceptors(
+            GenerationContext context
+    ) {
+        var service = context.settings().service(context.model());
+        if (!service.hasTrait(SigV4Trait.class)) {
+            return List.of();
+        }
+        return List.of(new CredentialsIdentitySetupInterceptor());
+    }
+
+    /**
+     * Initializes the default AWS credentials identity chain during client setup.
+     */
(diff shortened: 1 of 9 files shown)


codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsIdentityIntegration.java
```diff
@@ -0,0 +1,105 @@
...
+            writer.write("""
+                    if self._config.aws_credentials_identity_resolver is None:
+                        async with self._derive_lock:
+                            if self._config.aws_credentials_identity_resolver is None:
+                                config_context = self._config.resolution_context()
+                                config_file = None
+                                profile_name = None
+                                if config_context is not None:
+                                    config_file = await config_context.parsed_profiles()
+                                    profile_name = config_context.profile_name
```
Alan4506: Flagging from https://github.com/smithy-lang/smithy-python/pull/773#discussion_r3806810297: 

Here the generated setup passes `config_context.profile_name` into `IdentityChain.create()`, but the context's `profile_name` is never `None`: it falls back to `AWS_PROFILE`, then `"default"`. See https://github.com/smithy-lang/smithy-python/blob/c0150aa57be10037e7fbe1b1f7e0ba9df3aff391/packages/smithy-aws-core/src/smithy_aws_core/config/context.py#L184-L201

But #773's environment provider defers whenever `profile_name is not None`, so once both PRs land, a user with only env credentials and no profile configured would still get the environment provider skipped. 

We should update this PR to pass the profile only when it comes from the explicit argument.

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsIdentityIntegration.java
```diff
@@ -0,0 +1,105 @@
...
+ * Sets up the default AWS credentials identity chain during client setup.
+ */
+@SmithyInternalApi
+public class AwsIdentityIntegration implements PythonIntegration {
+
+    @Override
+    public List<? extends CodeInterceptor<? extends CodeSection, PythonWriter>> interceptors(
+            GenerationContext context
+    ) {
+        return List.of(
```
alexgromero: Small thing: these interceptors run for every AWS service, but `aws_credentials_identity_resolver` comes from `AwsAuthIntegration` which is gated on `hasSigV4Auth`. So a non SigV4 AWS service still assembles an AWS credentials chain in `_ensure_setup` that it never signs with. Since the interceptors method already receives the `GenerationContext`, would it make sense to return an empty list when the service has no SigV4Trait (mirroring `AwsAuthIntegration`)? Or is there a reason to wire it up regardless? Curious what you think

arandito: We can gate it behind the SigV4Trait. Added in [c5498f2](https://github.com/smithy-lang/smithy-python/pull/771/commits/c5498f2f7822e15a3157094eb626948535b4417c).

codegen/aws/core/src/main/java/software/amazon/smithy/python/aws/codegen/AwsIdentityIntegration.java
arandito: Added in [c5498f2](https://github.com/smithy-lang/smithy-python/pull/771/commits/c5498f2f7822e15a3157094eb626948535b4417c). Needed to improve how we track the profile source to avoid weak string comparisons. They now use `ConfigSource`. 

Chose not to make `profile` a config field since it's not a request lifecycle setting. The profile name mainly influences config resolution so the resolution context is the proper layer for it.

Alan4506 (dismissed):

alexgromero (approved):
LGTM! 🚀

PR #773, opened 2026-08-18
Prioritize assume role credentials and deprecate built-in network providers

arandito:
### Description

This PR aligns credential resolution behavior with the modular credential chain spec changes:

- Resolves assume-role profiles before profile session or static keys.
- Defers profile key providers when the selected profile declares `role_arn`.
- Suppresses environment credentials when an explicit `profile_name` is provided.

This PR also deprecates the built-in IMDS and container resolvers in favor of `aws-credentials-imds` and `aws-credentials-http`.

### Testing
- Added unit coverage for the new precedence and deferment behavior.
- Confirmed new behavior with `aws-credentials-sts` package and `aws-sdk-sts` client

### DIFF
--- a/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
@@ -14,7 +14,7 @@ _SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY"  # noqa: S105
 
 
 class EnvironmentCredentialsProvider:
-    """Adds an environment resolver when credentials are configured in the environment."""
+    """Adds an environment resolver unless an explicit profile is selected."""
 
     @property
     def name(self) -> str:
@@ -33,6 +33,9 @@ class EnvironmentCredentialsProvider:
     ) -> None:
         """Add an environment resolver when env credentials are configured."""
         if identity_type is not AWSCredentialsIdentity:
+            return
+
+        if setup.profile_name is not None:
             return
 
         if not os.getenv(_ACCESS_KEY_ID) or not os.getenv(_SECRET_ACCESS_KEY):
--- a/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
+++ b/packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
@@ -11,6 +11,7 @@ _ACCESS_KEY_ID = "aws_access_key_id"
 _SECRET_ACCESS_KEY = "aws_secret_access_key"  # noqa: S105
 _SESSION_TOKEN = "aws_session_token"  # noqa: S105
 _ACCOUNT_ID = "aws_account_id"
+_ROLE_ARN = "role_arn"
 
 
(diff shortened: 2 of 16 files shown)


packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
```diff
@@ -36,6 +37,9 @@ async def setup(self, identity_type: type[Identity], setup: ChainSetup) -> None:
         if config_file is None or profile_name is None:
             return
 
+        if config_file.get(profile_name, _ROLE_ARN) is not None:
```
Alan4506, line 40: [The comment applies to line 81 as well]

With this deferment, consider a user whose profile looks like:

```ini
[default]
aws_access_key_id = AKIA...
aws_secret_access_key = ...
role_arn = arn:aws:iam::123:role/Foo
source_profile = default
```

but who doesn't have `aws-credentials-sts` installed. The profile key providers now defer, the assume-role provider isn't discovered, and `_find_unclaimed_sources` can't flag it either -   `PROFILE_ASSUME_ROLE.is_detected()` is always `False`. So the user gets an `IdentityChainError` with no "install `aws-credentials-sts`" suggestion, but the SEP says the point of deferring is that "the chain reports the missing STS module".

Is this a gap we should address?

packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/environment.py
```diff
@@ -35,6 +35,9 @@ async def setup(
         if identity_type is not AWSCredentialsIdentity:
             return
 
+        if setup.profile_name is not None:
```
Alan4506, line 38: PR #771 passes `config_context.profile_name` into `IdentityChain.create()`, but it looks like `SharedConfigContext.profile_name` is never `None`? It falls back to AWS_PROFILE, then default. 

https://github.com/smithy-lang/smithy-python/blob/c0150aa57be10037e7fbe1b1f7e0ba9df3aff391/packages/smithy-aws-core/src/smithy_aws_core/config/context.py#L184-L201

So once both PRs land, a user with just `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` set and
no profile configured still gets the environment provider skipped.

Since the context already tracks `profile_origin`, should #771 pass the profile only when it came from the explicit argument?

arandito: Good catch. I'll update #771 to only set the profile if its origin is an explicit override.

packages/smithy-aws-core/src/smithy_aws_core/identity/chain/providers/profile.py
arandito: I am aware of this gap. However adding it to this PR expands its scope. This PR focuses on behavioral changes we want to get in before our next release. The module suggestions improves customer experience but doesn't change behavior.

I already have a follow up PR to add better module suggestions which will add profile based provider suggestions to `ChainSetup` at the `SharedConfigProvider`. We should try to squeeze it in but can definitely wait until a future release.

Alan4506 (approved):

PR #774, opened 2026-08-19
Improve async AWS config typing and plugin lifecycle

jonathan343:
## Overview

This PR improves generated async AWS configuration typing and ensures client and operation plugins run at the correct
scope without mutating caller-owned configuration.

## Summary

- Add typed AWS and service-specific `resolve()` overrides using `TypedDict` and `Unpack`.
- Apply client plugins once during lazy setup; isolate operation plugins with a copy only when needed.
- Restore generated interceptor typing and `set_auth_scheme()`.
- Update API Gateway and DynamoDB plugins to use concrete async config types.
- Make retry-mode resolution case-sensitive and pass resolved retry fields directly.
- Improve constructor behavior, credential documentation, and generated-code coverage.

---

### DIFF
--- a/codegen/aws/core/src/it/java/software/amazon/smithy/python/codegen/test/AwsCodegenTest.java
+++ b/codegen/aws/core/src/it/java/software/amazon/smithy/python/codegen/test/AwsCodegenTest.java
@@ -47,6 +47,13 @@ public class AwsCodegenTest {
         assertTrue(config.contains("api_key: str | None"));
         assertTrue(config.contains("@dataclass(kw_only=True, repr=False, init=False)"));
         assertTrue(config.contains("**overrides: Unpack["));
+        assertTrue(config.contains(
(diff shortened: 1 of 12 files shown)


arandito (approved):
Approved. Thanks @jonathan343!

I tested the codegen output for our clients and confirmed that functionality works. We definitely need to clean up some of the codegen logic but due to time restrictions, this looks good to me.

The `AsyncServiceIdConfig.resolve()` interface looks correct. We can probably also clean up how we deal with the overrides but I think these changes won't affect the overall customer interface.

One bug I found with these updates is that the incorrect initialization no longer surfaces the helpful error message:
```
config = AsyncSTSConfig(region='us-west-2')

Output:
TypeError: AsyncAwsConfig.__init__() got an unexpected keyword argument 'region'
```
I will put up a two-way door fix in my chain integration PR. It's not worth blocking this PR on that.

PR #779, opened 2026-08-21
smithy-http: stream aiohttp response bodies

arandito:
### Description

This PR updates `AIOHTTPClient` to stream response bodies instead of buffering the full payload before returning. This enables modeled streaming outputs to be consumed incrementally and aligns aiohttp response handling with the CRT transport.

The response body uses aiohttp’s raw chunk iterator and exposes `close()` so partially consumed responses can be released. Non-streaming responses continue to be consumed as a complete body during deserialization.

### Testing

- Added unit test coverage for streaming and non-streaming responses, partial consumption cleanup, and response cleanup on marshalling failure.
- Ran `make test-py` - 1975 passed, 9 skipped.
- Generated bedrock runtime client with unreleased changes from #765 and confirmed that output streams now receive events incrementally for `converse_stream` operation. Also tested `converse` to confirm non-streaming operations remain functional.
- Generated sts and dynamo db clients to confirm query and aws json clients are functional as well.


Alan4506 (approved):
LGTM! Thanks @arandito

PR #783, opened 2026-08-23
Fix blob default decoding in generated clients

jonathan343:
## Overview

Client error correction was implemented in [#702](https://github.com/smithy-lang/smithy-python/pull/702), but its AWS JSON protocol tests stayed included in our skip list. After removing those tests, the test for required members without modeled defaults passed. The explicit blob-default test failed with `required_blob: b'YmxvYg==' != b'blob'`. Smithy stores blob defaults as base64 strings, and codegen was copying that string into a bytes literal without decoding it. This PR generates blob defaults with `base64.b64decode()` and enables the related AWS JSON and REST JSON protocol tests.

## Testing

Ran `make test-protocols` successfully after removing the related tests from the protocol skip list.

---


arandito (approved):
:shipit:

PR #794, opened 2026-09-11
Attach response metadata to operation outputs and errors, escaping collisions

ubaskota:
*Issue #, if available:*

*Description of changes:*
Response metadata was previously discarded. Every operation output and modeled error now carries a `response_metadata` attribute holding the request ID, extended request ID, and HTTP status code of the response that produced it. It is always present, so callers need no null check; an `http_status_code` of `None` records that no response arrived.

Codegen writes the field into each generated output class, since outputs share no base class. Modeled errors inherit it from `CallError` instead. `extract_response_metadata()` is added to `ClientProtocol` and called once per call from `_execute_request`, on the path both successes and errors pass through, so each protocol decides where the identifiers live. AWS protocols read response headers, awsQuery reads the response body.

Services that already model a member named `responseMetadata` on an output or error have it renamed to `response_metadata_` so the two cannot collide. Only the Python attribute changes; the wire format is untouched.

Testing:
- All unit tests and protocol tests pass
- Built a test service modelling `responseMetadata` on an output, an error and an input. Verified that codegen escaped the output and error members to `response_metadata_` and left the input untouched.
- Ran the generated client against a fake transport and confirmed the service's own value and the attached metadata both survived, on the success and error paths.
- Exercised awsQuery end to end with no request-ID headers, so the identifier could only come from the body. Verified it was reported on both response shapes: nested under `ResponseMetadata` on success, and directly under the root element on failure.

### DIFF
--- a/codegen/core/src/main/java/software/amazon/smithy/python/codegen/PythonSymbolProvider.java
+++ b/codegen/core/src/main/java/software/amazon/smithy/python/codegen/PythonSymbolProvider.java
@@ -16,6 +16,7 @@ import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.codegen.core.SymbolReference;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.loader.Prelude;
 import software.amazon.smithy.model.shapes.BigDecimalShape;
 import software.amazon.smithy.model.shapes.BigIntegerShape;
@@ -74,6 +75,7 @@ public final class PythonSymbolProvider implements SymbolProvider, ShapeVisitor<
     private final PythonSettings settings;
     private final ServiceShape service;
     private final Set<String> allShapeNames;
+    private final OperationIndex operationIndex;
 
     public PythonSymbolProvider(Model model, PythonSettings settings) {
         this.model = model;
@@ -107,6 +109,9 @@ public final class PythonSymbolProvider implements SymbolProvider, ShapeVisitor<
         // Collect all shape names that will be generated as PascalCase classes in models.py.
         // Used to detect collisions with synthesized names (union variants, unknown types).
         this.allShapeNames = collectAllShapeNames();
+
+        // Built once because toMemberName runs for every member of every shape.
+        this.operationIndex = OperationIndex.of(model);
     }
 
     /**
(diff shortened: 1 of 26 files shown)


codegen/core/src/main/java/software/amazon/smithy/python/codegen/PythonSymbolProvider.java
```diff
@@ -149,12 +154,38 @@ public String toMemberName(MemberShape shape) {
...
+        // by the generator, so a modeled member of the same name is escaped to keep
+        // it from shadowing the attribute.
+        if (CodegenUtils.RESPONSE_METADATA_MEMBER.equals(memberName)
+                && carriesResponseMetadata(container)) {
+            memberName = escapeWord(memberName);
+            LOGGER.warning(() -> format(
+                    "Renamed member %s to \"%s\" because \"%s\" is reserved for response metadata.",
+                    shape.getId(),
+                    escapeWord(CodegenUtils.RESPONSE_METADATA_MEMBER),
+                    CodegenUtils.RESPONSE_METADATA_MEMBER));
```
alexgromero, line 168: nit: could we make this a little more explicit that the generated Python name is being escaped because  `response_metadata` is reserved for SDK response metadata?

```suggestion
            LOGGER.warning(() -> format(
                    "Escaped generated Python name for Smithy member %s from \"%s\" to \"%s\" because \"%s\" is reserved for SDK response metadata.",
                    shape.getId(),
                    CodegenUtils.RESPONSE_METADATA_MEMBER,
                    escapeWord(CodegenUtils.RESPONSE_METADATA_MEMBER),
                    CodegenUtils.RESPONSE_METADATA_MEMBER));
```

packages/smithy-core/.changes/next-release/smithy-core-breaking-ffcfa27938c44f9f9b839d62cde984bc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "breaking",
+  "description": "Added `extract_response_metadata()` to the `ClientProtocol`. Implementations that extend `HttpClientProtocol` inherit a working one and need no change. Protocols that implement `ClientProtocol` directly must add it."
+}
```
alexgromero: Should we also call out a breaking change for the generated name change in the codegen changelog since this is also a generated SDK surface change?

Maybe something like:
"Generated operation outputs and errors now reserve `response_metadata` for SDK response metadata. Modeled `responseMetadata` members on those shapes are generated as `response_metadata_`."

codegen/core/src/test/java/software/amazon/smithy/python/codegen/PythonSymbolProviderTest.java
```diff
@@ -108,6 +108,76 @@ public void testOperationNameCollidingWithClientMethodIsEscaped() {
                         .getName());
     }
 
+    @Test
+    public void testResponseMetadataIsEscapedOnOutputsAndErrors() {
```
alexgromero, line 112: Should we also add a small codegen assertion for this too? This covers the symbol provider name change but an integration assertion could verify that a generated output shape contains both the escaped modeled member and the SDK reserved `response_metadata` field.

packages/smithy-core/.changes/next-release/smithy-core-breaking-ffcfa27938c44f9f9b839d62cde984bc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "breaking",
```
alexgromero, line 2: Should we also call out a breaking change for the generated name change in the codegen changelog since this is also a generated SDK surface change?

Maybe something like:
"Generated operation outputs and errors now reserve `response_metadata` for SDK response metadata. Modeled `responseMetadata` members on those shapes are generated as `response_metadata_`."

PR #803, opened 2026-09-22
Add Smithy RPC v2 CBOR protocol and CBOR codec

kuhe:
Smithy RPC v2 CBOR in Python

- adds noauth codegen support (the CBOR test model has no auth traits)

I'll be dividing this into smaller PRs.


kuhe:
split into stack at https://github.com/smithy-lang/smithy-python/pull/806

PR #806, opened 2026-09-23
Always emit auth config so no-auth services can construct

kuhe:
This unblocks the no-auth CBOR protocol test model, although no-auth should be supported in general as well.


codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -163,7 +164,25 @@ private static List<ConfigProperty> getProtocolProperties(GenerationContext cont
...
-    private static List<ConfigProperty> getAuthProperties(GenerationContext context) {
+    private static List<ConfigProperty> getAuthProperties(GenerationContext context, boolean hasAuth) {
+        Consumer<PythonWriter> authSchemesInit = hasAuth
+                ? writer -> writeDefaultAuthSchemes(context, writer)
+                : writer -> writer.write("self.auth_schemes = auth_schemes or {}");
+
+        Symbol resolverType = hasAuth
+                ? CodegenUtils.getHttpAuthSchemeResolverSymbol(context.settings())
+                : Symbol.builder()
+                        .name("DefaultAuthResolver")
```
arandito: Is `DefaultAuthResolver` too narrow? Should we instead use the base [`AuthSchemeResolver`](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/src/smithy_core/interfaces/auth.py#L26) protocol to allow for custom resolvers?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
```diff
@@ -182,7 +183,25 @@ private static List<ConfigProperty> getProtocolProperties(GenerationContext cont
...
```
arandito: non-blocking/two-way door: Should we use the`AuthSchemeResolver` protocol as the resolver type instead of a concrete class? We can widen the type later if necessary.

I guess the same question goes for the `HTTPAuthSchemeResolver` which is currently a concrete class instead of an interface . Unless we want customers to subclass.

arandito (commented):
Thanks @kuhe!

Looks good just one non-blocking comment/question.

PR #807, opened 2026-09-23
Add smithy-test package and deep_equal protocol-test assertions

kuhe:
pytest somehow doesn't provide deepEquals and we need `NaN` to compare correctly with itself for protocol tests. 

I'm putting this in a new package because we'll need it anyway to house things like snapshot testing.


codegen/core/src/main/java/software/amazon/smithy/python/codegen/SmithyPythonDependency.java
```diff
@@ -64,6 +64,15 @@ public final class SmithyPythonDependency {
             Type.DEPENDENCY,
             false);
 
+    /**
+     * Test-support helpers shared by generated clients (e.g. deep_equal).
+     */
+    public static final PythonDependency SMITHY_TEST = new PythonDependency(
+            "smithy_test",
+            "~=0.0.1",
```
arandito: Since we have a `feature` changelog the next release will release this package as `0.1.0` for its first version. I believe our release infra should automatically update this file whenever the version constraint changes. Can we confirm that?

packages/smithy-test/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Free Threading :: 2 - Beta",
+  "Topic :: Software Development :: Libraries"
+]
+dependencies = [
+    "smithy-core",
```
arandito: Why is the `smithy-core` dependency needed? Are we just adding it now because we'll need in the future?

codegen-output/aws-json-1-0/tests/test_awsjson10_protocol.py
arandito: The NaN test IDs are still in the `TESTS_TO_SKIP` list for each protocol test generator, so these tests are still expected to fail even though they now pass. Should remove them from the skip list to remove the xfail decorator?

I believe there are 8 of these tests now passing across all protocols.

packages/smithy-test/src/smithy_test/_private/equality.py
```diff
@@ -0,0 +1,33 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+import dataclasses
+import math
+from typing import Any
+
+
+def deep_equal(a: object, b: object) -> bool:
```
arandito: question: should we handle `deep_equal(True, 1)`? Looks like an edge case that provides a false positive since it falls back to Python's `==`.

arandito (commented):
Thanks @kuhe!

Left a couple of comments. We should also add the new `smithy-test` package to the issue template.

PR #809, opened 2026-09-23
Add smithy-cbor codec, rpcv2Cbor runtime protocol, and benchmark

kuhe:
Adds the CBOR protocol and codec, an unoptimized Python implementation that passes the protocol tests.

Incremental optimization will be done in later PRs.

This PR also includes a quick benchmark to support optimization in the near future without running the heavier external suites.


packages/smithy-core/src/smithy_core/aio/interfaces/__init__.py
arandito: This is missing a changelog entry. Looks like there aren't any other changes for `smithy-core` staged for release so this won't get released and picked up without an entry.

packages/smithy-http/pyproject.toml
```diff
@@ -39,6 +39,9 @@ dependencies = [
 awscrt = [
     "awscrt~=0.32.0",
 ]
+cbor = [
```
arandito: Should this be a compatible version constraint like the `json` and `xml` extras in `smithy-aws-core`?

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -271,3 +274,211 @@ def _matches_content_type(self, response: HTTPResponse) -> bool:
...
+    with the CBOR-encoded input shape as the body.
+    """
+
+    _id: ShapeID = ShapeID("smithy.protocols#rpcv2Cbor")
+    _content_type: str = "application/cbor"
+    _smithy_protocol: str = "rpc-v2-cbor"
+
+    def __init__(self, settings: ProtocolSettings) -> None:
+        if not _HAS_CBOR:
+            raise ExpectationNotMetError(
```
arandito: We should use `MissingDependencyError` instead to match other protocols.

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -271,3 +274,211 @@ def _matches_content_type(self, response: HTTPResponse) -> bool:
...
+                "Attempted to use the rpcv2Cbor protocol, but smithy-cbor is not "
+                "installed."
+            )
+        self._service_name: str = settings.service_target
+        self._codec: Codec = CBORCodec(  # type: ignore[possibly-unbound]
```
arandito: Why is this suppression needed? I removed it and it passed `make check-py`.

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -271,3 +274,211 @@ def _matches_content_type(self, response: HTTPResponse) -> bool:
...
+    _content_type: str = "application/cbor"
+    _smithy_protocol: str = "rpc-v2-cbor"
+
+    def __init__(self, settings: ProtocolSettings) -> None:
+        if not _HAS_CBOR:
+            raise ExpectationNotMetError(
+                "Attempted to use the rpcv2Cbor protocol, but smithy-cbor is not "
+                "installed."
+            )
+        self._service_name: str = settings.service_target
```
arandito: nit: To stay consistent with other protocol implementations, `self._service_name` and `self._codec` should be typed as `Final`.

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -271,3 +274,211 @@ def _matches_content_type(self, response: HTTPResponse) -> bool:
...
+            ("Accept", self._content_type),
+        ]
+        if _is_unit(operation.input_schema):
+            payload = b""
+        else:
+            payload = self._codec.serialize(shape=input)
+            # The client omits top-level input members left at their modeled default.
+            defaults = _top_level_defaults(operation.input_schema)
+            if defaults:
+                payload = _cbor_strip_defaults(payload, defaults)  # type: ignore[possibly-unbound]
```
arandito: This suppression might also be unnecessary.

packages/smithy-http/src/smithy_http/aio/protocols.py
```diff
@@ -21,13 +22,15 @@
...
-from smithy_core.schemas import APIOperation
+from smithy_core.schemas import APIOperation, Schema
 from smithy_core.serializers import SerializeableShape
 from smithy_core.shapes import ShapeID
 from smithy_core.traits import EndpointTrait, HTTPTrait
 
+from .. import tuples_to_fields
 from ..deserializers import HTTPResponseDeserializer
 from ..serializers import HTTPRequestSerializer
+from . import HTTPRequest as _ConcreteHTTPRequest
```
arandito: Why `_ConcreteHTTPRequest`? Precedent is to rename to `_HTTPRequest`.

PR #811, opened 2026-09-23
Generate the rpcv2Cbor protocol and its protocol tests

kuhe:
initial snapshot of the cbor protocol test generated client after adding it to the generation list


codegen/core/src/main/java/software/amazon/smithy/python/codegen/integrations/Rpcv2CborProtocolGenerator.java
```diff
@@ -0,0 +1,84 @@
...
+        ObjectNode config = ObjectNode.builder()
+                .withMember("http", ArrayNode.fromStrings(trait.getHttp()))
+                .withMember("eventStreamHttp", ArrayNode.fromStrings(trait.getEventStreamHttp()))
+                .build();
+        return ApplicationProtocol.createDefaultHttpApplicationProtocol(config);
+    }
+
+    @Override
+    public void initializeProtocol(GenerationContext context, PythonWriter writer) {
+        writer.addDependency(SmithyPythonDependency.SMITHY_CBOR);
```
arandito: Looks like we're not using the "cbor" optional dependency extra from `smithy-http` that we added in #809. To keep `smithy-cbor` compatible with `RpcV2CborClientProtocol`, let's add smithy-http with the cbor optional dependency instead.

PR #812, opened 2026-09-23
Make protocol configuration user-friendly

kuhe:
Look at what is required to select a different protocol at runtime:

SQS' default is now JSON1, and I select AWS Query:

```py
"""
Runtime protocol selection for a generated Smithy client (SQS).
"""

import asyncio

from smithy_aws_core.aio.protocols import (
    AwsJson10ClientProtocol,
    AwsQueryClientProtocol,
)

from aws_sdk_sqs.client import AsyncSQSClient
from aws_sdk_sqs.config import AsyncSQSConfig
from aws_sdk_sqs.models import ListQueuesInput

from aws_sdk_sqs._private.schemas import AMAZON_SQS as _SERVICE_SCHEMA

REGION = "us-west-2"
SQS_QUERY_VERSION = "2012-11-05"


async def main() -> None:
    config = await AsyncSQSConfig.resolve(
        region=REGION,
        protocol=AwsQueryClientProtocol(_SERVICE_SCHEMA, SQS_QUERY_VERSION),
    )
    client = AsyncSQSClient(config=config)
    result = await client.list_queues(ListQueuesInput())
    print(f"  OK: {len(result.queue_urls)} queues.")
    await client.close()


if __name__ == "__main__":
    asyncio.run(main())

```

#### private import

```py
from aws_sdk_sqs._private.schemas import AMAZON_SQS as _SERVICE_SCHEMA
```

It's not a public API if it involves a private import. 

#### Undiscoverable metadata

```py
SQS_QUERY_VERSION = "2012-11-05"
# ...
    protocol=AwsQueryClientProtocol(_SERVICE_SCHEMA, SQS_QUERY_VERSION)
```

How exactly would I know to provide the required SQS_QUERY_VERSION of `2012-11-05`?

----

## In this PR

We enable a more user-friendly protocol override syntax:

```py
from smithy_aws_core.aio.protocols import (
    AwsQueryClientProtocol,
)
config = await AsyncSQSConfig.resolve(
    region=REGION,
    protocol=AwsQueryClientProtocol,
)
```

No private schema, no magic string. Protocol metadata needed to instantiate any of the protocols is now embedded with the client from its model. The protocol signature now allows passing of only the protocol constructor signaling which protocol the customer wants to use.

This does involve a breaking change to the protocols' constructors to accept a more uniform "ProtocolSettings" metadata object rather than ordered args that would vary depending on the protocol.


packages/smithy-aws-core/src/smithy_aws_core/aio/protocols.py
```diff
@@ -99,6 +100,31 @@ def _assert_event_stream() -> None:
         )
 
 
+@dataclass(kw_only=True, frozen=True)
```
arandito: Are these AWS-specific? If not, should we move them to `smithy-http` or `smithy-core`?

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
jonathan343: So the new approach seems to work for something like:

```python
config = await AsyncServiceConfig.resolve(protocol=SomeProtocolClass)
```

However, it does not support setting it post-resolution:

```python
config.protocol = SomeProtocolClass
```

It doesn't fail initially but seems like it will when it gets further in the request pipeline.

codegen-output/aws-json-1-0/src/awsjson10/config.py
```diff
@@ -94,11 +97,16 @@
     ],
     Interceptor[SimpleScalarPropertiesInput, SimpleScalarPropertiesOutput, Any, Any],
 ]
+_PROTOCOL_SETTINGS = ProtocolSettings(
+    namespace="aws.protocoltests.json10", service_target="JsonRpc10"
+)
```
jonathan343: *Posting our offline discussion for others reviewing the PR*

---

The AWS Query protocol requires the `version` to be specified in the `ProtocolSettings`. For a service like CloudWatch that supports query and more than one alternative protocol, we don't code generate the `version` if query isn't the default protocol chosen. As a result, we will fail if someone attempts to opt into the query protocol.

We should detect when query is a supported protocol and add the `version` setting. Alternatively, we would unconditionally set version since every AWS service defines it. Note: Idk if non-AWS services always have version, but also they can't use query so maybe that doesn't matter.

jonathan343 (commented):
Thanks George! Had two main comments

codegen-output/aws-json-1-0/src/awsjson10/config.py
kuhe: updated the logic - initial PR was wrong.

The selected protocol is not intended to supply the protocolsettings on its own. Rather, a centralized code component will look at all supported protocols and generate the combined set of ProtocolSettings fields. 

So, `version` should appear whenever AwsQuery is supported, not only when it is selected.

codegen/core/src/main/java/software/amazon/smithy/python/codegen/generators/ConfigGenerator.java
kuhe: added handling to the config's `__setattr__`

jonathan343 (approved):
LGTM

jonathan343 (approved):

